PR #58, opened 2018-09-12
Make terminal calls work with tools not on path

jrichter1:
- added ~/.vs-openshift directory to terminal's environment so the calls work even if it is not on path
 - made terminal cwd work
 - replaced some windows specific paths

### DIFF
--- a/src/windowUtils.ts
+++ b/src/windowUtils.ts
@@ -7,7 +7,8 @@ const toolsLocation = path.resolve(Platform.getUserHomePath(), '.vs-openshift');
 export function createTerminal(name: string, cwd: string, env: NodeJS.ProcessEnv = process.env): Terminal {
     let finalEnv: NodeJS.ProcessEnv = {};
     Object.assign(finalEnv, env);
-    finalEnv.PATH = `${toolsLocation}:${env.PATH}`;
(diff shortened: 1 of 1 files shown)


src/windowUtils.ts
```diff
@@ -1,5 +1,19 @@
...
+import { Platform } from './platform';
+
+const toolsLocation = path.resolve(Platform.getUserHomePath(), '.vs-openshift');
  
-export function createTerminal(name: string, cwd: string, env: NodeJS.ProcessEnv = process.env ): Terminal {
-    return window.createTerminal(name, process.platform === 'win32'? undefined : '/bin/bash');
+export function createTerminal(name: string, cwd: string, env: NodeJS.ProcessEnv = process.env): Terminal {
+    let finalEnv: NodeJS.ProcessEnv = {};
+    Object.assign(finalEnv, env);
+    finalEnv.PATH = `${toolsLocation}:${env.PATH}`;
```
dgolovin: It won't work on windows. 'PATH` should be `Path` at least in `finalEnv`

dgolovin: It won't work on windows. `PATH` should be `Path` at least in `finalEnv`

jrichter1: Right, also forgot the delimiter is different.

PR #75, opened 2018-09-18
Add jenkinsfile for builds/releases

jrichter1:
pipeline script for build -> stage -> release (if approved)

Needs a meaningful readme and complete metadata to fully function

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,7 @@ node('rhel7'){
 
 	stage('Package') {
         def packageJson = readJSON file: 'package.json'
-        sh "vsce package -o openshift-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+    		wrap([$class: 'Xvnc']) {
+    			sh "npm test --silent"
+    			junit 'report.xml'
+    		}
+        }
+	}
+
+	stage('Package') {
+        def packageJson = readJSON file: 'package.json'
+        sh "vsce package -o openshift-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
```
mohitsuman: @jrichter1 Lets rename this vsix file as:
`openshift-${packageJson.version}-${env.BUILD_NUMBER}.vsix `

jrichter1: errr, that is the exact same string :)

mohitsuman: ahh, I updated that.

PR #91, opened 2018-09-20
add login to cluster

### LINKED ISSUE #79: Help user if nothing is listed and terminal tells you to log in
I tried listing the project in the resource tree, but it does not show anything. In the terminal I noticed the following output:

```
Please login to your server: 

oc login https://mycluster.mydomain.com
```
* To me things are not obvious enough. One stumbles upon this output while refreshing repeatedly and noticing things moving in the output then. 
* Furthermore one discovers that one has to install an additional tool, oc.
* Then things would be more user-friendly if "odo" would output the correct command (with correct url) right away instead of https://mycluster.mydomain.com

An idea that we discussed is to offer a "log in" cmd in the context menu.
Even more great would be if the tooling was doing that automatically and prompt the user to provide pw/token?

mohitsuman:
Fix #79 

screen shot 2018-09-20 at 6 03 37 am


src/extension.ts
```diff
@@ -228,6 +228,36 @@ export namespace Openshift {
...
+                            return 'Invalid username';
+                        }
+                    }});
+                    if (!username) { return; }
+                    let passwd  = await vscode.window.showInputBox({prompt: "Mention the password"});
+                    odo.execute(`oc login ${clusterIP} -u ${username} -p ${passwd}`).then(()=>{
+                        explorer.refresh();
+                    });
+                } else {
+                    // add logic to store token in settings and use it.
```
dgolovin: Just run `oc login ${clusterIP}  --token='${token}'` and it will be saved in ~/.kube/config.

src/extension.ts
```diff
@@ -228,6 +228,36 @@ export namespace Openshift {
     }
 
     export namespace Explorer {
+        export const login = async function loginCluster(odo: odoctl.Odo, explorer: explorerFactory.OpenShiftExplorer) {
+            if(await odo.requireLogin()) {
+                const clusterIP = await vscode.window.showInputBox({prompt: "Mention the IP address of the cluster", validateInput: (value:string) => {
```
dgolovin: It should be URL.

src/extension.ts
```diff
@@ -228,6 +228,36 @@ export namespace Openshift {
...
+                const loginMethod = await vscode.window.showQuickPick(['Credentials', 'Token'], {placeHolder: 'Do you want to login using'});
+                if(loginMethod === "Credentials") {
+                    let username = await vscode.window.showInputBox({prompt: "Mention the username", validateInput: (value:string) => {
+                        if (value.trim().length === 0) {
+                            return 'Invalid username';
+                        }
+                    }});
+                    if (!username) { return; }
+                    let passwd  = await vscode.window.showInputBox({prompt: "Mention the password"});
+                    odo.execute(`oc login ${clusterIP} -u ${username} -p ${passwd}`).then(()=>{
```
dgolovin: Do not forget to filter out password from odo output channel.

src/odo.ts
```diff
@@ -178,4 +179,12 @@ class OdoImpl implements Odo {
     public async execute(command: string, cwd?: string): Promise<CliExitData> {
         return this.cli.execute(command, cwd ? {cwd} : { });
     }
+
+    public async requireLogin(): Promise<boolean> {
+        const result: cliInstance.CliExitData = await this.cli.execute(
+            `odo version`, {}
+        );
+        // console.log(result.stdout.trim().split('\n')[1] === "Please log in to the cluster");
+        return result.stdout.trim().split('\n')[1] === "Please log in to the cluster";
```
dgolovin: use `result.stdout.indexOf("Please log in to the cluster") > -1` instead.

src/extension.ts
mohitsuman: Updated in the latest commit.

src/extension.ts
mohitsuman: Fixed.

src/extension.ts
mohitsuman: Currently it shows the command as it is passed. Need to figure out a way to hide that from channel or we can also remove the oc login from the channel. Do you recommend any other way for this ?

src/odo.ts
mohitsuman: Fixed.

PR #104, opened 2018-09-24
'Open Console' command for root node in OpenShift Explorer

sudhirverma:


package.json
```diff
@@ -33,6 +33,11 @@
         "title": "About",
         "category": "OpenShift"
       },
+      {
+        "command": "openshift.openshiftconsole",
```
mohitsuman: Use camelCase for the command. `openshiftConsole`

src/extension.ts
```diff
@@ -16,6 +16,11 @@ export namespace Openshift {
         const result:CliExitData = await odo.executeInTerminal(`odo version`, process.cwd());
     };
 
+    export const openshiftconsole = async function (odo: odoctl.Odo) {
```
mohitsuman: Make the camelCase here too.

mohitsuman:
@sudhirverma Where is the action to open in browser. You need to add the action as context menu to `Open in Dashboard`, so that user has a visual representation to open it.

package.json
```diff
@@ -352,6 +361,11 @@
           "when": "view == openshiftProjectExplorer && viewItem == component",
           "group": "c3@3"
         },
+        {
+          "command": "openshift.component.openshiftConsole",
+          "when": "view == openshiftProjectExplorer && viewItem == component",
```
dgolovin: Should be `viewItem == cluster`.

src/extension.ts
```diff
@@ -236,6 +236,11 @@ export namespace Openshift {
             const tls = checkTls ? "https://" : "http://";
             opn(`${tls}${hostName.stdout}`);
         };
+
+        export const openshiftConsole = async function (odo: odoctl.Odo) {
```
dgolovin: since it is called from context menu on cluster node, there should be `context` parameter in the function. 
context is OpenShiftObject instance, therefore you do not need to search for clusters, just do
`opn(context.getName());`

PR #113, opened 2018-09-25
OpenShift Project name validation

sudhirverma:


src/extension.ts
```diff
@@ -28,8 +28,17 @@ export namespace Openshift {
...
             const value  = await vscode.window.showInputBox({prompt: "Project name"});
-            await cli.execute(`odo project create ${value.trim()}`, {});
-            await explorer.refresh();
+            const characterRegex = /[a-z0-9]([-a-z0-9]*[a-z0-9])?/;
+            if(characterRegex.test(value)) {
+                if(value.length<=63) {
+                    await cli.execute(`odo project create ${value.trim()}`, {});
+                    await explorer.refresh();
+                } else {
+                    await vscode.window.showErrorMessage('Project name is to long');
```
mohitsuman: We need to use `too` instead of `to`. And we need to inform the user that the limit length for the project name is 63 characters.

Use this message: `Project name should be within 63 characters`.

src/extension.ts
```diff
@@ -28,8 +28,17 @@ export namespace Openshift {
...
+                }
+            } else {
+                await vscode.window.showErrorMessage('Project name should be alpha or alphe numerical');
```
mohitsuman: Why are 2 words alpha and alphe numerical used. It does not sound right.

Use this message: `Project name should be alphanumeric.`

jeffmaury:
I think the logic of validation should be extracted from this code and exposed as a class or method to be reused

dgolovin:
See how it is done here https://github.com/redhat-developer/vscode-openshift-tools/blob/master/src/extension.ts#L175 and read VSCode API doc here https://code.visualstudio.com/docs/extensionAPI/vscode-api#InputBoxOptions.

src/extension.ts
```diff
@@ -25,20 +25,30 @@ export namespace Openshift {
     }
 
     export namespace Project {
-        export const create = async function createProjectCmd(cli: cli.ICli, explorer: explorerFactory.OpenShiftExplorer) {
-            const projectName = await vscode.window.showInputBox({
+
+        export const projectNameValidation = async function projectName() {
+            const characterRegex = /[a-z0-9]([-a-z0-9]*[a-z0-9])?/;
+            const lengthofCharacterRegex = /^[0-9A-Za-z!@\.;:'"?-]{1,63}$/;
```
dgolovin: why not just `if (value.trim().length <= 63)`

src/extension.ts
```diff
@@ -25,20 +25,30 @@ export namespace Openshift {
...
+            return await vscode.window.showInputBox({
```
dgolovin: method is named `projectNameValidation` but it also request for input, that is not right. Create a method that gets a name as a parameter and returns ether `null` or error message. Then just assign that method to 'validateInput` field in input options.

mohitsuman:
@sudhirverma  Few pointers:
1. Please link the issue which the PR relates too. It is how it should be tracked.
2.  If we have 3 commits, have different commit message for each. Having a generic commit message does not help to understand the change.

PR #141, opened 2018-09-27
add logout navigation if user is logged in

### LINKED ISSUE #139: logout option should hide when user is not logged in cluster
Hide the logout option from cluster navigation actions if user is not logged in to cluster.

mohitsuman:
fix #139 

logout

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
         },
         {
           "command": "openshift.explorer.logout",
-          "when": "view == openshiftProjectExplorer",
+          "when": "view == openshiftProjectExplorer && isLoggedIn",
           "group": "logout"
         },
         {
@@ -388,14 +388,14 @@
     "@types/js-yaml": "^3.11.2",
     "@types/mkdirp": "^0.5.2",
     "@types/mocha": "^5.2.5",
-    "@types/node": "^10.11.0",
+    "@types/node": "^10.11.2",
     "@types/request": "^2.47.1",
     "@types/shelljs": "^0.8.0",
     "@types/validator": "^9.4.2",
     "mocha": "^5.2.0",
     "mocha-jenkins-reporter": "^0.4.0",
     "tslint": "^5.11.0",
-    "typescript": "^3.0.3",
+    "typescript": "^3.1.1",
     "vscode": "^1.1.21"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -252,7 +252,7 @@
         },
         {
           "command": "openshift.explorer.logout",
-          "when": "view == openshiftProjectExplorer",
+          "when": "view == openshiftProjectExplorer && isLoggedIn",
```
dgolovin: should be `context == "isLoggedIn"` otherwise it does not work.

src/extension.ts
```diff
@@ -416,6 +416,7 @@ export namespace Openshift {
         const loginMessage = async (clusterURL, result, explorer)=> {
             if (result.stderr === "") {
                 explorer.refresh();
+                vscode.commands.executeCommand('setContext', 'isLoggedIn', true);
```
dgolovin: You set context only when you login/logout commands executed, but you are forgetting about initial state, right after extension activation.

src/cli.ts
```diff
@@ -127,7 +127,7 @@ async function getToolLocation(cmd): Promise<string> {
         } catch (error) {
             const response = await vscode.window.showInformationMessage(
                 `Cannot find ${tools[cmd].description}.`, 'Download and install', 'Help', 'Cancel');
-            if(response === 'Download and install') {
+            if (response === 'Download and install') {
```
dgolovin: Please do not mix fix with formatting changes.

mohitsuman: sure. will keep this in mind :)

package.json
mohitsuman: It works without setting the context word. Let me see if I can add a gif for it.

src/extension.ts
mohitsuman: I have added the initial state scenario too in this:
https://github.com/redhat-developer/vscode-openshift-tools/pull/141/files#diff-f8d556788b6298850280fcb4c474b5e2R168

package.json
mohitsuman: @dgolovin Updated the description with the gif.

src/extension.ts
```diff
@@ -429,6 +430,7 @@ export namespace Openshift {
                     if (result.stderr === "") {
                         explorer.refresh();
                         const logoutInfo = await vscode.window.showInformationMessage(`Successfully logged out. Do you want to login to a new cluster`, 'Yes', 'No');
+                        vscode.commands.executeCommand('setContext', 'isLoggedIn', false);
```
jrichter1: Just a minor nitpick - please put this line before the info message so the log out option disappears along with the cluster.

mohitsuman: Fixed this.

PR #205, opened 2018-10-05
Added "New Project" action to the cluster node

### LINKED ISSUE #159: Add "New Project" action to the cluster node
Rather than going to more actions -> new project, users should be able to create projects from context menu on the cluster node.

sudhirverma:
Fix #159

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,6 +218,10 @@
           "when": "view == openshiftProjectExplorer"
         },
         {
+          "command": "openshift.project.create",
+          "when": "view == openshiftProjectExplorer"
+        },
+        {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -291,10 +295,35 @@
         }
       ],
       "view/item/context": [
+        {
+          "command": "openshift.project.create",
```
dgolovin: Lets remove them from view/title commands and keep 'isLoggedIn' condition.

dgolovin: Lets leave only Login and Refresh command  in view's command bar.

PR #257, opened 2018-10-22
Unit test for extension.ts file

sudhirverma:
Issue #258

### DIFF
--- a/package.json
+++ b/package.json
@@ -436,7 +436,8 @@
     "sinon": "^7.0.0",
     "tslint": "^5.11.0",
     "typescript": "^3.1.1",
-    "vscode": "^1.1.21"
+    "vscode": "^1.1.21",
+    "sinon-chai": "^3.2.0"
   },
   "dependencies": {
     "byline": "^5.0.0",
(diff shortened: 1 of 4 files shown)


test/extension.test.ts
```diff
@@ -8,20 +8,91 @@
...
+
+    test('should activate extension', async () => {
+        const registerTreeDataProviderStub = sandbox.stub(vscode.window, 'registerTreeDataProvider');
+        await activate(context);
+        expect(registerTreeDataProviderStub).calledOnce;
+	});
+    test('should register all server commands', async () => {
+        return await vscode.commands.getCommands(true).then((commands) => {
+            const SERVER_COMMANDS = [
+                'openshift.about',
```
dgolovin: Would be good to read this from package.json file rather than hard coding it in test itself. This way we can avoid hurdle to update test every time we create new command.

package.json
```diff
@@ -445,6 +446,8 @@
     "opn": "^5.4.0",
     "request-progress": "^3.0.0",
     "shelljs": "^0.8.2",
+    "sinon": "^7.0.0",
+    "sinon-chai": "^3.2.0",
```
jrichter1: Please put testing related libraries into devDependencies

sudhirverma: @jrichter1 done

Issue #190, opened 2018-10-03
Clean up the code and test it


jrichter1:
Review the code, make some files less of a mess, use classes where possible, test them, profit.

dgolovin:
@jrichter1 @mohitsuman Lets open specific issues, like 'Move openshift commands out of extension.ts', 'Add unit tests for openshift explorer' and etc rather than 'Lets make everything perfect'.

jrichter1:
Or you can do that on PR level an just link them here.

mohitsuman:
@jrichter1 the PR should be linked to issues, that would be best way to proceed.

PR #260, opened 2018-10-24
Split commands into smaller logical blocks

jrichter1:
The single command file was broken into smaller pieces:
 - one class per openshift structure (project, application, service, etc.)
 - made resources like cli, explorer and odo singletons
 - moved reference to resources from method signature to static fields
 - moved info and error messages from individual commands to the code that calls them
   - methods return statuses as string
   - errors are handled through promise rejection / try-catch

The logic itself has not been changed.


jrichter1:
Part 1 of #190

sudhirverma:
@jrichter1 please resolve conflicts

PR #266, opened 2018-10-29
Move utility functions into util/ classes

jrichter1:
moving utilities into classes under src/util/
part 2 of #190


mohitsuman:
LGTM. +1

Issue #232, opened 2018-10-16
Add 'OpenShift: Delete Project' command to command palette


dgolovin:
Project node in OpenShift Application Explorer view has 'Delete' command 
![image](https://user-images.githubusercontent.com/620330/48109067-cb149f00-e1fa-11e8-9a90-1a7847dd48a6.png)
which is currently not visible in command palette. That was done because 'Delete' command expects to be called on Project node, has no check if node is not provided and would conflict with other `Delete` commands for Aplication/Component/Service/Storage.

To make it available in command palette there should be command that would be named 'Delete Project' and it would be shown in palette only. 
![image](https://user-images.githubusercontent.com/620330/48109122-feefc480-e1fa-11e8-9e33-b903dd74a14c.png)
This wrapper command should provide QuickPick to select projects from cluster and then call 'Delete' command for project with Project selected from QuickPick list.

PR #267, opened 2018-10-30
Add 'OpenShift: Delete Project' command to command palette

sudhirverma:
Issue: #232

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,11 @@
       },
       {
         "command": "openshift.project.delete",
+        "title": "Delete",
+        "category": "OpenShift"
+      },
+      {
+        "command": "openshift.project.delete.palette",
         "title": "Delete Project",
         "category": "OpenShift"
       },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
       },
       {
         "command": "openshift.project.delete",
-        "title": "Delete",
+        "title": "Delete Project",
```
dgolovin: Delete context menu items in OpeShift Application Explorer should stay named as 'Delete' not `Delete ${objectType}`.

src/openshift/project.ts
```diff
@@ -22,13 +22,22 @@ export class Project extends OpenShiftItem {
     }
 
     static async del(context: OpenShiftObject): Promise<string> {
```
dgolovin: This command is perfectly fine, it is always called from context menu in OpenShift App Explorer and never from Commend Palette, so there is no need to change it. What you want to do is to create new command 
``` javascript
static async delProject(): Promise<string> {
// collect all the required info here
// get instance of OpenShiftObject being deleted
// call static del function with this instance
}
```

Issue #234, opened 2018-10-16
Add 'OpenShift: Delete Component' command to command palette


dgolovin:
The idea is to provide 'OpenShift: Delete Component' command to command palette the same way it is described in #232.

dgolovin:
fixed via  ab00b7c.

PR #280, opened 2018-11-02
'OpenShift: Delete Application' command to command palette

sudhirverma:

PR #285, opened 2018-11-05
'OpenShift: Delete Component' command to command palette

sudhirverma:
Issue: #234

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,11 @@
         "category": "OpenShift"
       },
       {
+        "command": "openshift.component.delete.palette",
+        "title": "Delete Component",
+        "category": "OpenShift"
+      },
+      {
         "command": "openshift.component.describe",
         "title": "Describe",
         "category": "OpenShift"
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -36,7 +36,9 @@ export class Component extends OpenShiftItem {
             component = treeItem;
         } else {
             project = await vscode.window.showQuickPick(Component.odo.getProjects(), {placeHolder: "From which project you want to delete Component"});
-            application = await vscode.window.showQuickPick(Component.odo.getApplications(project), {placeHolder: "From which application you want to delete Component"});
+            if (project) {
+                application = await vscode.window.showQuickPick(Component.odo.getApplications(project), {placeHolder: "From which application you want to delete Component"});
+            }
             if (application) {
                 component = await vscode.window.showQuickPick(Component.odo.getComponents(application), {placeHolder: "Select Component to delete"});
             }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -301,6 +306,11 @@
           "when": "view == openshiftProjectExplorer && viewItem == cluster && isLoggedIn",
           "group": "c4"
         },
+        {
```
dgolovin: Let's do not pollute cluster's context menu.

src/openshift/component.ts
```diff
@@ -28,16 +28,43 @@ export class Component extends OpenShiftItem {
         }
     }
 
+    static async delComponent () {
```
dgolovin: Could you change it to work the same way as I implemented for `OpenShift: Delete Application`. It is much easier right now.

src/openshift/component.ts
```diff
@@ -28,16 +28,43 @@ export class Component extends OpenShiftItem {
...
+                const applications: OpenShiftObject[] = await Component.odo.getApplications(project);
+                const applicationsNames: string[] = applications.map((value)=> value.getName());
+                const applicationName = await vscode.window.showQuickPick(applicationsNames, {placeHolder: "From which Application you want to delete Component"});
+                applications.forEach(async (application)=> {
+                    if (application.getName() === applicationName ) {
+                        const components: OpenShiftObject[] = await Component.odo.getComponents(application);
+                        const componentsNames: string[] = components.map((value)=> value.getName());
+                        const componentName = await vscode.window.showQuickPick(componentsNames, {placeHolder: "Select Component to delete"});
+                        if (applicationName) {
+                            return Component.delComponentByName(componentName, application, project);
```
dgolovin: Do not forget to add 
``` JavaScript
.then(()=>Component.explorer.refresh());
```
here to avoid the same problem as reported in #280.

src/openshift/component.ts
```diff
@@ -28,16 +28,32 @@ export class Component extends OpenShiftItem {
         }
     }
 
+    static async delComponent () {
```
jrichter1: This method should not be in the API. Either move the code into del, or make it private and then call it from del.

src/openshift/component.ts
```diff
@@ -28,16 +28,32 @@ export class Component extends OpenShiftItem {
...
+    }
+
     static async del(context: OpenShiftObject): Promise<string> {
         const app: OpenShiftObject = context.getParent();
         const project: OpenShiftObject = app.getParent();
-        const value = await vscode.window.showWarningMessage(`Are you sure you want to delete component '${context.getName()}\'`, 'Yes', 'Cancel');
+        return Component.delComponentByName(context.getName(), app, project);
+    }
+
+    static async delComponentByName(name: string, app: OpenShiftObject, project: OpenShiftObject): Promise<string> {
```
jrichter1: This method should be private.

package.json
```diff
@@ -92,6 +92,11 @@
         "title": "New Component",
         "category": "OpenShift"
       },
+      {
+        "command": "openshift.component.delete.palette",
```
jrichter1: AFAIK, this does not need to be a separate command - why not use the existing component.delete?

dgolovin: The name in context menu 'Delete' if we use the same command for palette and context menu we end up having bunch of `OpenShift: Delete' commands. One per model object.

src/openshift/component.ts
dgolovin: I did not realize it is possible to register the same method twice with different names:
* `OpenShift: Delete Component` for palette
* `Delete` for `OpenShift Application Explorer`

Example is here https://github.com/redhat-developer/vscode-openshift-tools/pull/297/files#diff-85d649dcccfbc33846ac83964c2de415R32.

package.json
jrichter1: So, if the existing command was renamed to, say, 'Delete Component', would that be a problem? I'm guessing it would show as such in the context menu as well.

PR #296, opened 2018-11-08
Use && to concat several command to make it work both linux and win

### LINKED ISSUE #293: `Open in Browser` does not work when cluster has more than one project
The `Open in Browser` command tries  to load a route based on the component name but does not use the application and the project in the filter

dgolovin:
Fix #293.


dgolovin:
testPR

PR #306, opened 2018-11-09
Migrate odo commands to new options for seting current proj/app/comp

### LINKED ISSUE #270: Migrate to use odo --project/--application/--component options
OpenShift commands should use adwantage of new odo options instead of relying on several commands to set current context for command execution.

dgolovin:
Fix #270.


jrichter1:
I think you forgot to migrate https://github.com/redhat-developer/vscode-openshift-tools/blob/7e47011232f9293a15547e5467eb04cd195b0c3a/src/openshift/service.ts#L57

src/openshift/service.ts
```diff
@@ -33,11 +33,13 @@ export class Service extends OpenShiftItem {
...
             });
             if (serviceName) {
+                const app = context.getParent();
+                const project = app.getParent();
                 return Progress.execWithProgress({
                     cancellable: false,
                     location: vscode.ProgressLocation.Notification,
                     title: `Creating new service '${serviceName}'`
-                }, [{command: `odo project set ${context.getParent().getName()} && odo app set ${context.getName()} && odo service create ${serviceTemplateName} --plan ${serviceTemplatePlanName} ${serviceName.trim()}`, increment: 100}
+                }, [{command: `&& odo service create ${serviceTemplateName} --plan ${serviceTemplatePlanName} ${serviceName.trim()} --app ${app.getName()} --project ${project.getName()}`, increment: 100}
```
dgolovin: .. and this is wrong as well.

Issue #278, opened 2018-11-02
Save oc and odo exec binaries in ~/.vs-openshift/cache/${tool-name}/${version}  


dgolovin:
Different versions of OpenShift might require specific versions of `oc` and `odo`, would be good to be able to save them and use when specific version is needed. For now we are going to use specific one from `toolsConfig.ts` file.

dgolovin:
Closing due to #1227 and #1355.

PR #309, opened 2018-11-12
Save oc and odo exec binaries in `~/.vs-openshift/cache/${tool-name}/${version}`

sudhirverma:
Issue: #278

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14,6 +14,7 @@ export interface CliExitData {
 }
 
 export class Cli implements ICli {
+
     private static instance: Cli;
     private odoChannel: OdoChannel = new OdoChannelImpl();
 
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -96,22 +96,26 @@ export class ToolsConfig {
         ToolsConfig.tools = ToolsConfig.loadMetadata(configData, Platform.OS);
     }
 
+    private static getpath(cmd, filename) {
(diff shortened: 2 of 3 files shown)


src/cli.ts
```diff
@@ -65,8 +65,8 @@ export class Cli implements ICli {
     }
 
     private async getToolLocation(cmd: string): Promise<string> {
-        let toolLocation = path.resolve(Platform.getUserHomePath(), '.vs-openshift', this.tools[cmd].cmdFileName);
-        const toolDlLocation = path.resolve(Platform.getUserHomePath(), '.vs-openshift', this.tools[cmd].dlFileName);
+        let toolLocation = path.resolve(Platform.getUserHomePath(), `.vs-openshift\\cache\\${this.tools[cmd].name}\\${this.tools[cmd].version}`, this.tools[cmd].cmdFileName);
```
dgolovin: [path.resolve(...paths)](https://nodejs.org/api/path.html#path_path_resolve_paths) allows you to array of path' segments and avoid to deal with path separators directly. Instead of using strings with `\\` inside, which would work only for windows, just pass each segment as a parameter.

``` JavaScript
path.resolve(Platform.getUserHomePath(), '.vs-openshift', 'cache', cmd, this.tools[cmd].version)
```

dgolovin: Would be also good to have a template method for generating the path to avoid code duplication.

sudhirverma:
@dgolovin review this pr

PR #330, opened 2018-11-14
Propagate errors from odo

### LINKED ISSUE #249: odo execution errors are passing through undetected
This happens here https://github.com/redhat-developer/vscode-openshift-tools/blob/6ecc04d50709726d448d02d81128dcc674fbc740/src/cli.ts#L101, because odo always return 1 as error codewhich is common practice.

jrichter1:
attempt to fix #249. Not sure if it's the best solution, but I don't fancy changing all the code that uses odo.execute


jrichter1:
closing as this was done in #406

PR #331, opened 2018-11-14
Update odo url and checksums

jrichter1:
odo releases for linux and mac are no longer archives it seems


dgolovin:
https://github.com/redhat-developer/odo/issues/982 is fixed.

PR #333, opened 2018-11-14
Check for configured/downloaded tools to make sure correct version is used

### LINKED ISSUE #245: Download required version of odo if detected version does not match required one
Extension still should download and use specific version of odo to avoid command line options incompatibilities.

dgolovin:
Fix #245.


dgolovin:
testPR

PR #382, opened 2018-11-21
 support odo link for service/component to component

### LINKED ISSUE #15: 'Link' command and context menu item for component node

mohitsuman:
Fix #15


mohitsuman:
@dgolovin Please review and merge.

jeffmaury:
Comments:

- it is possible to link a component to itself
- if there is no component at all, linking from a service will show an empty quick pick which immediately disappear (if I remember correctly)

mohitsuman:
> Comments:
> 
> * it is possible to link a component to itself

Currently `odo link` does allows that. But from user point of view, we can remove that in VSCode.

> * if there is no component at all, linking from a service will show an empty quick pick which immediately disappear (if I remember correctly)

Yeah there should be a check for that before loading the quick pick.

src/openshift/component.ts
```diff
@@ -87,6 +88,18 @@ export class Component extends OpenShiftItem {
         Component.odo.executeInTerminal(`odo log ${context.getName()} -f --app ${app.getName()} --project ${project.getName()}`);
     }
 
+    static async link(context: OpenShiftObject): Promise<String> {
+        const app: OpenShiftObject = context.getParent();
+        const project: OpenShiftObject = app.getParent();
+        const componentToLink = await vscode.window.showQuickPick(Component.odo.getComponents(app), {placeHolder: "Select the component to link"});
```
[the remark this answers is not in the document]
mohitsuman: Here `OpenShiftObject` does not have filter function, so have implemented this using the array we have after the await call.

src/odo.ts
```diff
@@ -162,9 +162,17 @@ export class OdoImpl implements Odo {
         const result: cliInstance.CliExitData = await this.execute(
             `oc get dc --namespace ${proj} -o jsonpath="{range .items[?(.metadata.labels.app == \\"${application.getName()}\\")]}{.metadata.labels.app\\.kubernetes\\.io/component-name}{\\"\\n\\"}{end}"`
         );
-        return result.stdout.trim().split('\n')
+
+        const componentsList = result.stdout.trim().split('\n')
             .filter((value) => value !== '')
             .map<OpenShiftObject>((value) => new OpenShiftObjectImpl(application, value, 'component', this, TreeItemCollapsibleState.Collapsed));
+        if (componentsList.length>0) {
```
dgolovin: you can do simply:
``` JavaScript
commands.executeCommand('setContext', 'componentPresent', componentsList.length>0);
```
I would also add  `twoComponentsPresent' to block  `openshift.component.link` action if only one component is available, but that is optional.

mohitsuman: > you can do simply:
> 
> ```js
> commands.executeCommand('setContext', 'componentPresent', componentsList.length>0);
> ```

Fixed.

dgolovin:
please rebase and squash commits when merging.

PR #406, opened 2018-11-27
Stop downloading odo/oc when cluster is down

### LINKED ISSUE #386: ODO not found when cluster is down
When the extension checks the version of ODO it always complains it can't find ODO if the cluster is down. The problem seems to be that 'odo version' command actually tries to connect to the cluster and returns an error if it can't.

dgolovin:
fix #386.

### DIFF
--- a/src/util/progress.ts
+++ b/src/util/progress.ts
@@ -7,6 +7,8 @@
 
 import * as vscode from 'vscode';
 import * as odoctl from '../odo';
+import { CliExitData } from '../cli';
+import { ExecException } from 'child_process';
 
 export interface Step {
     command: string;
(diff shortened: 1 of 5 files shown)


src/util/progress.ts
```diff
@@ -20,30 +22,34 @@ export class Progress {
...
-                    current.total = previous.total + current.increment;
-                    return currentIndex+1 === steps.length ? result.then(()=> {
+                    return new Promise<CliExitData>(async (resolve,reject) => {
                         progress.report({
                             increment: previous.increment,
                             message: `${previous.total}%`
                         });
-                    }) : result;
+
+                            const result: CliExitData = await odo.execute(current.command);
```
jrichter1: If an error happens here, it gets swallowed and the progress is stalled. You might need to call this with fail = false to pass it into the handling below.

PR #415, opened 2018-11-28
Throw error with output as a message if odo command fails for catalog

### LINKED ISSUE #414: `New Service` command fails silently if service-catalog is not eanble for active cluster

dgolovin:
Fix #414.


dgolovin:
testPR

PR #432, opened 2018-12-03
'OpenShift: Create URL' command to command palette

sudhirverma:
Issue: #320


sudhirverma:
testPR

src/openshift/url.ts
```diff
@@ -22,11 +23,60 @@ export class Url {
...
+        if (!urlName) {
+            name = await Url.getProjectName();
+        }
+        if (urlName) {
+            urlName = url;
+        }
+        return name || urlName ? name || urlName : undefined;
+    }
+
+    static async getProjectName() {
```
jrichter1: I have a fundamental issue with the way these helper methods are being added. From what I've seen the same ~4 methods appear in about 5 pull requests, each time in a different class.

Copy-paste on this level is one of the worst coding practices (really, it is). In order to materialize some more elegant solution, I've created #499 with a proposal on how to proceed. Feel free to chime in and discuss potential solutions. 
But this code is not an acceptable way forward.

sudhirverma:
@jrichter1 review this pr

PR #446, opened 2018-12-04
OpenShift: Describe Application command to command palette

### LINKED ISSUE #327: Add 'OpenShift:  Describe Application' command to command palette

sudhirverma:
Fix #327.

### DIFF
--- a/src/openshift/openshiftItem.ts
+++ b/src/openshift/openshiftItem.ts
@@ -13,4 +13,20 @@ export abstract class OpenShiftItem {
     static create(context: OpenShiftObject): Promise<String> { return Promise.reject(); }
     static del(context: OpenShiftObject): Promise<String> { return Promise.reject(); }
     static wait(timeout: number = 2500): Promise<void> { return  new Promise((res)=>setTimeout(res, timeout)); }
+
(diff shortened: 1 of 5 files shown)


sudhirverma:
@dgolovin please review this pr

src/openshift/openshiftItem.ts
```diff
@@ -13,4 +13,20 @@ export abstract class OpenShiftItem {
     static create(context: OpenShiftObject): Promise<String> { return Promise.reject(); }
     static del(context: OpenShiftObject): Promise<String> { return Promise.reject(); }
     static wait(timeout: number = 2500): Promise<void> { return  new Promise((res)=>setTimeout(res, timeout)); }
+
+    static async getProjectNames(): Promise<OpenShiftObject[]> {
```
jrichter1: My vision was that instead of several methods being available to all subclasses, each concerned subclass would have its own implementation of the same method. Which got me thinking, all the functionality in the OpenShiftItem classes is static, making the overarching method definition basically useless, since static methods can't be overridden. We might need to rethink this API.

dgolovin: Lets go ahead and implement all the commands using approach we took. Then we'll create new API and migrate to it.

PR #447, opened 2018-12-04
OpenShift: Describe Component command to command palette

sudhirverma:
Issue: #328

### DIFF
--- a/src/openshift/openshiftItem.ts
+++ b/src/openshift/openshiftItem.ts
@@ -5,6 +5,7 @@
 
 import { Odo, OdoImpl, OpenShiftObject } from '../odo';
 import { OpenShiftExplorer } from '../explorer';
+import * as vscode from 'vscode';
 
 export abstract class OpenShiftItem {
     protected static readonly odo: Odo = OdoImpl.getInstance();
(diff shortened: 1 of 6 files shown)


src/openshift/openshiftItem.ts
```diff
@@ -29,4 +30,24 @@ export abstract class OpenShiftItem {
...
+
+    static async getComponentNames(application) {
+        const applicationList: Array<OpenShiftObject> = await OpenShiftItem.odo.getComponents(application);
+        if (applicationList.length === 0) {
+            throw Error('You need at least one Component available. Please create new OpenShift Component and try again.');
+        }
+         return applicationList;
+    }
+
+    static async getOpenShiftCmdData(treeItem: OpenShiftObject, projectPlaceholder: string, appPlaceholder: string, compPlaceholder?: string) {
```
dgolovin: This belongs to component.ts file, because it is used only for component related commands.

sudhirverma: n this function, I am checking for both application and component related commands

dgolovin: Good thinking, but I was aiming for much simpler functional approach, when you have one method to collect all required objects per OpenShift resource:
* Application
* Component
* Storage
Now you have method that works for Application and Component. It is already hard enough to understand, now imagine you add support for Storage into this method.

PR #477, opened 2018-12-13
Added unit test for linkService method

sudhirverma:


test/openshift/component.test.ts
```diff
@@ -286,6 +288,40 @@ suite('Openshift/Component', () => {
...
+        });
+
+        test('returns null when no service Link found', async () => {
+            quickPickStub.resolves();
+            const result = await Component.linkService(componentItem);
+
+            expect(result).null;
+        });
+
+        test('errors when a subcommand fails', async () => {
```
dgolovin: This test should fail if there was no error triggered. In current implementation it will pass.

test/openshift/component.test.ts
```diff
@@ -286,6 +288,40 @@ suite('Openshift/Component', () => {
...
+
+        test('works from context menu', async () => {
+            quickPickStub.resolves(serviceItem);
+            const result = await Component.linkService(componentItem);
+
+            expect(result).equals(`service '${serviceItem.getName()}' successfully linked with component '${componentItem.getName()}'`);
+            expect(execStub).calledOnceWith(`odo project set ${projectItem.getName()} && odo application set ${appItem.getName()} && odo component set ${componentItem.getName()} && odo link ${serviceItem.getName()} --wait`);
+        });
+
+        test('returns null when no service Link found', async () => {
```
dgolovin: Should it be `returns null when no service to link found'?

test/openshift/component.test.ts
```diff
@@ -286,6 +288,40 @@ suite('Openshift/Component', () => {
...
+            expect(result).null;
+        });
+
+        test('errors when a subcommand fails', async () => {
+            quickPickStub.resolves(serviceItem);
+            execStub.rejects(errorMessage);
+
+            try {
+                await Component.linkService(componentItem);
+            } catch (err) {
```
dgolovin: should be 
``` JavaScript
let savedErr;
try {
    await Component.linkService(componentItem);
} catch (err) {
    savedErr = err;
}
expect(savedErr).equals(`Failed to link service with error '${errorMessage}'`);
```

PR #483, opened 2018-12-21
Added validation for New Component to check there are component with the same name in cluster

### LINKED ISSUE #182: Add validation for New Commands to check there are no resources with the same name in cluster
Currently New Application allow to continue and command fails to create new resource.

sudhirverma:
Fix #182.

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -14,6 +14,14 @@ import { Url } from './url';
 import { CliExitData } from '../cli';
 import { V1ServicePort, V1Service } from '@kubernetes/client-node';
 
+export const Validator = {
+    emptyName: (message: string, value: string) => {
+        if (validator.isEmpty(value)) {
+            return message;
+        }
+    }
+};
+
 export class Component extends OpenShiftItem {
 
     static async create(context: OpenShiftObject): Promise<string> {
@@ -184,11 +192,8 @@ export class Component extends OpenShiftItem {
 
     private static async validateComponentName(value: string, application: OpenShiftObject) {
         const componentList: Array<OpenShiftObject> = await Component.odo.getComponents(application);
-        return componentList.map((component) => {
-            if (component.getName().trim() === value) {
-                return `The Component '${component.getName()}' already exists. Please try some other component name`;
-            }
-        }).filter((component) => { return component; })[0];
+        const componentName =  componentList.find((component) =>  component.getName() === value);
+        return componentName && `This component name already exists. Please try some other component name`;
     }
 
     private static async createFromLocal(application: OpenShiftObject): Promise<string> {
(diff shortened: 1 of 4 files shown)


src/openshift/component.ts
```diff
@@ -163,6 +163,15 @@ export class Component extends OpenShiftItem {
         return null;
     }
 
+    private static async validateComponentName(value: string, application: OpenShiftObject) {
```
dgolovin: I am sure you have one project, one application and one component. This is the only condition when it works.

src/openshift/component.ts
```diff
@@ -163,6 +163,15 @@ export class Component extends OpenShiftItem {
...
+        const componentList: Array<OpenShiftObject> = await Component.odo.getComponents(application);
+        return componentList.map((component) => {
```
dgolovin: This is not how you should do this 👎 try to use `Array.indexOf()` or `Array.includes()`.

src/openshift/component.ts
```diff
@@ -171,10 +180,11 @@ export class Component extends OpenShiftItem {
 
         const componentName = await vscode.window.showInputBox({
             prompt: "Component name",
-            validateInput: (value: string) => {
+            validateInput: async (value: string) => {
```
dgolovin: Starting from this one all validations below are code duplication or CDs. Even if the message is different. Try to create a additional object  which would be Validator Factory and function to create isEmpty validator. Validator created would be a function that check if name is empty and return provided message. That would look like:
``` JavaScript
export const Validator = {
    emptyName: function (message: string): string | null | undefined {
        return function(name: string) {
            if(validator.isEmpty(name.trim()) {
                return message;
        }
    } 
}
```
so everywhere when you need to check for empty name you would use
``` JavaScript
Validator.emptyName('Message about something being empty');
```
To validate git repo URL is empty you would do something like this

``` JavaScript
        const repoURI = await vscode.window.showInputBox({
            prompt: 'Git repository URI',
            validateInput: (value: string) => {
                let message = Validator.emptyName('Empty Git repository URL')(value);
                if(message) {
                     return message;
                }
                if (!validator.isURL(value)) {
                     return 'Invalid URL provided';
                 }
             }
```

dgolovin:
And please fix failing tests.

src/openshift/component.ts
sudhirverma: I think we can't use indexof() and includes() because we don't have the reference of the object here. We only have the component name as string which is a property of the object inside array. So we need to use a higher order functions like `Array.prototype.find()` or `Array.prototype.findIndex()`.

sudhirverma:
testPR

sudhirverma:
testPR

sudhirverma:
@dgolovin  please review this pr

PR #504, opened 2019-01-09
Fix openshift explorer not realizing it is logged out of a cluster

### LINKED ISSUE #506: OpenShift Explorer sometimes does not detect when current session is expired and new loging required
This happens because detection is done based on stdout content and extension does not have all possible messages that indicates expired session.

jrichter1:
Because it could not find the output that would say login is required.
Fix #506.

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -207,6 +207,12 @@ export function getInstance(): Odo {
 export class OdoImpl implements Odo {
     private static cli: cliInstance.ICli = cliInstance.Cli.getInstance();
     private static instance: OdoImpl;
+    private readonly odoLoginMessages = [
+        'Please log in to the cluster',
+        'the server has asked for the client to provide credentials',
+        'Please login to your server'
+    ];
+
     private constructor() {}
 
     public static getInstance(): OdoImpl {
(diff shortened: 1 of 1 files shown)


src/odo.ts
```diff
@@ -298,7 +298,7 @@ export class OdoImpl implements Odo {
         const result: cliInstance.CliExitData = await this.execute(
             Command.printOdoVersionAndProjects(), process.cwd(), false
         );
-        if (result.stdout.indexOf('Please log in to the cluster') > -1 || result.stdout.indexOf('the server has asked for the client to provide credentials') > -1) {
+        if (result.stdout.indexOf('Please login to your server') > -1 || result.stdout.indexOf('the server has asked for the client to provide credentials') > -1) {
```
mohitsuman: The odo(0.0.16) output shows the following message during login: `Please log in to the cluster`.

jrichter1: not for me, it doesn't

jrichter1: right, there are in fact two spots in odo code that fail on login with different messages

Please log in:
https://github.com/redhat-developer/odo/blob/eec9a26ec4b7520e90a1d94370c3ce92f84d5f59/pkg/occlient/occlient.go#L246

Please login:
https://github.com/redhat-developer/odo/blob/eec9a26ec4b7520e90a1d94370c3ce92f84d5f59/pkg/occlient/occlient.go#L182 

No idea what the differentiating factor is.

jrichter1: Hah!
```

jeffmaury: We should open an issue on the odo side, WDYT ?

mohitsuman: We need to open an issue on odo side then.

jrichter1: I don't think this is odo's fault, it makes sense to have different errors for these two cases. We just haven't realized this path exists in odo.

dgolovin: It might be just message from openshift and it could be different for different versions.
But anyway it worth to open issue for odo.

PR #517, opened 2019-01-17
code refactor for application

sudhirverma:


src/openshift/application.ts
```diff
@@ -8,44 +8,23 @@ import { OpenShiftObject, Command } from '../odo';
...
 
-    static async create(project: OpenShiftObject): Promise<String> {
-
-        const data = await Application.getApplicationData(project);
+    static async create(treeItem: OpenShiftObject): Promise<String> {
 
-        if (data) {
+        const project = await Application.getOpenShiftCmdData(treeItem,
+            "In which Project you want to create an Application");
+        const name = await Application.getApplicationName();
```
dgolovin: Yo are missing a check for project being actually selected.

dgolovin:
``` JavaScript
let name;
if (project) {
    name = await Application.getApplicationName();
}
```

PR #522, opened 2019-01-18
Code refactor for service delete method

sudhirverma:


src/openshift/service.ts
```diff
@@ -51,24 +51,18 @@ export class Service extends OpenShiftItem {
...
-        let service: OpenShiftObject = treeItem;
-        if (!service) {
-            project = await vscode.window.showQuickPick(Service.odo.getProjects(), {placeHolder: "From which Project you want to delete Service"});
-            application = await vscode.window.showQuickPick(Service.odo.getApplications(project), {placeHolder: "From which Application you want to delete Service"});
-            if (application) {
-                service = await vscode.window.showQuickPick(Service.odo.getServices(application), {placeHolder: "Select Service to delete"});
-            }
-        }
+        let service: OpenShiftObject;
+        service = await Service.getOpenShiftCmdData(treeItem,
```
dgolovin: this should be ```const application = await Service.getOpenShiftCmdData(treeItem,```

src/openshift/service.ts
```diff
@@ -51,24 +51,18 @@ export class Service extends OpenShiftItem {
...
+            "From which Project you want to delete Service",
+            "From which Application you want to delete Service"
+        );
+        if (!treeItem) service = await vscode.window.showQuickPick(Service.getServiceNames(service), {placeHolder: "Select Service to delete"});
```
dgolovin: ```if (!application) service = await vscode.window.showQuickPick(Service.getServiceNames(application), {placeHolder: "Select Service to delete"});```

sudhirverma: `!application` always return false

src/openshift/service.ts
dgolovin: I am confused. First you call 
``` TypeScript
Service.getOpenShiftCmdData(treeItem, 
    "From which Project you want to delete Service",
    "From which Application you want to delete Service"
);
```
and it passes through treeItem, or returning application istance or null.
Why variable is called `service` then?

src/openshift/service.ts
dgolovin: After 
``` TypeScript
Service.getOpenShiftCmdData(treeItem, 
    "From which Project you want to delete Service",
    "From which Application you want to delete Service"
);
```
`application` variable would be ether the same `treeItem` or application instance if `treeItem` is `null | undefined` or `null` if quickpick was canceled.

dgolovin: Your code would probably work, I have not tested it, but it just look confusing to me at first sight.

sudhirverma:
testPR

PR #537, opened 2019-01-21
Refactor validation in Cluster's commands

sudhirverma:


src/openshift/cluster.ts
```diff
@@ -92,9 +90,7 @@ export class Cluster extends OpenShiftItem {
             ignoreFocusOut: true,
             prompt: "Provide Username for basic authentication to the API server",
             validateInput: (value: string) => {
```
dgolovin: Should be 
``` TypeScript
validateInput: (value: string) => Cluster.emptyName('User name cannot be empty', value)
```

src/openshift/cluster.ts
```diff
@@ -73,9 +73,7 @@ export class Cluster extends OpenShiftItem {
             ignoreFocusOut: true,
             prompt: "Provide URL of the cluster to connect",
             validateInput: (value: string) => {
-                if (!validator.isURL(value)) {
-                    return 'Invalid URL provided';
-                }
+                if (!validator.isURL(value)) return 'Invalid URL provided';
```
dgolovin: Implement it the same way you did with emptyName. Move this method to `OpenShiftItem` and call it here as 
``` TypeScript
validateInput: (value: string) => Cluster.isURL('Invalid URL provided', value)
```

PR #577, opened 2019-01-30
Updated dependency

sudhirverma:


test/odo.int.ts
```diff
@@ -12,10 +12,12 @@ import { Cluster } from '../src/openshift/cluster';
 suite('odo integration', () => {
```
dgolovin: Only  changes in package.json is good to go. Everything else does not add any value .

test/odo.int.ts
```diff
@@ -12,10 +12,12 @@ import { Cluster } from '../src/openshift/cluster';
...
     const oi: odo.Odo = odo.getInstance();
     let sb: sinon.SinonSandbox;
+    let showInformationMessageStub: sinon.SinonStub;
 
     setup(() => {
         sb = sinon.createSandbox();
-        sb.stub(vscode.window, 'showInformationMessage').resolves('Download and install');
```
dgolovin: Why replace single line with two with exactly the same result?

test/openshift/application.test.ts
```diff
@@ -22,8 +22,8 @@ suite('Openshift/Application', () => {
     let sandbox: sinon.SinonSandbox;
     let execStub: sinon.SinonStub;
     let getProjectNamesStub: sinon.SinonStub;
-    const projectItem = new TestItem(null, 'project');
-    const appItem = new TestItem(projectItem, 'app');
+    const projectItem: OpenShiftObject = new TestItem(null, 'project');
```
dgolovin: No need to declare type for direct assignment, it is going to be inferred during the compilation.

dgolovin: All the changes below have the same problems.

test/odo.int.ts
sudhirverma: after updating package.json it was giving an error for tslint that why I have replaced the single line.

dgolovin: could you fix errors by configuring rules for linter?

PR #582, opened 2019-01-31
Refresh only component node afrer deleting storage

### LINKED ISSUE #581: `Create Storage` command refresh whole Explorer tree to reveal new storage item
Considering that size of the tee could be big enough and take long time to refresh it would be much better to restore original behavior when only targeted Component is refreshed to reveal new Storage item.

dgolovin:
Fix #581.


dgolovin:
@sudhirverma it was introduced with latest refactoring and should be fixed everywhere where refresh is called without parameters.

Issue #552, opened 2019-01-22
When linking a component to a service, if there is a single service, user should not been asked


jeffmaury:
When a component is linked to a service, if there is a single service available then the user shoud not been proposed with a single item combo box. The same also applies when linking a component to another component (if that case there are 2 components available but a single is being candidate)

dgolovin:
I am reverting commit for this issue. The fix solves corner case with one service, but brings usability issues for generic case when there are several services. 

The problem is that a call collecting services now placed outside of window.showQuickPicks and that means there is no feedback of any kind for the user that would show there is something going on. For big projects where collecting services might take some time it would look like nothing is happening after component was selected and then suddenly it would continue with message long list of services to select form.

There is no way around this for now, except we can select first service by default an that means there would be on click less for user to continue.

PR #609, opened 2019-02-08
When linking a component to a service, if there is a single service, user should not been asked

sudhirverma:
Fix: #552


sudhirverma:
testPR

PR #615, opened 2019-02-11
Increase project create command test coverage

dgolovin:


test/openshift/project.test.ts
```diff
@@ -66,6 +67,62 @@ suite('Openshift/Project', () => {
...
+        test('validator returns undefinded for valid project name', async () => {
+            let result: string | Thenable<string>;
+            inputStub.restore();
+            inputStub = sandbox.stub(vscode.window, 'showInputBox').onFirstCall().callsFake((options?: vscode.InputBoxOptions, token?: vscode.CancellationToken): Thenable<string> => {
+                result = options.validateInput('goodvalue');
+                return Promise.resolve('goodvalue');
+            });
+            await Project.create();
+
+            if (!isThenable(result)) {
```
jrichter1: What exactly does a check like this accomplish? If you know the type beforehand, you don't really need it. And if you don't know the type, then you need the thenable branch as well.

dgolovin: It makes sure the result is string, because 
``` javascript
validateInput?(value: string): string | undefined | null | Thenable<string | undefined | null>;
```
but you are correct, we know that it would be `undefined |  string` in our case, so we don't really needed it here.

Issue #613, opened 2019-02-09
Add validation for mountPath for storage creation.


sudhirverma:

mohitsuman:
Storage tree node is going to be removed in next iteration. Hence closing this.

PR #627, opened 2019-02-13
Add validation for mountPath for storage creation. [WIP]

sudhirverma:
Issue: #613


mohitsuman:
@sudhirverma Can you resolve the conflicts and fix this on priority.

src/openshift/openshiftItem.ts
```diff
@@ -54,6 +57,26 @@ export abstract class OpenShiftItem {
...
     }
 
+    static isAbsolute(message: string, value: string) {
+        return path.isAbsolute(value) ? null : message;
+    }
+
+    static validateFormatWindows(message: string, value: string) {
+        const pathWindowsRegex = /^[a-zA-Z]:\\(?:[^#%\\/:*?"<>|\r\n\[\]{}]+\\)*[^#%\\/:*?"<>|\r\n\[\]{}]*$/;
+        if (Platform.OS === 'win32') {
+            return pathWindowsRegex.test(value) ? null : message;
```
mohitsuman: Instead of using .test method, try to use the `validator` function.

package.json
```diff
@@ -582,6 +582,7 @@
     "hasha": "^3.0.0",
     "js-yaml": "^3.12.1",
     "opn": "^5.4.0",
+    "path": "^0.12.7",
```
jrichter1: path is in nodejs by default, don't use the npm module, especially when it's this out of date

sudhirverma: Thanks, @jrichter1 :)

src/openshift/storage.ts
```diff
@@ -21,11 +20,16 @@ export class Storage extends OpenShiftItem {
...
         if (!storageName) return null;
 
-        const mountPath = await vscode.window.showInputBox({prompt: "Specify the mount path", validateInput: (value: string) => {
-            if (validator.isEmpty(value.trim())) {
-                return 'Invalid mount path';
+        const mountPath = await vscode.window.showInputBox({
+            prompt: "Specify the mount path",
+            validateInput: (value: string) => {
+                let validationMessage = Storage.emptyName('Empty mount path', value.trim());
+                if (!validationMessage) validationMessage = Storage.isAbsolute('Invalid mount path', value.trim());
```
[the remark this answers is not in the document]
jrichter1: On the subject of unix paths - path.isAbsolute returns true for windows style paths when called on windows. That would allow windows users to pass validation without actually using a unix path.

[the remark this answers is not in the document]
dgolovin: Path should be always valid absolute unix path.

src/openshift/openshiftItem.ts
```diff
@@ -54,6 +55,10 @@ export abstract class OpenShiftItem {
         return validator.isURL(value) ? null : message;
     }
 
+    static isAbsolute(message: string, value: string) {
```
dgolovin: Create validator function for unix in addition to this one.
``` JavaScript
static isPosixAbsolute(message: string, value: string) {
   return path.posix.isAbsolute(value);
}
```
then use it in storage mount path validation.

src/openshift/storage.ts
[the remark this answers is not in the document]
dgolovin: use  isPosixAbsolute here:
```
Storage.isPosixAbsolute('Invalid mount path',  value.trim());
```

src/openshift/storage.ts
```diff
@@ -21,11 +20,14 @@ export class Storage extends OpenShiftItem {
...
 
-        const mountPath = await vscode.window.showInputBox({prompt: "Specify the mount path", validateInput: (value: string) => {
-            if (validator.isEmpty(value.trim())) {
-                return 'Invalid mount path';
+        const mountPath = await vscode.window.showInputBox({
+            prompt: "Specify the mount path",
+            validateInput: (value: string) => {
+                let validationMessage = Storage.emptyName('Empty mount path', value.trim());
+                if (!validationMessage) validationMessage = Storage.isPosixAbsolute('Invalid mount path', value.trim());
+                return validationMessage;
```
dgolovin: There should be additional validation to check if provided path uses correct characters. Now it is possible to use any characters even back slash.

dgolovin: Also it is not enough to say `Invalid mount path`, because it gives no clue why? It should say literally 'Not absolute mount path'.

Issue #598, opened 2019-02-04
Use extension default config folder to download and save tools


dgolovin:
When installed in Che resources downloaded to ~/.vs-openshift does not survive container restart, because they are not persistent. Extension should use [ExtensionContext.storagePath](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext) to save downloaded tools that it wants to be persistent.

PR #631, opened 2019-02-15
Use extension default config folder to download and save tools 

sudhirverma:
Issue: #598

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,7 +16,10 @@ import { Storage } from './openshift/storage';
 import { Url } from './openshift/url';
 import { Service } from './openshift/service';
 
+export let contextStoragePath: vscode.ExtensionContext;
+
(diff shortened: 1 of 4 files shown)


jrichter1:
This might not be a particularly great idea, since storagePath is bound to a workspace.

The first hurdle is that if no workspace is currently open, storagePath is undefined (as per https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/api/node/extHostExtensionService.ts#L114).

Second, the workspace id is actually part of the path, which would make you download everything every time the extension activates in a new workspace.

dgolovin:
@jrichter1 those are really good points. Should discuss possible solution with Che team. 
We might just include `odo` binaries into extension package.

dgolovin:
binaries now included into extension.

PR #653, opened 2019-02-26
Use Service Template name to create service

sudhirverma:


dgolovin:
Replaced with https://github.com/redhat-developer/vscode-openshift-tools/pull/654,

PR #662, opened 2019-02-28
fix project name length validation

### LINKED ISSUE #642: Name validation for odo objects should check if name is at least two characters long
When creating project 'a' `odo project create a` fails with error
`✗  ProjectRequest.project.openshift.io "b" is invalid: metadata.name: Invalid value: "b": must be at least 2 characters long`

mohitsuman:
Fix #642

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,15 +44,21 @@
       }
     },
     "@sinonjs/samsam": {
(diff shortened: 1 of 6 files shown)


sudhirverma:
why are you updating  package-lock.json? you have not added any new package for this PR

PR #666, opened 2019-03-01
add git ref for component creation

### LINKED ISSUE #590: Support `reference` option for creating components from git repository
Support odo https://github.com/redhat-developer/odo/pull/1042

Support providing [git reference](https://git-scm.com/book/en/v2/Git-Internals-Git-References) as a value for `--ref` or `-r` option for creating components from git repository.

mohitsuman:
Fix #590

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -248,6 +248,14 @@ export class Component extends OpenShiftItem {
 
         if (!repoURI) return null;
 
+        const gitRef = await vscode.window.showInputBox({
+            value: 'master',
+            prompt: 'Specify git reference (branch/tags/commits | default: master)',
+            validateInput: (value: string) => {
+                if (!value.trim()) return 'Empty reference, specify master if no reference needed.';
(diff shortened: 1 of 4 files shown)


src/openshift/component.ts
```diff
@@ -248,6 +248,14 @@ export class Component extends OpenShiftItem {
 
         if (!repoURI) return null;
 
+        const gitRef = await vscode.window.showInputBox({
+            value: 'master',
+            prompt: 'Specify git reference (branch/tags/commits | default: master)',
+            validateInput: (value: string) => {
+                if (!value.trim()) return 'Empty reference, specify master if no reference needed.';
```
dgolovin: should accept empty value, which would mean 'master'.

dgolovin: Add tests for validation. See example here https://github.com/redhat-developer/vscode-openshift-tools/blob/master/test/openshift/component.test.ts#L212.

Issue #668, opened 2019-03-02
Split `OpenShift: New Component' command in three: 'New Component from local folder', 'New Component from Git repository' and 'New Component form binary'


dgolovin:
That would let us to get rid of extra step form original action and let user see in command palette what we can use to create new Component

dgolovin:
Resolved with PR [#672](https://github.com/redhat-developer/vscode-openshift-tools/pull/672).

PR #672, opened 2019-03-04
Split `OpenShift: New Component' command in three: 'New Component from local folder', 'New Component from Git repository' and 'New Component form binary'

sudhirverma:
Issue: #668


package.json
```diff
@@ -130,8 +132,18 @@
         "category": "OpenShift"
       },
       {
-        "command": "openshift.component.create",
-        "title": "New Component",
+        "command": "openshift.component.create.local",
```
dgolovin: this looks good.

package.json
```diff
@@ -130,8 +132,18 @@
         "category": "OpenShift"
       },
       {
-        "command": "openshift.component.create",
```
dgolovin: let's keep this command and make it visible only in context menu for Application node.

Issue #669, opened 2019-03-02
Split 'OpenShift: Login into Cluster' to two commands


dgolovin:
Two commands would be available from command palette:
* 'OpenShift: Login into Cluster with credentials'
* 'OpenShift: Login into Cluster with token' 

Original command should be changed. 
* It should ask about the way to login to the cluster as a first step, then execute command specific to selected way.
* It should be available through context menu and the view title.

dgolovin:
Resolved with [PR #681](https://github.com/redhat-developer/vscode-openshift-tools/pull/681) .

PR #681, opened 2019-03-06
Split 'OpenShift: Login into Cluster' to two commands

sudhirverma:
Issue: #669

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -16,7 +16,9 @@ export class Cluster extends OpenShiftItem {
             return Cluster.loginDialog();
         } else {
             const value = await vscode.window.showInformationMessage(`You are already logged in the cluster. Do you want to login to a different cluster?`, 'Yes', 'No');
-            if (value === 'Yes') return Cluster.loginDialog();
+            if (value === 'Yes') {
+                return Cluster.loginDialog();
+            }
             return null;
         }
     }
@@ -65,7 +67,7 @@ export class Cluster extends OpenShiftItem {
         }
     }
 
-    static async getUrl() {
+    static async getUrl(): Promise<string> {
         return await vscode.window.showInputBox({
             ignoreFocusOut: true,
             prompt: "Provide URL of the cluster to connect",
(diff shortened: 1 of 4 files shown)


src/openshift/cluster.ts
```diff
@@ -67,12 +65,16 @@ export class Cluster extends OpenShiftItem {
...
+    static async getUrl() {
+        return await vscode.window.showInputBox({
             ignoreFocusOut: true,
             prompt: "Provide URL of the cluster to connect",
             validateInput: (value: string) => Cluster.validateUrl('Invalid URL provided', value)
         });
+    }
+
+    private static async loginDialog(): Promise<string> {
+        const clusterURL = await Cluster.getUrl();
```
dgolovin: Should work the same way we have done for `Create Component` command.  First it should ask for login method, then call specific commands to login with credentials or with token.
`const clusterURL = await Cluster.getUrl();` call should be part for those commands.

sudhirverma: I think it's  better to keep `const clusterURL = await Cluster.getUrl();` here only because if I call credentialsLogin/tokenLogin command from the `log in icon view/command palette` then I don't have any condition to check whether I am running from command palette or not. https://github.com/redhat-developer/vscode-openshift-tools/pull/681/files#diff-acaf210d416301d185dd5622c49f9093R89 will run twice if I try to login from the icon view and I want to show this first when the user tries to login from the login icon view.

dgolovin: let's say we have two different commands. One does login with credentials and another with token.
Each command would do:
* check for user being logged in
* ask for cluster URL
* ask credentials for specific login type: username/password or token.
Each command would have a parameter `do not check login status'.

When you run from command palette we pass values for parameter 'do not check login status' and in this case each command supposed to run check.

Now we need command that asks what login do you want to use username/password or token. It would do:
* check for login status and ask confirmation for login to different cluster
* show quick picks for username/password and token logins
* call specific login function directly with 'do not check login status' parameter value set to `true'

That would be much cleaner workflow.

PR #720, opened 2019-03-19
Read error message from stderr stream in getServiceTemplates() method

### LINKED ISSUE #719: 'New Service' command fails silently with latest odo
It happens, because error message is printed to stderr stream instead of stdout.

dgolovin:
Fix #719.


dgolovin:
testPR

PR #734, opened 2019-03-21
Add tags/branch to git reference quickpick

### LINKED ISSUE #710: Add tags + branch list for a git repo when selecting the ref within QuickPick item

mohitsuman:
Fix #710

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -12,6 +12,7 @@ import { ChildProcess } from 'child_process';
 import { CliExitData } from '../cli';
 import { V1ServicePort, V1Service } from '@kubernetes/client-node';
 import { isURL } from 'validator';
+import { Refs } from '../util/refs';
 
 export class Component extends OpenShiftItem {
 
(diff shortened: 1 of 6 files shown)


src/util/refs.ts
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
dgolovin: There should be original copyright header and I believe we should put link to original repository.

src/openshift/component.ts
```diff
@@ -261,13 +261,8 @@ export class Component extends OpenShiftItem {
...
 
-        const gitRef = await vscode.window.showInputBox({
-            value: 'master',
-            prompt: 'Specify git reference (branch/tags/commits | default: master)',
-            validateInput: (value: string) => {
-                if (!value.trim()) return 'Empty reference, specify master if no reference needed.';
-            }
-        });
+        const references = await Refs.fetchTag(repoURI);
+        const gitRef = await vscode.window.showQuickPick([ ...references.keys()], {placeHolder: "Specify git reference (branch/tags)"});
```
dgolovin: Should be `Select git reference (branch/tag)`.

src/util/refs.ts
mohitsuman: Updated.

src/openshift/component.ts
mohitsuman: Fixed.

PR #741, opened 2019-03-26
Fix open url with multiple routes and add custom name to route

### LINKED ISSUE #728: Use 'odo url list -o json` to fetch list of URLs for component
Rather than using oc command to fetch host name and then figure out protocol based on TLS flag, `odo url list -o json` can be used to simplify 'Open in Browser' command.

```odo url list --component nodejs-ex-nodejs-exra --app app1 -o json```
``` json
{
    "kind": "List",
    "apiVersion": "odo.openshift.io/v1alpha1",
    "metadata": {},
    "items": [
        {
            "kind": "url",
            "apiVersion": "odo.openshift.io/v1alpha1",
            "metadata": {
                "name": "nodejs-ex-nodejs-exra-8080",
                "creationTimestamp": null
            },
            "spec": {
                "path": "nodejs-ex-nodejs-exra-8080-app1-myproject.10.0.0.46.nip.io",
                "protocol": "http",
                "port": 8080
            }
        }
    ]
}
```

that would eliminate two oc commands we use:
1. https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L97
2. https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L99

mohitsuman:
Fix #728 

Output Screen:
Screenshot 2019-03-26 at 9 49 10 PM

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -95,8 +95,6 @@ export const Command = {
         `odo watch ${component} --app ${app} --project ${project}`,
     getRouteHostName: (namespace: string, component: string) =>
         `oc get route --namespace ${namespace} -o jsonpath="{range .items[?(.metadata.labels.app\\.kubernetes\\.io/component-name=='${component}')]}{.spec.host}{end}"`,
-    getRouteTls: (namespace: string, component: string) =>
-        `oc get route --namespace ${namespace} -o jsonpath="{range .items[?(.metadata.labels.app\\.kubernetes\\.io/component-name=='${component}')]}{.spec.tls.termination}{end}"`,
     createLocalComponent: (project: string, app: string, type: string, version: string, name: string, folder: string) =>
         `odo create ${type}:${version} ${name} --local ${folder} --app ${app} --project ${project}`,
     createGitComponent: (project: string, app: string, type: string, version: string, name: string, git: string, ref: string) =>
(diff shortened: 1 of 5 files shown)


mohitsuman:
Currently the unit tests are broken. Need to be fixed.

src/odo.ts
```diff
@@ -111,8 +111,10 @@ export const Command = {
         `oc get ServiceInstance ${service} --namespace ${project} -o jsonpath="{$.metadata.labels.app\\.kubernetes\\.io/component-type}"`,
     waitForServiceToBeGone: (project: string, service: string) =>
         `oc wait ServiceInstance/${service} --for delete --namespace ${project}`,
-    createComponentUrl: (project: string, app: string, component: string, port: string) =>
-        `odo url create --port ${port} --project ${project} --app ${app} --component ${component}`,
+    createComponentCustomUrl: (project: string, app: string, component: string, name: string, port: string) =>
+        `odo url create ${name} --port ${port} --project ${project} --app ${app} --component ${component}`,
+    getComponentUrl: (app: string, component: string) =>
+        `odo url list --component ${component} --app ${app} -o json`,
```
dgolovin: should have `--project` option as well.

mohitsuman: Fixed

Issue #567, opened 2019-01-28
Add new login method from oc command line


jeffmaury:
It should be great if we could login to a cluster using an oc command line parsing like in JBoss Tools.
When you can't connect to a cluster using username and password, we could reuse the oc command line that is copied to the clipboard from the OpenShift console and deduce the cluster URL and the user token from that command line.

dgolovin:
API to read clipboard https://code.visualstudio.com/api/references/vscode-api#Clipboard.
Format for `oc login`
```
oc login https://10.0.0.45:8443 --token=x9meaJYA1ZigygLlhpChaQgEOcVPhv3hfBo4QSHfcfk
```

dgolovin:
OpenShift login commands should analyse content through API and if it matches the pattern deduce url and token. Then use both for token login or url only for credentials.

dgolovin:
Fixed with https://github.com/redhat-developer/vscode-openshift-tools/commit/b6ed250055b011aaff0dd17e9100b8c4fdd7d1bc.

PR #745, opened 2019-03-27
Add new login method from oc command line

sudhirverma:
Issue: #567


src/openshift/cluster.ts
```diff
@@ -127,19 +128,27 @@ export class Cluster extends OpenShiftItem {
             .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${error}'!`));
     }
 
+    static async copyToClipboard(url: string, token: string) {
+        return await copyPaste.copy({clusterURL: url, ocToken: token});
```
mohitsuman: @sudhirverma Why are we copying the url and token here. The purpose here is that the user need not have the token and can still login to the cluster based on the oc cli parameters.

package.json
```diff
@@ -627,6 +627,7 @@
   "dependencies": {
     "@kubernetes/client-node": "^0.8.1",
     "byline": "^5.0.0",
+    "copy-paste": "^1.3.0",
```
mohitsuman: I think we can use a better npm module than this which is maintained or has more downloads atleast.

sudhirverma:
I can get an access token write in ~/.kube/config file by using `oc whoami -t` command and I can use that token for login like `oc login clusterURL --token=userToken`. I have not found the way to get the token from clipboard from the OpenShift console UI.

jeffmaury:
Login to your cluster using the web console, then click on your profile (top right) and select `Copy Login Command`:

image

and the `oc login command` will be copied to the clipboard.

package.json
```diff
@@ -628,6 +634,7 @@
   "dependencies": {
     "@kubernetes/client-node": "^0.8.1",
     "byline": "^5.0.0",
+    "clipboardy": "^1.2.3",
```
jeffmaury: Why don't you use the VSCode API as mentioned by @dgolovin

src/openshift/cluster.ts
```diff
@@ -71,12 +72,14 @@ export class Cluster extends OpenShiftItem {
     static async login(): Promise<string> {
         const response = await Cluster.requestLoginConfirmation();
         if (response !== 'Yes') return null;
-        const loginMethod = await window.showQuickPick(['Credentials', 'Token'], {placeHolder: 'Select the way to log in to the cluster.'});
+        const loginMethod = await window.showQuickPick(['Credentials', 'Token', 'ocLogin'], {placeHolder: 'Select the way to log in to the cluster.'});
```
jeffmaury: I think we should have a more explicit label here

src/openshift/cluster.ts
```diff
@@ -143,6 +146,23 @@ export class Cluster extends OpenShiftItem {
             .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${error}'!`));
     }
 
+    static async readFromClipboard() {
+        return await clipboardy.readSync();
+    }
+
+    static async ocLogin(skipConfirmation: boolean = false): Promise<string> {
+        const response = await Cluster.requestLoginConfirmation(skipConfirmation);
```
jeffmaury: Why are you calling it again when it's already called in login method ?

src/openshift/cluster.ts
```diff
@@ -143,6 +146,23 @@ export class Cluster extends OpenShiftItem {
...
+        if (response !== 'Yes') return null;
+        const oclogin = Cluster.ocLoginCommandMatches(await clipboardy.readSync());
```
jeffmaury: You wrote a method to read from clipboard, why don't you call it ?

src/openshift/cluster.ts
```diff
@@ -143,6 +146,23 @@ export class Cluster extends OpenShiftItem {
...
+        if (!oclogin) throw Error('oc login command not found in clipboard');
+        const clusterURL = Cluster.clusterURL(oclogin);
+        if (!clusterURL) return null;
+        return Promise.resolve()
```
jeffmaury: I don't understand how it works cause you're not extracting the token but only the cluster URL, something is missing here

jeffmaury: OK I got it as you're running the whole oc command line. See with @dgolovin as I think we try to lower our dependency on `oc` binary

package.json
dgolovin: @sudhirverma same question here. VSCode already has API for accessing clipboard, no need to use additional modules.

src/openshift/cluster.ts
dgolovin: ... and it is not ocLogin, it's still the same token login, we just reading some of the data, cluster url and token, form clipboard.

sudhirverma: @dgolovin and @jeffmaury  should I keep only token login option only (instead of oclogin) and if a token is detected from clipboard then I will show Information message that token is detected if user select `yes` then it will automatically login and if user select `No` then I will redirect to manual login throw url and token(https://github.com/redhat-developer/vscode-openshift-tools/blob/master/src/openshift/cluster.ts#L130)

src/openshift/cluster.ts
sudhirverma: I am calling requestLoginConfirmation() method here for command palette to check the login status.

package.json
sudhirverma: okay I will use the VSCode API for accessing clipboard

dgolovin:
Another option is use https://code.visualstudio.com/api/references/vscode-api#window onDidWindowStateChange event and check clipboard. If cluster and  token are detected show request to login to cluster with token detected in clipboard.

sudhirverma:
@dgolovin please review this pr

sudhirverma:
test PR

sudhirverma:
testPR

sudhirverma:
testPR

sudhirverma:
testPR

sudhirverma:
testPR

sudhirverma:
testPR

sudhirverma:
testPR

src/openshift/cluster.ts
```diff
@@ -61,7 +61,9 @@ export class Cluster extends OpenShiftItem {
     }
 
     static async getUrl(): Promise<string | null> {
+        const clusterURl = await Cluster.getUrlFromClipboard();
```
dgolovin: You should use VSCode API for reading data from clipboard https://code.visualstudio.com/api/references/vscode-api#Clipboard..

sudhirverma: I am using VSCode API for reading data from clipboard.

PR #763, opened 2019-04-09
Fix revealing objects

### LINKED ISSUE #762: New created object is not selected in OpenShift Application Explorer when parent object expanded and tree is not visible
This could be upstream issue in VSCode.

dgolovin:
Fix #762. 
Workaround is calling reveal second time, to make newly created object selected.


dgolovin:
testPR

Issue #754, opened 2019-04-03
Add verbose log option for operation like push


amitkrout:
It will be helpfull to debug failure like odo push and many more operation like app create, component create, watch. 

For example :

odo push fuse --app fuse-rest --project myproject -v 9

See more for verbose level in odo
```
-v, --v Level                          log level for V logs. Level varies from 0 to 9 (default 0).
```

dgolovin:
Implementation for first iteration is fairly simple. 

To set up verbosity level use standard settings editor. It is accessible through bar menu ```File->Preferences->Settings``` or ```Ctrl+,``` keys combination.  Find ```Openshift Connector: Output Verbosity Level``` and set integer value between 0 and 9. This will add additional ```-v <number>``` option to all create, push and watch commands.

To preserve output more than 1000 lines update Terminal Scrollback to desired number, default is 1000 lines.

dgolovin:
PR #816  is released to master. For additional improvement please create new issues.

PR #771, opened 2019-04-09
Add verbose log in push command 

sudhirverma:
Issue: #754

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -93,10 +93,10 @@ export const Command = {
         `oc get service ${component}-${app} --namespace ${project} -o jsonpath="{range .spec.ports[*]}{.port}{','}{end}"`,
     linkComponentTo: (project: string, app: string, component: string, componentToLink: string, port?: string) =>
         `odo project set ${project} && odo application set ${app} && odo component set ${component} && odo link ${componentToLink} --wait${port ? ' --port ' + port : ''}`,
-    pushComponent: (project: string, app: string, component: string) =>
-        `odo push ${component} --app ${app} --project ${project} -v 9`,
-    pushLocalComponent: (project: string, app: string, component: string, location: string) =>
-        `${Command.pushComponent(project, app, component)} --local ${location}`,
+    pushComponent: (project: string, app: string, component: string, level: number) =>
+        `odo push ${component} --app ${app} --project ${project} -v ${level}`,
+    pushLocalComponent: (project: string, app: string, component: string, location: string, level: number) =>
+        `${Command.pushComponent(project, app, component, level)} --local ${location}`,
     watchComponent: (project: string, app: string, component: string) =>
         `odo watch ${component} --app ${app} --project ${project}`,
     getRouteHostName: (namespace: string, component: string) =>
@@ -241,7 +241,7 @@ export interface Odo {
     createApplication(project: OpenShiftObject, name: string): Promise<OpenShiftObject>;
     deleteApplication(application: OpenShiftObject): Promise<OpenShiftObject>;
     createComponentFromGit(application: OpenShiftObject, type: string, version: string, name: string, repoUri: string, ref: string): Promise<OpenShiftObject>;
-    createComponentFromFolder(application: OpenShiftObject, type: string, version: string, name: string, path: string): Promise<OpenShiftObject>;
+    createComponentFromFolder(application: OpenShiftObject, type: string, version: string, name: string, path: string, level: number): Promise<OpenShiftObject>;
     createComponentFromBinary(application: OpenShiftObject, type: string, version: string, name: string, path: string): Promise<OpenShiftObject>;
     deleteComponent(component: OpenShiftObject): Promise<OpenShiftObject>;
     createStorage(component: OpenShiftObject, name: string, mountPath: string, size: string): Promise<OpenShiftObject>;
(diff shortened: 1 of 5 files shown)


src/odo.ts
```diff
@@ -93,7 +93,7 @@ export const Command = {
     linkComponentTo: (project: string, app: string, component: string, componentToLink: string, port?: string) =>
         `odo project set ${project} && odo application set ${app} && odo component set ${component} && odo link ${componentToLink} --wait${port ? ' --port ' + port : ''}`,
     pushComponent: (project: string, app: string, component: string) =>
-        `odo push ${component} --app ${app} --project ${project}`,
+        `odo push ${component} --app ${app} --project ${project} -v 9`,
```
dgolovin: It should be configurable not just hard coded. 
Options:
* configuration settings
   - verbose output level - could be integer 0 - no verbose output, > 0 - verbosity level
* command `OpenShift: Set log verbosity level` that would request verbosity level from user and write it to settings (if that possible)
* after first `OpenShift: Push` ask if dev want to increase verbosity level with options: Yes, No, Do not show again.  Yes would call command above.

sudhirverma: @dgolovin there should be three option `Yes, No, Do not show again` or just `Yes, No` option?. what if I want to change the verbose level. I think we can skip the `Do not show again` part. Every time when the user uses the `push` command we will show the information message that he/she wants to change the verbose level.

mohitsuman:
> * Create `Verbosity Level` setting with workspace scope defaulted to 0
> * Add `OpenShift: Set Output Verbosity Level` command that would request number up to 12 (because search for `glog.V(13)`  or higher does not return any results)

Let's keep the verbose level from **0-9**, this is what is supported by `odo`.

 `-v, --v Level                          log level for V logs. Level varies from 0 to 9 (default 0).`

src/openshift/component.ts
```diff
@@ -150,13 +150,44 @@ export class Component extends OpenShiftItem {
         );
     }
 
+    static async verbosityLevel() {
```
dgolovin: Make it a `const VERBOSITY_LEVEL = [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9];`

src/openshift/component.ts
```diff
@@ -150,13 +150,44 @@ export class Component extends OpenShiftItem {
...
+            level.push(i.toString());
+        }
+        return level;
+    }
+
+    static async showInformationMessage() {
+        return await window.showInformationMessage(`Do you want change verbosity level (Default value 0)`, 'Yes', 'No');
+    }
+
+    static async getVerbosityLevel() {
```
dgolovin:
``` TypeScript
return (await Component.showInformationMessage() === 'Yes') && (await window.showQuickPick(Component.VERBOCITY_LEVEL, {placeHolder: "Select verbosity level"}) || 0;`
```
Or if you like more verbose and not so wide code (have not tested it though):
``` TypeScript
const option = await Component.showInformationMessage();
const level = option === 'Yes' && await window.showQuickPick(Component.VERBOCITY_LEVEL, {placeHolder: "Select verbosity level"});
return level || 0;
```

src/openshift/component.ts
```diff
@@ -150,13 +150,44 @@ export class Component extends OpenShiftItem {
...
+        if (await Component.showInformationMessage() === 'Yes') {
+            const level = await window.showQuickPick(Component.verbosityLevel(), {placeHolder: "Select verbosity level"});
+            return level ? level : 0;
+        }
+        return 0;
+    }
+
+    static async verbosity(context: OpenShiftObject): Promise<string> {
+        const component = await Component.getOpenShiftCmdData(context,
```
dgolovin: There is no reason to ask about Project/Application/Component. In first implementation we just set verbosity level for vscode workspace. That will affect all the commands.

src/odo.ts
```diff
@@ -92,10 +92,10 @@ export const Command = {
         `oc get service ${component}-${app} --namespace ${project} -o jsonpath="{range .spec.ports[*]}{.port}{','}{end}"`,
     linkComponentTo: (project: string, app: string, component: string, componentToLink: string, port?: string) =>
         `odo project set ${project} && odo application set ${app} && odo component set ${component} && odo link ${componentToLink} --wait${port ? ' --port ' + port : ''}`,
-    pushComponent: (project: string, app: string, component: string) =>
-        `odo push ${component} --app ${app} --project ${project}`,
-    pushLocalComponent: (project: string, app: string, component: string, location: string) =>
-        `${Command.pushComponent(project, app, component)} --local ${location}`,
+    pushComponent: (project: string, app: string, component: string, level: number) =>
+        `odo push ${component} --app ${app} --project ${project} -v ${level}`,
```
dgolovin: rather than changing every factory method and adding level parameter, lets just get verbosity level directly from workspace settings here
``` typescript
`odo push ${component} --app ${app} --project ${project} -v ${Odo.VERBOCITY}`,
```

``` typescript
public get VERBOCITY() {
    return workspace.getConfiguration(........);
}
```

See docs for configuration API [here](https://code.visualstudio.com/api/references/vscode-api#workspace).

dgolovin:
Replaced with https://github.com/redhat-developer/vscode-openshift-tools/pull/816.

Issue #793, opened 2019-04-25
Upgrade odo to v1.0.0 beta release


apupier:
new release of odo is available https://github.com/openshift/odo/releases/tag/v1.0.0-beta3

it noticeably contains fix to have watch functionality working on windows (even if with some caveats https://github.com/openshift/odo/issues/1630)

apupier:
1.0.0-beta4 has been released https://github.com/openshift/odo/releases/tag/v1.0.0-beta4

mohitsuman:
We will be migrating to odo-beta-5 release.

PR #794, opened 2019-04-25
Upgrade to odo v1.0.0-beta1 (#793)

apupier:
tested only download of odo binary on windows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.0.20
+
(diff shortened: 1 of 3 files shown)


mohitsuman:
There are lots of change with `odo 0.20` and `odo 1.0-beta` release based on the odo workflow. We need to see how the integration needs to be done and what changes have to be made on adapters side for the same.

dgolovin:
Replaced with https://github.com/redhat-developer/vscode-openshift-tools/pull/847.l

PR #798, opened 2019-04-26
Add url routes and delete route from tree view

### LINKED ISSUE #781: urls node and action
Now that we can have several URLs per component, I think it would be nice to see them in the tree so that we can delete it later at least

mohitsuman:
Fix #781 #797 

1. Shows all the routes present in the tree view
2. Allows to delete the route individually from the tree view
3. Perform delete route operation using command palette

Screenshot 2019-04-26 at 6 39 11 AM


jeffmaury:
I used a globe icon in the IJ plugin: https://github.com/redhat-developer/intellij-openshift-connector/blob/master/src/main/resources/images/url-node.png
Seems more representing a navigation. WDYT ?

mohitsuman:
> We should find out the way to show info about URL:
> 
> * Hostname
> * Target Port
> 
> child nodes with property: value as label would work this purpose:
> 
> ```
> component
>  |- URLname
>     |- Hostname: http://whatever-name-is-generated.com
>     |- Target Port: 8080
> ```

@dgolovin I have created a new issue to tackle this scenario as a separate one https://github.com/redhat-developer/vscode-openshift-tools/issues/800

mohitsuman:
> I used a globe icon in the IJ plugin: https://github.com/redhat-developer/intellij-openshift-connector/blob/master/src/main/resources/images/url-node.png
> Seems more representing a navigation. WDYT ?

@jeffmaury I tried out the globe icon, but I think this new `link icon` might give the UX a better feel. Here is the new screenshot. Let me know if this looks consistent.

Screenshot 2019-04-26 at 9 48 21 PM

PR #807, opened 2019-04-28
add webpack configuration

### LINKED ISSUE #806: Cofigure webpack build development

mohitsuman:
Fix #806

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+//@ts-check
+
+'use strict';
+
+const path = require('path');
+
+/**@type {import('webpack').Configuration}*/
+const config = {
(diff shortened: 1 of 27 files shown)


webpack.config.js
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
```
jeffmaury: Microsoft ?

mohitsuman: Yeah the webpack configuration as shown in the official documentation for vscode extension has the Microsoft license and even vscode-java uses the same. So I preferred the same.

mohitsuman:
> What's the exact purpose of this change ? What's the origin of this ? It's not clear at all for novices

Currently when we build the vscode extension for the release, it prompts us to use the webpack build as the size of the extension and no of files have increased. So we will be using webpack to compile these multiple small files into one big file and bundle them.

webpack is the preferred bundler for javascript & typescript application.
https://code.visualstudio.com/api/working-with-extensions/bundling-extension

jeffmaury:
Unit test are failing, we need to fix that

mohitsuman:
> It is not a priority for extension with not so many files.

When we do a build for release, it prompts to do so. It is a good use case for the future coming releases.

mohitsuman:
Will fix this in next iteration, hence closing the PR.

Issue #820, opened 2019-04-30
Provide keybinding for all the commands


sudhirverma:
Key bindings will provide shortcut keys for Login, create, delete and etc.
for e.g:
```
Login: command/ctrl+shift+l
refresh: command/ctrl+shift+r
project
 - create: command/ctrl+shift+c+p
 - delete: command/ctrl+shift+d+p
appliction:
 - create: command/ctrl+shift+c+a
 - delete: command/ctrl+shift+d+a
```
will perform the action depend on the key press.

jeffmaury:
I noticed we tend to be too much concise when adding issue, so can you add a description and also propose the key bindings you plan to implement

dgolovin:
There is no need to have key binding for every single command though.
I would suggest to add it for Push at least.

PR #835, opened 2019-05-02
Provide keybinding for Login, push and refresh commands

sudhirverma:
Fix: #820

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "onCommand:openshift.component.followLog.palette",
     "onCommand:openshift.component.delete",
     "onCommand:openshift.component.push",
+    "onCommand:openshift.component.last.push",
     "onCommand:openshift.component.push.palette",
     "onCommand:openshift.component.watch",
     "onCommand:openshift.component.watch.palette",
@@ -249,6 +250,11 @@
         "category": "OpenShift"
       },
       {
+        "command": "openshift.component.last.push",
+        "title": "Repeat last push command",
+        "category": "OpenShift"
+      },
+      {
         "command": "openshift.component.watch",
         "title": "Watch",
         "category": "OpenShift"
@@ -357,7 +363,7 @@
         "mac": "ctrl+shift+r"
       },
       {
-        "command": "openshift.component.push",
+        "command": "openshift.component.last.push",
         "key": "alt+shift+p",
         "mac": "ctrl+shift+p"
       }
--- a/package.json
+++ b/package.json
@@ -348,18 +348,18 @@
     "keybindings": [
       {
         "command": "openshift.explorer.login",
-        "key": "ctrl+shift+l",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -339,6 +339,23 @@
...
       }
     ],
+    "keybindings": [
+      {
+        "command": "openshift.explorer.login",
+        "key": "ctrl+shift+l",
+        "mac": "cmd+shift+l"
+      },
+      {
+        "command": "openshift.explorer.refresh",
```
dgolovin: @sudhirverma  could you check if it is possible to activate this only for OpenShift Views Container? Otherwise the result of this command called with key combination would not be visible to the user.

package.json
```diff
@@ -339,6 +339,23 @@
         "category": "OpenShift"
       }
     ],
+    "keybindings": [
+      {
+        "command": "openshift.explorer.login",
+        "key": "ctrl+shift+l",
```
dgolovin: This conflicts with key combination to show occurrences in active editor.

src/openshift/component.ts
```diff
@@ -148,13 +149,27 @@ export class Component extends OpenShiftItem {
...
+
     static async push(context: OpenShiftObject): Promise<string> {
-        const component = await Component.getOpenShiftCmdData(context,
-            "In which Project you want to push the changes",
-            "In which Application you want to push the changes",
-            "For which Component you want to push the changes");
-        if (!component) return null;
-        Component.odo.executeInTerminal(Command.pushComponent(component.getParent().getParent().getName(), component.getParent().getName(), component.getName()));
+        const getPushCmd = await Component.getPushCmd();
+        if (getPushCmd && !context) {
```
dgolovin: With this check in place there is no way to use push command from command palette. After first push it will always push the same component until context menu for component in tree is used.
You need to create new command 'OpenShift: Repeat last push command'.

src/openshift/component.ts
```diff
@@ -148,13 +149,27 @@ export class Component extends OpenShiftItem {
...
+            Component.odo.executeInTerminal(getPushCmd);
+        } else {
+            const component = await Component.getOpenShiftCmdData(context,
+                "In which Project you want to push the changes",
+                "In which Application you want to push the changes",
+                "For which Component you want to push the changes");
+            if (!component) return null;
+            Component.setPushCmd(component.getName(), component.getParent().getName(), component.getParent().getParent().getName());
```
dgolovin: You should be just saving 
``` JavaScript
() => {
    Component.odo.executeInTerminal(Command.pushComponent(component.getParent().getParent().getName(), component.getParent().getName(), component.getName()));
}
```
here.

dgolovin:
> Issue: #820

@sudhirverma please use: Fix #820 description instead, so merging the PR would automatically close the issue.

PR #836, opened 2019-05-08
Verify during the build that tools download info is correct.

sudhirverma:


build/verify-tools.ts
```diff
@@ -0,0 +1,93 @@
...
+const path = require("path");
+const readline = require('readline');
+const progress = require('request-progress');
+const platform = require('../out/src/util/platform');
+const configData = require('../requirements.json');
+
+async function verifyTools() {
+  for (const key in configData) {
+    for (const os in configData[key].platform) {
+      const toolLocations = path.resolve(__dirname, '..', 'cache', os, configData[key].platform[os].dlFileName);
```
dgolovin: Use OS specific temp folder for download. There is no need to save files. We are not including them into extension, just checking if download info is correct.

build/verify-tools.ts
```diff
@@ -0,0 +1,93 @@
...
+      const targetFolder =  path.resolve(__dirname, '..', 'cache', os);
+      if (fs.existsSync(toolLocations)) {
```
dgolovin: No need to check if file exists, always download and verify checksum. 
Throw exception if  checksum for downloaded file is not matching with one from download configuration.

requirements.json
```diff
@@ -0,0 +1,61 @@
+{
```
dgolovin: Name file ```tools.json``` ang put next to ```tools.ts```.

src/tools.ts
```diff
@@ -16,67 +16,7 @@ import * as fs from 'fs';
...
-                url: "https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz",
-                sha256sum: "4b0f07428ba854174c58d2e38287e5402964c9a9355f6c359d1242efd0990da3",
-                fileName: "oc.tar.gz",
-                dlFileName: "oc.tar.gz",
-                filePrefix: "openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit"
-            }
-        }
-    }
-};
+const configData = require('../../requirements.json');
```
dgolovin: This would be loading ```./tools.js```.

package.json
```diff
@@ -618,6 +618,7 @@
   },
   "extensionDependencies": [],
   "scripts": {
+    "package": "node ./build/verify-tools.ts",
```
dgolovin: Why it is called ```package``` BTW? Name script ```verify```, because it is exactly what it does.

dgolovin:
testPR

sudhirverma:
testPR

sudhirverma:
@dgolovin please review this PR

build/verify-tools.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!fs.existsSync(targetFolder)) {
+    await mkdirp.sync(targetFolder);
+  }
+  const currentFile = path.join(targetFolder, fileName);
+  await downloadFile(reqURL, currentFile);
+  const currentSHA256 = await getSHA256(currentFile);
+  if (currentSHA256 === sha256sum) {
+    console.log( `[INFO] ${currentFile} is downloaded and sha256 is correct`);
+  } else {
+    throw console.log(`[Error] ${currentFile} is downloaded and sha256 is not correct`);
```
dgolovin: This will produce 
```
npm run verify

> vscode-openshift-connector@0.0.22 verify c:\p\vscost
> node ./build/verify-tools.ts

[INFO] odo-windows-amd64.exe.gz is not downloaded yet
[INFO] C:\Users\dgolo\AppData\Local\Temp\win32\odo-windows-amd64.exe.gz download started from https://github.com/redhat-developer/odo/releases/download/v0.0.20/odo-windows-amd64.exe.gz
0%
1%
9%
50%
99%
100%
[Error] C:\Users\dgolo\AppData\Local\Temp\win32\odo-windows-amd64.exe.gz is downloaded and sha256 is not correct
(node:1988) UnhandledPromiseRejectionWarning: undefined
```

Instead of use console.log, just use ```throw Error('error message')```.

build/verify-tools.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+      console.log('100%');
+    })
+    .pipe(fs.createWriteStream(toFile))
+    .on('close', resolve)
+    .on('error', reject);
+  });
+}
+
+verifyTools();
```
dgolovin: before running verifyTools() lets do
```
git diff --name-only master..HEAD
```

check if tools.json file is listed in output and skip check if it is not there.

package.json
```diff
@@ -640,7 +641,7 @@
     "test": "npm run clean && npm run compile && node ./node_modules/vscode/bin/test",
     "update-deps": "node_modules/.bin/ncu --upgrade --loglevel verbose --packageFile package.json && npm update",
     "coverage:upload": "codecov -f coverage/coverage-final.json",
-    "build": "npm run clean && tslint -p tslint.json && npm run compile"
+    "build": "npm run clean && tslint -p tslint.json && npm run compile && npm run verify"
```
dgolovin: Let's run ```npm run verify``` in test script instead so it would be called only in PR check builds.

PR #841, opened 2019-05-10
 Use odo -o json option when getting list of projects

### LINKED ISSUE #721: Use `odo project list -o json` to get list of projects
It should replace current implementation with `oc` here https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L33
Upstream issue https://github.com/redhat-developer/odo/issues/1361.

dgolovin:
Fix #721.


dgolovin:
Included into 9de60c2.

PR #887, opened 2019-06-06
Remove Create Application command

### LINKED ISSUE #886: OpenShift: Create Application does not supported in odo starting version 1.0.0-beta1
OpenShift: Create Application command should be removed, because odo has no corresponding command anymore.

dgolovin:
FIx #886.


jrichter1:
This looks like applications are basically obsolete now. I believe that is going to require a major overhaul to the rest of the logic to work properly.
That being said, can we handle the odo 1.0 migration in a feature branch with all the breaking changes?

dgolovin:
Released as part of 9de60c2414253274576a85dfb2d78c1ffa057f16.

PR #908, opened 2019-06-26
Add description to the login actions

### LINKED ISSUE #907: Add description to quickPick for login selections
This will add the `label` and `description` to the QuickPick item while selecting the logic selection for `Credentials` and `Token`.

mohitsuman:
Fix #907


src/openshift/cluster.ts
```diff
@@ -72,14 +72,20 @@ export class Cluster extends OpenShiftItem {
...
     static async login(): Promise<string> {
         const response = await Cluster.requestLoginConfirmation();
+        const loginActions = [
+            {
+                label: 'Credentials',
+                description: 'Log in to the given server using credentials'
+            },
+            {
+                label: 'Token',
+                description: 'Login using bearer token for authentication to the API server'
```
dgolovin: What if we change it to `Log in to the given server using bearer token` for consistency?

mohitsuman: Fixed.

PR #910, opened 2019-06-27
Explicitly use codecov token

jrichter1:
trying out loading the codecov token directly from the scripts


jrichter1:
Apparently, this version works

mohitsuman:
It still gets FetchError though. But we need to fix it somehow for the other OS variants.

mohitsuman:
Closinng this as the fix is done by #911

Issue #869, opened 2019-05-17
Passwords are not masked in logs and error messages


jrichter1:
The logs/errors from login tasks contain passwords in plain text, which sort of defeats the purpose of hiding them in the input field.
```
"/home/jrichter/.vs-openshift/odo" login https://192.168.42.197:8443 -u developer -p developer ..."
```

PR #912, opened 2019-06-28
hide password in output channel [WIP]

sudhirverma:
Fix: #869

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6,6 +6,7 @@
 import * as childProcess from 'child_process';
 import * as vscode from 'vscode';
 import { ExecException, ExecOptions } from 'child_process';
+import { HidePassword } from './util/hidepassword';
 
 export interface CliExitData {
     readonly error: ExecException;
@@ -65,19 +66,12 @@ class OdoChannelImpl implements OdoChannel {
     }
 
     prettifyJson(str: string) {
-        const tokenRegex = /--token\s*=\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
-        const credentialRegex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
         let jsonData: string;
         try {
             jsonData = JSON.stringify(JSON.parse(str), null, 2);
         } catch (ignore) {
-            const hideTokenRegex = str.match(tokenRegex);
-            if (hideTokenRegex) {
-                return hideTokenRegex ? str.replace(hideTokenRegex[1], '**********'): str;
-            } else {
-                const hideCredentialRegex = str.match(credentialRegex);
-                return (hideCredentialRegex) ? str.replace(hideCredentialRegex[0], '-p **********'): str;
-            }
+            const hidePass = HidePassword.hideTokenpassword(str);
+            return HidePassword.hideCredentialPassword(hidePass);
         }
         return jsonData;
     }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -65,31 +65,25 @@ class OdoChannelImpl implements OdoChannel {
     }
 
     prettifyJson(str: string) {
+        const tokenRegex = /--token\s*=\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
+        const credentialRegex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
         let jsonData: string;
         try {
             jsonData = JSON.stringify(JSON.parse(str), null, 2);
         } catch (ignore) {
-            return str;
+            const hideTokenRegex = str.match(tokenRegex);
+            if (hideTokenRegex) {
+                return hideTokenRegex ? str.replace(hideTokenRegex[1], '**********'): str;
+            } else {
+                const hideCredentialRegex = str.match(credentialRegex);
+                return (hideCredentialRegex) ? str.replace(hideCredentialRegex[0], '-p **********'): str;
+            }
         }
         return jsonData;
     }
 
-    getToken(value: string) {
-        const regex = /--token\s*=\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
-        const tokenRegex = value.match(regex);
-        return (tokenRegex) ?  value.replace(tokenRegex[1], '**********') : value;
-    }
-
-    getCredentialsLogin(value: string) {
-        const regex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
-        const tokenRegex = value.match(regex);
-        return (tokenRegex) ? value.replace(tokenRegex[0], '-p **********'): value;
-    }
-
     print(text: string): void {
-        let textData = this.prettifyJson(text);
-        textData = this.getToken(textData);
-        textData = this.getCredentialsLogin(textData);
+        const textData = this.prettifyJson(text);
         this.channel.append(textData);
         if (textData.charAt(textData.length - 1) !== '\n') {
             this.channel.append('\n');
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6,7 +6,7 @@
 import * as childProcess from 'child_process';
 import * as vscode from 'vscode';
 import { ExecException, ExecOptions } from 'child_process';
-import { HidePassword } from './util/hidepassword';
+import { Filters } from './util/filters';
 
 export interface CliExitData {
     readonly error: ExecException;
@@ -70,8 +70,8 @@ class OdoChannelImpl implements OdoChannel {
         try {
             jsonData = JSON.stringify(JSON.parse(str), null, 2);
         } catch (ignore) {
-            const hidePass = HidePassword.hideTokenpassword(str);
-            return HidePassword.hideCredentialPassword(hidePass);
+            const hidePass = Filters.filterToken(str);
+            return Filters.filterPassword(hidePass);
         }
         return jsonData;
     }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -74,8 +74,22 @@ class OdoChannelImpl implements OdoChannel {
         return jsonData;
     }
 
+    getToken(value: string) {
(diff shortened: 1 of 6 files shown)


mohitsuman:
This fix will cover the scenario where in the output channel the password is masked. 
But do we need to also have a scenario when it is displayed on the error box (when the password field is incorrect).

src/cli.ts
```diff
@@ -75,8 +75,22 @@ class OdoChannelImpl implements OdoChannel {
...
+        const regex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
+        const tokenRegex = value.match(regex);
+        return (tokenRegex) ? value.replace(tokenRegex[0], '-p **********'): value;
+    }
+
     print(text: string): void {
-        const textData = this.prettifyJson(text);
+        let textData = this.prettifyJson(text);
+        textData = this.getToken(textData);
+        textData = this.getCredentialsLogin(textData);
```
mohitsuman: Can we merge these 3 lines and make use of one field. It seems that it is using the recursive method to determine the final value of `textData`.

src/openshift/openshiftItem.ts
```diff
@@ -55,6 +55,18 @@ export abstract class OpenShiftItem {
         return (validator.matches(value, '^[a-z]([-a-z0-9]*[a-z0-9])*$')) ? null : message;
     }
 
+    static hideTokenpassword(value: string) {
+        const regex = /--token\s*=\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
```
jrichter1: This regex looks quite complicated. Not sure if tokens can contain spaces, but if not, it can be simplified to this:
```typescript
/--token\s*=\s*([^\s]*)/
```
Or a more universal regex can be used, like the one I suggested for the password masking.

src/openshift/openshiftItem.ts
```diff
@@ -55,6 +55,18 @@ export abstract class OpenShiftItem {
...
+        const tokenRegex = value.match(regex);
+        return (tokenRegex) ?  value.replace(tokenRegex[1], '**********') : value;
+    }
+
+    static hideCredentialPassword(value: string) {
+        const regex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
+        const tokenRegex = value.match(regex);
+        return (tokenRegex) ? value.replace(tokenRegex[0], '-p **********'): value;
```
jrichter1: You could replace just the first group and get rid of the '-p' like in the token masking method.

src/openshift/openshiftItem.ts
```diff
@@ -55,6 +55,18 @@ export abstract class OpenShiftItem {
...
     }
 
+    static hideTokenpassword(value: string) {
+        const regex = /--token\s*=\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
+        const tokenRegex = value.match(regex);
+        return (tokenRegex) ?  value.replace(tokenRegex[1], '**********') : value;
+    }
+
+    static hideCredentialPassword(value: string) {
+        const regex = /-p\s*(.\w[a-zA-Z0-9!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]*)/;
```
jrichter1: This regex is not going to match all the possible characters in a password, spaces for example will throw it off. I'd be in favour of something more universal and simpler, like
```typescript
/--token\s*=\s*(.*)\s/
```
That way it matches anything until the last whitespace that separates the next flag.

src/cli.ts
```diff
@@ -66,11 +66,19 @@ class OdoChannelImpl implements OdoChannel {
...
         try {
             jsonData = JSON.stringify(JSON.parse(str), null, 2);
         } catch (ignore) {
-            return str;
+            const hideTokenRegex = str.match(tokenRegex);
+            if (hideTokenRegex) {
+                return hideTokenRegex ? str.replace(hideTokenRegex[1], '**********'): str;
+            } else {
+                const hideCredentialRegex = str.match(credentialRegex);
+                return (hideCredentialRegex) ? str.replace(hideCredentialRegex[0], '-p **********'): str;
```
jrichter1: This method could use the same treatment as the openshiftItem ones. Please see the comments about these.

jeffmaury: Why are you not reusing  the methods from openShiftItem ?

src/openshift/openshiftItem.ts
sudhirverma: @jrichter1  if I use the tokenRegex[1], remove the -p from my regex for replacing and if there are two same string  for example 
```
username: abc
password: abc
then the output would be 
`login https://192.168.64.74:8443 -u ***(which is abc) -p abc --insecure-skip-tls-verify`
it will replace the first string match.
```
that is why I am using the `-p and password ` so that it will update the only password.

jrichter1: I meant keep the -p part of the regex, use tokenRegex[1] and replace the match with only asterisks. That shouldn't cause this sort of problem.

sudhirverma: If I do `value.replace(tokenRegex[1], '-p **********')`
```
username: abc
password: abc
then the output would be 
login https://192.168.64.74:8443 -u -p ***(which is abc) -p abc --insecure-skip-tls-verify
```

jrichter1: nevermind, I kept thinking about the regex I totally forgot replace takes the first occurence of a string

src/util/hidepassword.ts
```diff
@@ -0,0 +1,14 @@
+export class HidePassword {
+
+    static hideTokenpassword(value: string) {
+        const regex = /--token\s*=\s*([^\s]*)/;
+        const tokenRegex = value.match(regex);
+        return (tokenRegex) ?  value.replace(tokenRegex[1], '**********') : value;
```
dgolovin: should be as simple as:
``` JavaScript
 return value.replace(/--token\s*=\s*([^\s]*)/], '**********') ;
```

src/util/hidepassword.ts
```diff
@@ -0,0 +1,14 @@
...
+    }
+
+    static hideCredentialPassword(value: string) {
```
dgolovin: The same here.

src/util/hidepassword.ts
```diff
@@ -0,0 +1,14 @@
+export class HidePassword {
```
dgolovin: HidePassword.hideTokenpassword is not a good name.  
Why not just export two functions instead without class at all:
* filterPassword
* filterToken

name this file as `filters.ts`.

Then in code it would look like:

``` JavaScript
import * as Filters from './utils/filters';

Filters.filterPassword(string);
```

src/util/hidepassword.ts
sudhirverma: I am matching the string and replacing the password on a group-wise. if I just use the regex here it will replace the password for token
 ```oc login https://192.168.64.82:8443 --token=nLKyqdaaSii54GZWHJQVeO_fVEPK8rgve6warak_yA0 -> oc login https://192.168.64.82:8443 *******``` 
and for credential 
``` `login https://192.168.64.74:8443 -u abc -p abc --insecure-skip-tls-verify`  -> `login https://192.168.64.74:8443 -u ***(which is abc) -p abc --insecure-skip-tls-verify` ```

 ```

File: src/openshift/openshiftItem.ts (hunk 3 above)
Comment (dgolovin) [reply src/openshift/openshiftItem.ts]: Lets keep it simple:

``` JavaScript
export class Filters {
    static readonly tokenRegex = /--token\s*=\s*([^\s]*)/;

    static filterToken(value: string) {
        return value.replace(Filters.tokenRegex, '--token **********');
    }

    static readonly passwordRegex = /-p\s*(.*)\s/;

    static filterPassword(value: string) {
        return value.replace(Filters.passwordRegex, '-p **********');
    }
}
```

src/util/filters.ts
```diff
@@ -0,0 +1,14 @@
+export class Filters {
```
dgolovin: Copyright comment is missing.

dgolovin: Should be simply:

``` JavaScript
export class Filters {
    static readonly tokenRegex = /--token\s*=\s*([^\s]*)/;

    static filterToken(value: string) {
        return value.replace(Filters.tokenRegex, '--token **********');
    }

    static readonly passwordRegex = /-p\s*(.*)\s/;

    static filterPassword(value: string) {
        return value.replace(Filters.passwordRegex, '-p **********');
    }
}
```

test/openshift/cluster.test.ts
```diff
@@ -17,6 +17,7 @@ import { TestItem } from './testOSItem';
 import { OpenShiftItem } from '../../src/openshift/openshiftItem';
```
dgolovin: This does not look like a test for a change. It looks like do stubbing to make test to pass. Methods from Filters class should work just fine if test data is correct. So I would stick to fixing test data to make tests pass. Now it fails because of changes in error message generation. 

```
const err = new Error('FATAL ERROR');
err['cmd'] = 'command text'; 
```

instead of just text.

src/openshift/cluster.ts
```diff
@@ -172,7 +173,7 @@ export class Cluster extends OpenShiftItem {
             .then(() => Cluster.odo.execute(Command.odoLoginWithUsernamePassword(clusterURL, username, passwd)))
             .then((result) => Cluster.save(username, passwd, password, result))
             .then((result) => Cluster.loginMessage(clusterURL, result))
-            .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${error}'!`));
+            .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${Filters.filterPassword(error.cmd)}'!`));
```
dgolovin: by doing this we are loosing error message and printing out only command line with filtered password.

src/openshift/cluster.ts
```diff
@@ -205,7 +206,7 @@ export class Cluster extends OpenShiftItem {
         return Promise.resolve()
             .then(() => Cluster.odo.execute(Command.odoLoginWithToken(clusterURL, ocToken)))
             .then((result) => Cluster.loginMessage(clusterURL, result))
-            .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${error}'!`));
+            .catch((error) => Promise.reject(`Failed to login to cluster '${clusterURL}' with '${Filters.filterToken(error.cmd)}'!`));
```
dgolovin: Same issue here.

PR #924, opened 2019-07-09
add description to component status(pushed, not pushed, no context)

### LINKED ISSUE #885: Distinguish pushed and not pushed component in OpenShift Application Explorer tree
Since odo v1.0.0-beta1 component's deployment state can be deployed and not deployed. That state should be visible in OpenShift Application Explorer tree.
Simplest way to do that use tree node label with suffix `deployed`/`not deployed`. Another option is different icons for components in different states.

mohitsuman:
Fix #885 

Screenshot 2019-07-10 at 5 07 59 AM


src/util/constants.ts
```diff
@@ -0,0 +1,30 @@
+/*-----------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE file in the project root for license information.
+ *-----------------------------------------------------------------------------------------------*/
+
+'use strict';
+
+export enum ContextType {
```
dgolovin: Please, move this back to odo.ts.

mohitsuman: Done.

Issue #915, opened 2019-07-03
Add command `Start Build` for BuildConfig resources in Clusters view


dgolovin:
OpenShift BuildConfig resource items have standard context menu.

![image](https://user-images.githubusercontent.com/620330/60568795-b0131400-9d22-11e9-9855-43e285bec766.png)

There should be OpenShift related commands it that menu, similar to OpenShift Console page for BuildConfig.

![image](https://user-images.githubusercontent.com/620330/60569045-3deeff00-9d23-11e9-8467-91e32b5ce457.png)

There are two available that would make sense to have:
* Delete (already in standard context menu)
* `Start Build` should be added

![image](https://user-images.githubusercontent.com/620330/61509839-53139100-a9a5-11e9-9440-2fa257e349a2.png)

`Start Build` command should use `oc` command line tool to start a build.

PR #930, opened 2019-07-12
Add command `Start Build` for BuildConfig resources in Clusters view

sudhirverma:
Fix: #915

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,11 @@
         "category": "OpenShift"
       },
       {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -89,6 +90,11 @@
         "title": "About",
         "category": "OpenShift"
       },
+      {
+        "command": "openshift.build.start",
```
dgolovin: please use `clusters.openshift.*` namespace.

PR #1015, opened 2019-08-06
fix open console for 4.x clusters

### LINKED ISSUE #981: open console command for clusters for OpenShift 4.x does not work
For OpenShift clusters 4.x, the api url to login is different to console URL. Hence the `OpenConsole` commands leads to the API url which is incorrect.

Resolution: The command should open the `OpenShift web-console` URL.

mohitsuman:
fix #981

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -77,7 +77,6 @@ export class Cluster extends OpenShiftItem {
             const serverUrl = await Cluster.odo.execute(Command.showServerUrl());
             consoleUrl = `${serverUrl.stdout}/console`;
         }
-        window.showInformationMessage(`Opening console url: ${consoleUrl}`);
         open(consoleUrl);
     }
 
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -66,17 +66,19 @@ export class Cluster extends OpenShiftItem {
         Cli.getInstance().showOutputChannel();
     }
 
-    static async openshiftConsole(context: OpenShiftObject): Promise<void> {
-        if (context) {
-            open(`${context.getName()}/console`);
+    static async openshiftConsole(): Promise<void> {
+        let consoleUrl: string;
+        const versionInfo = await Cluster.odo.execute(Command.clusterVersionDescribe(), process.cwd(), false);
+        if (versionInfo.error === null) {
+            const routeObj = await Cluster.odo.execute(Command.getOpenshiftClusterRoute());
+            const spec = JSON.parse(routeObj.stdout).items[0].spec;
+            consoleUrl = `${spec.port.targetPort}://${spec.host}`;
         } else {
-            const result: OpenShiftObject[] = await Cluster.odo.getClusters();
-            if (result.length>0 && result[0].getName().startsWith('http')) {
-                open(`${result[0].getName()}/console`);
-            } else {
-                window.showErrorMessage(result[0].getName());
-            }
+            const serverUrl = await Cluster.odo.execute(Command.showServerUrl());
+            consoleUrl = `${serverUrl.stdout}/console`;
         }
+        window.showInformationMessage(`Opening console url: ${consoleUrl}`);
+        open(consoleUrl);
     }
 
     static async switchContext() {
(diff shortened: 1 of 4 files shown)


src/openshift/cluster.ts
```diff
@@ -66,17 +66,19 @@ export class Cluster extends OpenShiftItem {
         Cli.getInstance().showOutputChannel();
     }
 
-    static async openshiftConsole(context: OpenShiftObject): Promise<void> {
-        if (context) {
-            open(`${context.getName()}/console`);
+    static async openshiftConsole(): Promise<void> {
+        let consoleUrl: string;
+        const versionInfo = await Cluster.odo.execute(Command.clusterVersionDescribe(), process.cwd(), false);
```
dgolovin: Why is it required to call `oc describe` first?

src/openshift/cluster.ts
```diff
@@ -66,17 +66,19 @@ export class Cluster extends OpenShiftItem {
...
-            const result: OpenShiftObject[] = await Cluster.odo.getClusters();
-            if (result.length>0 && result[0].getName().startsWith('http')) {
-                open(`${result[0].getName()}/console`);
-            } else {
-                window.showErrorMessage(result[0].getName());
-            }
+            const serverUrl = await Cluster.odo.execute(Command.showServerUrl());
+            consoleUrl = `${serverUrl.stdout}/console`;
         }
+        window.showInformationMessage(`Opening console url: ${consoleUrl}`);
```
dgolovin: No need to show the message.

src/openshift/cluster.ts
```diff
@@ -66,17 +66,19 @@ export class Cluster extends OpenShiftItem {
...
     }
 
-    static async openshiftConsole(context: OpenShiftObject): Promise<void> {
-        if (context) {
-            open(`${context.getName()}/console`);
+    static async openshiftConsole(): Promise<void> {
+        let consoleUrl: string;
+        const versionInfo = await Cluster.odo.execute(Command.clusterVersionDescribe(), process.cwd(), false);
+        if (versionInfo.error === null) {
+            const routeObj = await Cluster.odo.execute(Command.getOpenshiftClusterRoute());
```
dgolovin: What is output for OpenShift 3.x for this command?

src/openshift/cluster.ts
mohitsuman: Now to get to know which version of server is used, either 3.x or 4.x, there is no way from oc or odo.
The ``oc describe clusterversion` command works for 4.x server and errors out for 3.x. So the check is done for that.

For 3.x, there is no `clusterversion`.

src/openshift/cluster.ts
mohitsuman: The message was added to inform that the url is being loaded for some cases where the open in browser takes some time.

src/openshift/cluster.ts
mohitsuman: For 3.x, there is no namespace `openshift-console` and hence the route does not work here. 

For 4.x, the cluster url is generates using operator and we get the route using the namespace for `openshift-console`.

mohitsuman:
@dgolovin After 4.2 release, I think `oc version` will have the server version also mentioned and then we can directly use that command and determine the server version.

src/openshift/cluster.ts
dgolovin: It should be shown with progress then.

src/openshift/cluster.ts
dgolovin: then clusterVersionDescribe is not needed,  just analyse getOpenshiftClusterRoute result.

mohitsuman: But `getOpenshiftClusterRoute` for 3.x does not throw an error. It responds with empty items list. Do you want to just use the condition over the `items` object and iterate ?

src/openshift/cluster.ts
mohitsuman: Actually now I tried with progress, but it's an overkill, the open browser is sufficient for the action. Hence removing with the new commit.

src/openshift/cluster.ts
mohitsuman: @dgolovin After conversation with openshift dev team, they too suggested that `oc get clusterversion` is the correct differentiator as it works only on 4.x.

PR #1018, opened 2019-08-07
Make sure icon code is not propagated in context switching (fix #990)

### LINKED ISSUE #990: Switching Context opens a new instance of VS Code if code is on path
The following command is called when switching oc contexts:
```oc config use-context $(code)  /192-168-42-99:8443/developer```

Not sure what the  ```$(code)``` part is for, but my bash is interpreting it as a command to launch VS Code.

jrichter1:


mohitsuman:
@jrichter1 I am gonna change this logic and and replace the context switch to explorer tree. Will link the PR for this.

dgolovin:
Will be fixed in #1019.

PR #1019, opened 2019-08-07
add switch context to tree view title

### LINKED ISSUE #996: add action to switchContexts for the entire tree
Currently the switch user action will be changed to just switching between the `users` based on the cluster loggedin.

A new action will be added to the Explorer to switch between different contexts available in the kubeconfig.

mohitsuman:
fix #996


package.json
```diff
@@ -142,6 +142,15 @@
         "title": "Login into Cluster with credentials",
         "category": "OpenShift"
       },
+      {
+        "command": "openshift.explorer.switchContext",
+        "title": "Manage Contexts",
```
dgolovin: It is not really about managing contexts, but rather about changing contexts.

src/openshift/cluster.ts
```diff
@@ -82,11 +82,11 @@ export class Cluster extends OpenShiftItem {
     static async switchContext() {
         const k8sConfig = new KubeConfigUtils();
         const contexts = k8sConfig.contexts.filter((item) => item.name !== k8sConfig.currentContext);
-        const contextName: QuickPickItem[] = contexts.map((ctx) => ({ label: `$(code)  ${ctx.name}`}));
-        const choice = await window.showQuickPick(contextName, {placeHolder: "Select the new context"});
+        const contextName: QuickPickItem[] = contexts.map((ctx) => ({ label: `${ctx.name}`}));
+        const choice = await window.showQuickPick(contextName, {placeHolder: "Select the new openshift context"});
```
dgolovin: openshift => 'OpenShift'

package.json
mohitsuman: There is a very funny thing here, the order in which the titles appear on the explorer bar is based on the title name. So to make the order correct I used something which starts after 'L' and before 'R' and hence Manage. Is there a way to sort the order ?

dgolovin: The same way  it is done in context menus I guess.

mohitsuman: No, it doesn't take group ordering. It just sorts name alphabetically.
And the contents displayed are only there if it's under `"group": "navigation"`.

Can we rename it to `Manage Context Switch` ?

mohitsuman: okay, i think I have a solution. sending the updates.

mohitsuman: Fixed by https://github.com/redhat-developer/vscode-openshift-tools/pull/1019/commits/886eb7f2b9b0df729efcf33a9e36466004005438

src/openshift/cluster.ts
mohitsuman: Fixed by https://github.com/redhat-developer/vscode-openshift-tools/pull/1019/commits/886eb7f2b9b0df729efcf33a9e36466004005438

mohitsuman:
> ```
> error: server took too long to respond with version information.
> ```

Can you help in how you reproduced this issue, I can then have a look. 
Is this happening from context switch 3.x to 4.x or vice versa ?

mohitsuman:
@dgolovin I have added a working scenario with the attached gif. This allows me to perform the context switching between different clusters. 
Here I am doing a switch between local 3.x and local 4.x cluster. The actions works seamlessly. Can you please check and let me know.

switch-ctxt

PR #1022, opened 2019-08-08
Add k8s ext dependency and install stepo for travis config

dgolovin:


dgolovin:
This https://github.com/redhat-developer/vscode-openshift-tools/blob/10714c55a33bcf0897824a80eb4cc7540ed4a2cf/Jenkinsfile#L32  should be removed now.

PR #1063, opened 2019-08-19
Add 'Create Service' command context menu for Projects

### LINKED ISSUE #1036: Add 'Create Service' command context menu for Projects
Command should handle application selection the same way it is done in 'Create' command for Component:
1. `+ Create new Application` item in quick pick list along with existing applications
2. Create Application node if new Application requested

mohitsuman:
fix #1036

### DIFF
--- a/package.json
+++ b/package.json
@@ -678,9 +678,14 @@
           "group": "p3"
         },
         {
+          "command": "openshift.service.create",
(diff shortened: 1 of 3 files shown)


mohitsuman:
> Looks good, but corner use case should be covered when there is no applications yet in targeted project.

@dgolovin This case is already covered. Can you please check and ack.

PR #1105, opened 2019-08-27
Add `Show logs` command for DeploymentConfig node

sudhirverma:
Fix: #1103


src/k8s/deployment.ts
```diff
@@ -71,6 +75,15 @@ export class DeploymentConfig {
             'You have no replicas available');
     }
 
+    static async dcShowLog(context: { id: any; }): Promise<string> {
```
dgolovin: Should be named `showLog`. No need for 'dc' prefix, because the method belongs to DeploymentConfig class.

src/k8s/deployment.ts
```diff
@@ -17,6 +17,10 @@ export class Command {
         return `oc get deploymentConfig -o json`;
     }
 
+    static showLogDc(deploymentConfig: string) {
```
dgolovin: `showDeploymentConfigLog` would be better, because other methods are using full names as well.

src/k8s/deployment.ts
```diff
@@ -71,6 +75,15 @@ export class DeploymentConfig {
             'You have no replicas available');
     }
 
+    static async showLog(context: { id: any; }): Promise<string> {
+        let deployName: string = context ? context.id : null;
```
dgolovin: `context.id => context.name` otherwise it is asking to select deployment config from quick pick.

PR #1112, opened 2019-08-28
Add unit test for DeploymentConfigNodeContributor.

sudhirverma:
Fix: #1110

### DIFF
--- a/test/unit/k8s/deployment.test.ts
+++ b/test/unit/k8s/deployment.test.ts
@@ -9,9 +9,11 @@ import * as vscode from 'vscode';
 import * as chai from 'chai';
 import * as sinonChai from 'sinon-chai';
 import * as sinon from 'sinon';
+import * as k8s from 'vscode-kubernetes-tools-api';
 import { OdoImpl } from '../../../src/odo';
 import { Progress } from '../../../src/util/progress';
-import { DeploymentConfig, Command } from '../../../src/k8s/deployment';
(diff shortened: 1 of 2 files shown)


test/k8s/deployment.test.ts
```diff
@@ -73,6 +74,39 @@ suite('K8s/deployment', () => {
...
+            }
+        };
+
+        setup(async () => {
+            kubectl = await extension.kubectl.v1;
+        });
+
+        test('should able to get the children node of deployment Config', async () => {
+            if (kubectl.available) {
+                sandbox.stub(kubectl.api, 'invokeCommand').resolves({code: 0,
```
dgolovin: Would be good to test with actual output here, so getChildren would return at least two nodes.

dgolovin:
@sudhirverma please rebase on top of the master.

test/unit/k8s/deployment.test.ts
```diff
@@ -98,6 +100,40 @@ suite('K8s/deployment', () => {
...
+                description: "",
+                displayName: "DeploymentConfigs",
+                label: "DeploymentConfigs",
+                manifestKind: "DeploymentConfig",
+                pluralDisplayName: "DeploymentConfigs"
+            }
+        };
+
+        setup(() => {
+            const api: k8s.API<k8s.KubectlV1> = {
```
dgolovin: @sudhirverma I was asking to do something like this. 
You stub external api, create object instance and test actual methods, not prototype.

sudhirverma: Thanks, Denis :). I will check why test are failing.

PR #1130, opened 2019-09-06
add webpack support

### LINKED ISSUE #806: Cofigure webpack build development

mohitsuman:
FIX #806

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
                 "--extensionDevelopmentPath=${workspaceFolder}"
             ],
             "outFiles": [
-                "${workspaceFolder}/out/src/**/*.js"
+                "${workspaceFolder}/dist/**/*.js"
             ],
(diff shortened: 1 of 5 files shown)


jrichter1:
There is a whole bunch of failed tests, but the process still returns 0.

jrichter1:
Plus what @dgolovin  said, sort of defeats the purpose of using webpack.

mohitsuman:
I had configured webpack as per the official docs and how other projects use it. But still tests are failing and return type is 0. Need to check why is this happening. 

Ideally webpack bundle should reduce the size of the build. :(

dgolovin:
This is happening, because you changed compilation script to use webpack and that in turn changed location where compiled files land, to 'dist' folder. Tests are using compilation to build everything before running the tests and then try to run it from `./out/src` folder.

mohitsuman:
@dgolovin You mean that the compilation folder should be changed to `out` ?

Issue #1134, opened 2019-09-09
Passwords with spaces need handling


jrichter1:
Logging in using a password with spaces is interpreted as additional parameters, needs fixing.

PR #1138, opened 2019-09-09
Passwords with spaces need handling

sudhirverma:
Fix: #1134


jrichter1:
I'd say white space is a valid password character, we just need a way to pass it to odo as a single argument. Wrapping it in quotes would be the first thing I'd try.

PR #1158, opened 2019-09-13
git+binary workflow fix

### LINKED ISSUE #1159: context folder selection for git based workflow
context folder selection for git based workflow should be before the git url step. This also checks whether the selection context folder has any previous component configuration or not.

mohitsuman:
1. Add context folder selection at start for git workflow
2. Use workspace.ts for folderSelection for git & binary workflow
2. Check for component folder already existing at the selected path.

Fix #1159


mohitsuman:
> Calling file inside util module as workspaceUtil.ts is too much 'util' ;)
> Have you checked workspace.ts BTW?

workspace.ts is renamed as workspaceUtil. As `class workspace` from vscode was conflicting.

src/util/workspaceUtil.ts
```diff
@@ -0,0 +1,67 @@
...
+            const folders = await window.showOpenDialog({
+                canSelectFiles: false,
+                canSelectFolders: true,
+                canSelectMany: false,
+                defaultUri: Uri.file(Platform.getUserHomePath()),
+                openLabel: "Add context folder for component in workspace."
+            });
+            if (!folders) return null;
+            if (await this.checkComponentFolder(folders[0])) {
+                window.showErrorMessage('The folder selected already contains a component. Please select a different folder');
```
dgolovin: It should not return null, it should show quick pick again. This way it would not required to call create component command manually.

mohitsuman: You mean to show the same QuickPick for context folder selection again to be displayed ?

dgolovin: yes

dgolovin: @mohitsuman there were not problems with file name before, so lets keep it workspace.ts.

mohitsuman: Fixed as per review comments.

PR #1162, opened 2019-09-14
add `delete config directory` with component delete action

### LINKED ISSUE #1161: Delete the config file from directory after delete operation
Added a --all flag for component delete which deletes the component from the cluster as well as removes the local config file and directory.

mohitsuman:
fix #1161

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -145,7 +145,7 @@ export class Command {
         return `oc wait pvc/${storage}-${app}-pvc --for=delete --namespace ${project}`;
     }
     static deleteComponent(project: string, app: string, component: string) {
-        return `odo delete ${component} -f --app ${app} --project ${project}`;
+        return `odo delete ${component} -f --app ${app} --project ${project} --all`;
     }
     static describeComponent(project: string, app: string, component: string) {
         return `odo describe ${component} --app ${app} --project ${project}`;
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -144,6 +144,9 @@ export class Command {
     static waitForStorageToBeGone(project: string, app: string, storage: string) {
         return `oc wait pvc/${storage}-${app}-pvc --for=delete --namespace ${project}`;
     }
+    static undeployComponent(project: string, app: string, component: string) {
+        return `odo delete ${component} -f --app ${app} --project ${project}`;
+    }
     static deleteComponent(project: string, app: string, component: string) {
         return `odo delete ${component} -f --app ${app} --project ${project} --all`;
     }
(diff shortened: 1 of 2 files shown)


src/odo.ts
```diff
@@ -145,7 +145,7 @@ export class Command {
         return `oc wait pvc/${storage}-${app}-pvc --for=delete --namespace ${project}`;
     }
     static deleteComponent(project: string, app: string, component: string) {
-        return `odo delete ${component} -f --app ${app} --project ${project}`;
+        return `odo delete ${component} -f --app ${app} --project ${project} --all`;
```
dgolovin: 👍

src/openshift/component.ts
```diff
@@ -61,13 +61,20 @@ export class Component extends OpenShiftItem {
...
 
     static async del(treeItem: OpenShiftObject): Promise<string> {
+        let warningMsg: string;
         const component = await Component.getOpenShiftCmdData(treeItem,
             "From which Project do you want to delete Component",
             "From which Application you want to delete Component",
             "Select Component to delete");
         if (!component) return null;
         const name: string = component.getName();
-        const value = await window.showWarningMessage(`Do you want to delete Component '${name}\'?`, 'Yes', 'Cancel');
```
dgolovin: 👎

src/openshift/component.ts
```diff
@@ -87,7 +94,7 @@ export class Component extends OpenShiftItem {
             (component) => component.contextValue === ContextType.COMPONENT_PUSHED);
         if (!component) return null;
         const name: string = component.getName();
-        const value = await window.showWarningMessage(`Do you want to undeploy Component '${name}\'?`, 'Yes', 'Cancel');
+        const value = await window.showWarningMessage(`Do you want to undeploy Component '${name}\'? This will remove the component from the cluster.`, 'Yes', 'Cancel');
```
dgolovin: 👎

src/openshift/component.ts
```diff
@@ -61,13 +61,20 @@ export class Component extends OpenShiftItem {
     }
```
dgolovin: I would change only command, keep the messages the same.

mohitsuman: The message is added so that the user is notified what exactly delete action will do. We already have a command for undeploy too. So to give `Delete` action more clarity, I have added the message specific to `pushed` components.

jrichter1: I'm gravitating towards the current messages, feels to me like the distinction between undeploy and delete is quite intuitive. ¯\_(ツ)_/¯

mohitsuman: Sure thing then, will then roll back those changes.

PR #1165, opened 2019-09-16
Add `Pause/Resume` command for DeploymentConfig node

sudhirverma:
Fix: #1109

### DIFF
--- a/package.json
+++ b/package.json
@@ -513,6 +513,16 @@
         "category": "OpenShift"
       },
       {
+        "command": "clusters.openshift.deploy.pause",
+        "title": "Pause",
+        "category": "OpenShift"
+      },
+      {
+        "command": "clusters.openshift.deploy.resume",
+        "title": "Resume",
+        "category": "OpenShift"
+      },
+      {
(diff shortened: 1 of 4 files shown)


dgolovin:
Please, rebase.

sudhirverma:
> What would happen if you call 'Pause' or 'Resume' finished deployment?

What I am trying to do is if the deployment is on Resume state then we can deploy new replica
and if the deployment is on Pause state it will not allow to deploy new replica.

> What if you call 'Pause' on paused deployment, or 'Resume' on running deployment?

Here what I am trying is if the deployment is in Resume state the I will show only Pause command
and if the deployment is in Pause state then I will show only Resume Command

dgolovin:
@sudhirverma Let's simplify the task. Add both commands to the DeploymentConfig node. Check deployment state before running specific command. When 'Resume' command called and dc state is not 'paused' show warning without executing command. When 'Pause' called and dc is not in  'running' state show warning without executing command.

mohitsuman:
Closing this PR as the workflow is old and we should tackle this in a different way if needed.

Issue #1157, opened 2019-09-13
Sometimes it does't clone the git repository while creating Component.


sudhirverma:
Steps:
1. Using https://github.com/sclorg/nodejs-ex to create a component.
2. Following all the steps to create a component from the context view.
3. Selecting `yes` for information message `Do you want to clone the git repository for created Component?`

and It doesn't clone the repo in the selected folder.

dgolovin:
@sudhirverma Please add description to replicate.

dgolovin:
@sudhirverma  It newer meant to clone git repo to selected folder, because we creating component with --git option.
The reason is `git.clone` command does not have and API to control location where to clone git repository.

sudhirverma:
@dgolovin but when I am selecting `Yes` to clone the git repository then it was not executing https://github.com/redhat-developer/vscode-openshift-tools/blob/master/src/openshift/component.ts#L453 code and It doesn't show me the UI to select the location to clone.

PR #1169, opened 2019-09-17
Fix clone the git repository while creating Component.

sudhirverma:
Fix: #1157

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -448,10 +448,8 @@ export class Component extends OpenShiftItem {
 
         if (!componentTypeVersion) return null;
 
-        window.showInformationMessage('Do you want to clone git repository for created Component?', 'Yes', 'No').then((value) => {
-            value === 'Yes' && commands.executeCommand('git.clone', repoURI);
-        });
-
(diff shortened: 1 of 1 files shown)


src/openshift/component.ts
```diff
@@ -448,10 +448,8 @@ export class Component extends OpenShiftItem {
 
         if (!componentTypeVersion) return null;
 
-        window.showInformationMessage('Do you want to clone git repository for created Component?', 'Yes', 'No').then((value) => {
-            value === 'Yes' && commands.executeCommand('git.clone', repoURI);
-        });
-
+        const response = await window.showInformationMessage('Do you want to clone git repository for created Component?', 'Yes', 'No');
+        if (response) await commands.executeCommand('git.clone', repoURI);
```
mohitsuman: what if the response is `No` ?

sudhirverma: Should we return null if the response is no or we should create the component without cloning the component that's what is was happening before?

mohitsuman: check for Yes condition and then clone the repo. Cloning the repo is an optional stuff provided to the user. If the response is NO, skip and proceed to component creation.

Issue #724, opened 2019-03-20
 Use `odo catalog list services -o json` to get list of available service templates and plans


dgolovin:
Command generation code:
https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L47
Output parsing code:
https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L345
https://github.com/redhat-developer/vscode-openshift-tools/blob/0ea1753d4666fe5956d7156476f1d4a754d5cc93/src/odo.ts#L353

Upstream issue https://github.com/openshift/odo/issues/1357

PR #1211, opened 2019-10-01
Add json output for catalog list services

sudhirverma:
Fix: #724

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -99,7 +99,7 @@ export class Command {
         return `odo catalog list components`;
     }
     static listCatalogServices () {
-        return `odo catalog list services`;
+        return `odo catalog list services -o json`;
     }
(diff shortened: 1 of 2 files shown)


src/odo.ts
```diff
@@ -773,29 +773,27 @@ export class OdoImpl implements Odo {
...
-        let items: Array<any>;
+        let items: any[] = [];
         const result: cliInstance.CliExitData = await this.execute(Command.listCatalogServices(), Platform.getUserHomePath());
         try {
             items = JSON.parse(result.stdout).items;
-        } catch (err) {
-            // ignore
+        } catch (ignore) {
         }
-        const stpObject: Array<any> = items.filter((value) => value.metadata.name === svcName);
```
dgolovin: No need filter and map when you can just search. It means less iterations over array.

Issue #1247, opened 2019-10-23
Open console command doesn't open the console web page.


sudhirverma:
Open console command doesn't open the console web page.

for login: using `oc login -u developer -p developer https://api.crc.testing:6443` command and using latest crc.

Openshift version: 4.x

mohitsuman:
@sudhirverma It works for me when running CRC 1.0.0. What are you using to run OpenShift ?

sudhirverma:
@mohitsuman I am using 
```
crc version: 1.0.0+575079b
OpenShift version: 4.2.0 (embedded in binary)
```

https://github.com/code-ready/crc/releases/tag/1.0.0

mohitsuman:
Do you mean the issue is while using command palette?

Please update the description accordingly.

jrichter1:
Just tried this with an actual 4.2 cluster (not CRC). The thing is, the login URL is completely different to the console URL. And all the tooling does is slap `/console` at the end of the login URL.
So I ended up on a non-existent API endpoint
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/console\"",
  "reason": "Forbidden",
  "details": {
    
  },
  "code": 403
}
```

jrichter1:
Just for reference:
 - the console URL format is `https://console-openshift-console.apps.openshift4-${cluster-hostname}`
 - the login (API) URL `https://api.openshift4-${cluster-hostName}:6443`

jeffmaury:
You should ask @adietish because the console URL must be retrieved from configMap for OCP 4 and he did the job for openshift-rest-client.

sudhirverma:
I will use `oc get configmaps console-public  -n openshift-config-managed -o json` to get Web console URL for Openshift 4.2 url

mohitsuman:
@sudhirverma The console url from Tree node action works correctly. It is the command palette thing which needs to be fixed.

mohitsuman:
> Just tried this with an actual 4.2 cluster (not CRC). The thing is, the login URL is completely different to the console URL. And all the tooling does is slap `/console` at the end of the login URL.
> So I ended up on a non-existent API endpoint
> 
> ```
> {
>   "kind": "Status",
>   "apiVersion": "v1",
>   "metadata": {
>     
>   },
>   "status": "Failure",
[...]

PR #1248, opened 2019-10-23
Fix open console command

sudhirverma:
Fix: #1247

### DIFF
--- a/src/k8s/console.ts
+++ b/src/k8s/console.ts
@@ -7,86 +7,82 @@ import * as vscode from 'vscode';
 import { Command } from "../odo";
 import { KubeConfigUtils } from "../util/kubeUtils";
 import { OpenShiftItem } from '../openshift/openshiftItem';
+import { CliExitData } from '../cli';
 
 const k8sConfig = new KubeConfigUtils();
-const clusterUrl = k8sConfig.getCurrentCluster().server;
 const project = (k8sConfig.contexts).find((ctx) => ctx.name === k8sConfig.currentContext).namespace;
 
 export class Console extends OpenShiftItem {
 
-    static async fetchClusterVersion() {
-        const versionInfo = await Console.odo.execute(Command.getclusterVersion(), process.cwd(), false);
-        return versionInfo.error;
-    }
-
     static async fetchOpenshiftConsoleUrl() {
-        if (await Console.fetchClusterVersion() === null) {
-            const routeObj = await Console.odo.execute(Command.getOpenshiftClusterRoute());
-            const spec = JSON.parse(routeObj.stdout).items[0].spec;
-            const consoleUrl = `${spec.port.targetPort}://${spec.host}`;
-            return consoleUrl;
-        } else {
+        try {
+            return await Console.odo.execute(Command.showConsoleUrl());
+        } catch (ignore) {
             const consoleUrl = await Console.odo.execute(Command.showServerUrl());
             return consoleUrl.stdout;
         }
     }
 
-    static async openBuildConfig(context: { id: any; }) {
+    static openShift4ClusterUrl(consoleUrl: string | CliExitData): string {
+        return JSON.parse(consoleUrl['stdout']).data.consoleURL;
+    }
+
+    static async openBuildConfig(context: { name: any; }) {
         let url = '';
         if (!context) {
             vscode.window.showErrorMessage("Cannot load the build config");
             return;
         }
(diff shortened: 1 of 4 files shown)


src/k8s/console.ts
```diff
@@ -27,8 +27,8 @@ export class Console extends OpenShiftItem {
             const consoleUrl = `${spec.port.targetPort}://${spec.host}`;
             return consoleUrl;
         } else {
-            const consoleUrl = await Console.odo.execute(Command.showServerUrl());
-            return consoleUrl.stdout;
+            const getUrlObj = await Console.odo.execute(Command.showConsoleUrl());
+            return JSON.parse(getUrlObj.stdout).data.consoleURL;
```
mohitsuman: This change is specific to 3.x cluster and not 4.x cluster. In the issue you have mentioned that the 4.x cluster does not work properly.

sudhirverma: okay @mohitsuman I will check.

sudhirverma: old code works only when we login with kubeadmin but it doesn't work when we login with `oc login -u developer -p developer https://api.crc.testing:6443` for 4.x cluster.

now it works for both kubeadmin and for developer.

dgolovin:
Will test on 3.x and 4.x clusters, before pushing to the master.

src/k8s/console.ts
```diff
@@ -7,86 +7,82 @@ import * as vscode from 'vscode';
...
         let url = '';
         if (!context) {
             vscode.window.showErrorMessage("Cannot load the build config");
             return;
         }
         const consoleUrl = await Console.fetchOpenshiftConsoleUrl();
-        if (await Console.fetchClusterVersion() === null) {
-            url = `${consoleUrl}/k8s/ns/${context.id}/buildconfigs`;
+        if (consoleUrl['stdout']) {
+            url = `${Console.openShift4ClusterUrl(consoleUrl)}/k8s/ns/${context.name}/buildconfigs`;
```
dgolovin: This does not work for 4.2.2. It should be
``` JavaScript
url = `${Console.openShift4ClusterUrl(consoleUrl)}/k8s/ns/${project}/buildconfigs/${context.name}`;
```

The same applies to other resources in OpenShift 4.2.2 cluster.

src/k8s/console.ts
```diff
@@ -7,86 +7,82 @@ import * as vscode from 'vscode';
...
         let url = '';
         if (!context) {
             vscode.window.showErrorMessage("Cannot load the deployment config");
             return;
         }
         const consoleUrl = await Console.fetchOpenshiftConsoleUrl();
-        if (await Console.fetchClusterVersion() === null) {
-            url = `${consoleUrl}/k8s/ns/${context.id}/deploymentconfigs`;
+        if (consoleUrl['stdout']) {
+            url = `${Console.openShift4ClusterUrl(consoleUrl)}/k8s/ns/${context.name}/deploymentconfigs`;
```
dgolovin: See comment above

src/k8s/console.ts
```diff
@@ -7,86 +7,82 @@ import * as vscode from 'vscode';
...
         if (!context) {
             vscode.window.showErrorMessage("Cannot load the image stream");
             return;
         }
-        const consoleUrl = await this.fetchOpenshiftConsoleUrl();
-        if (await Console.fetchClusterVersion() === null) {
-            url = `${consoleUrl}/k8s/ns/${context.id}/imagestreams`;
+        const consoleUrl = await Console.fetchOpenshiftConsoleUrl();
+        if (consoleUrl['stdout']) {
+            url = `${Console.openShift4ClusterUrl(consoleUrl)}/k8s/ns/${context.name}/imagestreams`;
```
dgolovin: See comment above

src/k8s/console.ts
```diff
@@ -7,86 +7,82 @@ import * as vscode from 'vscode';
...
         let url = '';
         if (!context) {
             vscode.window.showErrorMessage("Cannot load the Project");
             return;
         }
         const consoleUrl = await Console.fetchOpenshiftConsoleUrl();
-        if (await Console.fetchClusterVersion() === null) {
-            url = `${consoleUrl}/overview/ns/${context.id}`;
+        if (consoleUrl['stdout']) {
+            url = `${Console.openShift4ClusterUrl(consoleUrl)}/overview/ns/${context.name}`;
```
dgolovin: this works, because context == project == namespace.

PR #1250, opened 2019-10-25
Fix vulnerabilities issue for package-lock.json

sudhirverma:


mohitsuman:
@sudhirverma which vulnerability is getting fixed here ? There is no change in package.json file.

sudhirverma:
@mohitsuman 
```
updated 1 package and audited 1009 packages in 4.525s
found 2 high severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details
```

mohitsuman:
> @mohitsuman
> 
> ```
> updated 1 package and audited 1009 packages in 4.525s
> found 2 high severity vulnerabilities
>   run `npm audit fix` to fix them, or `npm audit` for details
> ```

This can vary machine to machine and depends on what npm version you are using. If there is a change in package.json entry, it makes sense to update the lock file. Say, if we remove the node modules and again to do `npm install`, it is bound to update the package-lock file. So we should refrain from updating it everytime.

Issue #1260, opened 2019-11-04
Sometimes login does not work anymore


ch-stark:
**Environment**
VS Code version: 1.39.2
OS: linux
Extension version: 0.1.1

**Description**

I tested the update from friday (1.1) and it works quite well but there is one issue that the login is some kind of stalled.

This means there is no way to see the projects anymore.
It says you are already logged in.....

Workaround is to open a terminal and login manually.
After that, when I refresh, I get all the projects

Further question:

If I click: Show Output Channel is there a way to disable this again?

dgolovin:
> If I click: Show Output Channel is there a way to disable this again?
There is no way to remove channel in vscode. Extension itself does not provide any options to disable output to that channel.
You can switch to different tab: Problems, Output, Debug Console and Terminal.
Another option to switch to less noisy output channel: Kubernetes, Extensions or any from the list on the right. Here is the screen-shot for reference.
![image](https://user-images.githubusercontent.com/620330/68158433-73d80600-ff04-11e9-8ebf-5b434a8c3ffa.png)

dgolovin:
@ch-stark what is the version of OpenShift you are working with?

dgolovin:
Should be fixed with resent #1267 and #1279.

PR #1279, opened 2019-11-14
Use double quotes in login command on windows

jrichter1:
- cmd doesn't support wrapping arguments in single quotes, which ended up in credentials being passed as 'username' and 'password' - use double quotes instead
 - at the same time, unix shells should still use single quotes to avoid string interpolation


dgolovin:
Related to #1260.

PR #1293, opened 2019-11-18
Create 'OpenShift: Show Resource in Console' for command palette 

sudhirverma:
Fix: #1281


mohitsuman:
We are not going this route for the current workflow, hence closing the PR.

Issue #1307, opened 2019-11-27
Use ocp cli 4.3.3 instead of 3.11


dgolovin:
https://mirror.openshift.com/pub/openshift-v4/clients/ocp/ (4.3.3 currently latest)

Should be tested with all commands we currently use and with 3.x clusters as well before we switch.

PR #1308, opened 2019-11-28
updated ocp cli version

sudhirverma:
Fix: #1307

### DIFF
--- a/src/tools.json
+++ b/src/tools.json
@@ -33,9 +33,9 @@
         "vendor": "Red Hat, Inc.",
         "name": "oc",
         "cmdFileName": "oc",
-        "version": "4.2.9",
-        "versionRange": "^4.2.9",
-        "versionRangeLabel": "version >= 3.11.0 and <= 4.2.9",
+        "version": "4.2.2",
+        "versionRange": "^4.2.2",
+        "versionRangeLabel": "version >= 3.11.0 and <= 4.2.2",
         "filePrefix": "",
         "platform": {
             "win32": {
--- a/src/tools.json
+++ b/src/tools.json
@@ -34,8 +34,8 @@
         "name": "oc",
         "cmdFileName": "oc",
         "version": "4.2.2",
-        "versionRange": "^4.2.2",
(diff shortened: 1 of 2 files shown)


dgolovin:
@sudhirverma have you had a chance to try it with 3.x?

sudhirverma:
@dgolovin I have check on cluster 3.x it was working fine.

src/tools.json
```diff
@@ -33,28 +33,28 @@
         "vendor": "Red Hat, Inc.",
         "name": "oc",
         "cmdFileName": "oc",
-        "version": "3.11.0",
-        "versionRange": "^3.11.0",
-        "versionRangeLabel": "version >= 3.11.0 and < 4.0.0",
+        "version": "4.2.2",
+        "versionRange": "^4.2.2",
+        "versionRangeLabel": "version >= 3.11.0 and <= 4.2.2",
```
dgolovin: ^4.2.2 version range means `version >= 3.11.0 and < 5.0.0` according to semver module doc https://www.npmjs.com/package/semver#caret-ranges-123-025-004.
We should allow to use oc found in path if it differs only in patch version ~4.2.2 which means `version >= 4.2.2  and < 4.3.0`.

src/tools.ts
```diff
@@ -106,7 +106,12 @@ export class ToolsConfig {
     public static async getVersion(location: string, cmd: string = path.parse(location).name): Promise<string> {
         let detectedVersion: string;
         if (fs.existsSync(location)) {
-            const version = new RegExp(`${cmd.toLocaleLowerCase()} v((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?).*`);
+            let version: RegExp;
```
dgolovin: This can be single RegExp expression.

dgolovin:
Replaced with #1387.

PR #1440, opened 2020-03-14
some changes in the grammar articles used

girishramnani:
This PR consists of some grammar changes to certain texts which are shown to the user. The changes suggested in my opinion sound grammatically correct ( I tried to research as well ).

These changes are subjective and hence if it doesn't make sense we can skip it.


dgolovin:
Thank you @girishramnani, those definitely should be indefinite articles.

PR #1462, opened 2020-03-19
[WIP]add editor view for show log command

### LINKED ISSUE #1319: Use webview to show or follow component's log
Calling 'Follow/Show Log' command form OpenShift Application View creates new terminal every time. Cons of this approach are:
1. There is no way to close all/part of terminals with one operation
2. Showing long log in terminal takes time and cannot be canceled

Using special editor to show/follow log would solve all those problems.

mohitsuman:
Fix #1319 

This implementation is WIP and currently handles only `Show Log` command prototype. Changes will be made for the actual command to work.


mohitsuman:
The logs display will be supported using the webview mode in #1490 . Hence closing this PR.

PR #1490, opened 2020-04-09
WIP: Filter view for show logs using webview implementation

mohitsuman:
Still there will  be some UX improvements. This PR does not use LazyLog implementation. It showcases the log directly using React components.


dgolovin:
Implemented with React LazyLog component in different PR.

PR #1584, opened 2020-05-22
Ignore k8s config loading error and let odo report it when called

### LINKED ISSUE #1580: Error when kubeconfig has duplicate clusters.
I have `KUBECONFIG=/home/josh/kubeconfig:/home/josh/kubedir/kubeconfig` set in my ENV. 

Both kubeconfigs point to a cluster with the same name but a different set of keys. Basically, one is older and not even up anymore. 

When I open up VS Code with my kubeconfigs set this way, I get an error in the lower right. 
`Activating extension 'redhat.vscode-openshift-connector' failed: Duplicate cluster: [NAME_OF_CLUSTER].`

dgolovin:
This PR fixes #1580.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -46,13 +46,17 @@ export class OpenShiftExplorer implements TreeDataProvider<OpenShiftObject>, Dis
         .onDidChangeTreeDataEmitter.event;
 
     private constructor() {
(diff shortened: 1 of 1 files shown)


dgolovin:
The fix is to ignore the error when loading extension and let odo report failure when it called first time.

PR #1587, opened 2020-05-27
Recommend extension, based on installed operators

evidolob:
It would be nice if we provide recommendation for extension based on installed operators in OpenShift cluster. It may help end developer to have all tools that he need, based on his working environment(cluster).

This is first, and quite naive implementation, I want to start discussion about this feature.

In this PR I add recommendation for Tekton and KNative extensions. As they relative new and this feature may help to spread them.

Demo: 
ezgif com-video-to-gif (15)

> This demo recorded with cluster which have `openshift-pipelines-operator` and `serverless-operator` operators installed, also that vscode instance doesn't have Tekton and KNative extensions. I filter out recommendations with currently installed extensions, and if user has all recommended extensions I will not bother him with any recommendation.


evidolob:
@gorkem @siamaksade @veillard @joshuawilson @mohitsuman @dgolovin 
WDYT?

jeffmaury:
Would it be more generic to base it on CRD instead of operators ?

veillard:
Sounds good to me in principle but I can't juge the details of the implementation TBH :-)

evidolob:
Hm, it may, I need to look on how to match CRD to extension, as each operator may provide a bunch of CRD's.

joshuawilson:
I like the idea. However, without a connected story/experience why would we suggest to users to do this?
Recommended extensions is usually for those extensions that a tool requires to run properly. I am concerned that make our users feel forced or pushed into something. I think this kind of thing requires PM oversight.

jeffmaury:
My idea is to have a file where we have, for each extension, a list of CRD that this extension is about. Ex for the OpenShift Connector, you will have Route, DeploymentConfig, etc. As soon as you find one of this CRD in the cluster, then you select this extension.

@joshuawilson this has been long asked by @veillard so I suppose PM is ok

joshuawilson:
In that case, I really do like the idea.

gorkem:
I like it. Should we make this a generic library that we include on all extensions that talk with clusters? There is no reason why vscode-knative extension to not recommend vscode-tekton.

dgolovin:
@veillard  @gorkem @evidolob Should it work with any k8s cluster? If so, I would use kubernetes js clients (at least in vscode extensions) to check for CRDs. That way we not depend on any cli tool.

dgolovin:
@jeffmaury are you suggesting to do something similar to runtimes in JBossTools. Where runtimes would be extensions and we can leverage throughout all IDEs we have, meaning it would be:
 - extension or library for Java to use in Intellij and eclipse
 - extension or npm module to use in vscode

jeffmaury:
Yes, my idea is that at least the format can be reused for IntelliJ and Eclipse

gorkem:
> Should it work with any k8s cluster?

Yes, it should. We should seek to have this on kubernetes extension too. Or perhaps only there if all of our k8s extensions have a dependency to it already.

evidolob:
> Should we make this a generic library that we include on all extensions that talk with clusters? There is no reason why vscode-knative extension to not recommend vscode-tekton.

@gorkem 
Having separate library may be not so easy to integrate with vscode UI for each extension, because it will require to have  own treeview with proposed extensions list. And may be not so convenient to end user when several extensions will promote same extensions in different UI parts. 

 I think it would be better to have one separate extension for extension recommendation, it may integrate recommendation view in to k8s extension view container(like in this PR it integrate with openshift view), and all our k8s related extensions may depend on it. 
I lake to have single point and UI where recommendations are prompted.

>Yes, it should. We should seek to have this on kubernetes extension too. Or perhaps only there if all of our k8s extensions have a dependency to it already.
I know that tekton and knative extensions do not have dependency on `vscode-openshift-tools`, so yes we need to share this functionality across extensions, including `vscode-openshift-tools`.

Until we discuss this, I will continue this, I will look on CRD and use kubernetes js clients instead of `oc` CLI call.

evidolob:
@jeffmaury I think it would be easier to map extensions to `group` field of `CustomResourceDefinitionSpec ` object. Almost all Tekton CRD has `tekton.dev` and KNative has `operator.knative.dev` value. 
In this case json map may be:
```json
{
  "tekton.dev" : ["extension.foo", "extension.bar"],
  "operator.knative.dev": ["extension.foo1", "extension.bar2"]
}
```
Another option may be matching of CRD name, by providing regexp to match name(metadata.name field value).
In this case json map may be:
```json
{
  "/.*\.operator\.knative\.dev/" : ["extension.foo", "extension.bar"],
  "/.*\.tekton\.dev/": ["extension.foo1", "extension.bar2"]
}
```
WDYT?

jeffmaury:
Agree we I said CRD I meant api groups so the format looks good to me. The question is do we have a single file or one file per IDE as id are different per IDE

evidolob:
In my opinion, we can start with vscode only, and after we may think how to make that mapping file more generic to fit all IDE's.

One more question: Should we show recommended extensions as tree? I mean root items is "api group" and theirs child is extensions itself, or we should keep extensions as plain list?

evidolob:
I update PR, now it use `@kubernetes/client-node` to get CRD from cluster, 
and change mapping file to use CRD API group instead of operator name.

I had to update `@kubernetes/client-node` to `0.12.0` because `0.11.1`  do not have API to get CRD

PR #1520, opened 2020-04-29
Add webview to run CRC

mohitsuman:
Fix #1235

Issue #1605, opened 2020-06-10
Provide more usability to the crc webviews


mohitsuman:
Based on the current `crc` webview implementation, there needs to be some more usability fixes in the workflow.

- [x] When `crc setup` is running, the other actions should be disabled till the command has fixed execution.
- [x] Add more information to the steps so that user has idea about what the steps exactly does
- [x] Provide back option so that user need not restart from top.
- [x] Only use latest version of crc to be downloaded for the user.

dgolovin:
editor should also check the status of the vm before going forward, because it might be running already.

dgolovin:
Initialize values for input fields based on `crc config view` output. For example if pull secret is already in config, the pull secret file field should contain path from config.

dgolovin:
Download for crc should be done by extension with report progress, so there would be no need asking user to point to downloaded file.

dgolovin:
We should open issue for crc to provide command to show setup status. This way extension do not need ask user to do setup, it would just check and run it if needed.

mohitsuman:
The fixes are released in `0.1.6` version.

PR #1606, opened 2020-06-15
Add webview to run CRC (#1520)

dgolovin:
Provide CRC view, with start and stop features


dgolovin:
Current implementation will require to go through all steps to get to the point where something could be done with CRC: Setup, Start and Stop.
Editor should save last selected location for CRC and understand in what state CRC at the moment using `crc status`. That would let to handle use cases like:
1. Open editor
2. Download
3. Setup CRC
4. Run CRC
5. Close the editor
6. Do your work 
7. Open editor and stop CRC

Otherwise it is going to be hurdle to go through all the steps over and over again for every simple Stop/Start operation.

dgolovin:
Also there is no need to have a `Setup`  button, it can be executed every time before start. Basically we should have start and stop button on last step. If status of CRC is running only stop button available and vice versa.

mohitsuman:
The current commit https://github.com/redhat-developer/vscode-openshift-tools/pull/1606/commits/e342a6a47b611c0c99451ecd52ada48b8817df3e makes the webview UX in alignment with vscode color themes to make it look more native to vscode.

Dark Theme Representation:

Screenshot 2020-08-01 at 4 10 40 AM

Light Theme Representation:
Screenshot 2020-08-01 at 4 11 21 AM

mohitsuman:
Now CRC can be stopped from the editor using the command palette and it fetches the binary path from settings or user can specify new path for the same. So it won't be needed to go through the workflow again to stop the crc running.

mohitsuman:
Added `crc status` output as separate card. Now the status is reflected after start command has successfully finished.
Once the view is closed and it is opened again, the `crc binary` path is checked from workspace settings and accordingly status card is shown on the UI. Thus the user need not do everything from scratch.

If the user wishes to reset the configuration, Reset action is available which allows the step by step workflow to start new instance of CRC.

mohitsuman:
Screenshot 2020-08-11 at 4 22 41 AM


This is how the crc status card looks on the UI.

mohitsuman:
This is how the loading screen looks when the `crc status` is being fetched.

loading-screen

mohitsuman:
@dgolovin Please review.

mohitsuman:
Fixes #1707 #1605

dgolovin:
View fails at first start 

```
rejected promise not handled within 1 second: SyntaxError: Unexpected end of JSON input
extensionHostProcess.js:1007
stack trace: SyntaxError: Unexpected end of JSON input
	at JSON.parse (<anonymous>)
	at Function.<anonymous> (c:\p\vscode-openshift-tools\out\src\view\cluster\clusterViewLoader.js:116:87)
	at Generator.next (<anonymous>)
	at fulfilled (c:\p\vscode-openshift-tools\out\src\view\cluster\clusterViewLoader.js:5:58)
	at processTicksAndRejections (internal/process/task_queues.js:85:5)
``` 

that results in detecting crc start is running forever.

dgolovin:
I can open as many 'Add OpenShift Cluster' pages as I want.

image

dgolovin:
The number of CRC Log items in Output View is confusing

image

Which one is current one?

dgolovin:
On windows it does not work without `-n dns.ip.add.ress/hostname` there is no way to set that through the editor.

mohitsuman:
Fixes #1707 #1605 

> -n dns.ip.add.ress/hostname

@dgolovin What doesn't work here ?  Is it the setup or start action?

mohitsuman:
> On windows it does not work without `-n dns.ip.add.ress/hostname` there is no way to set that through the editor.

@dgolovin What doesn't work ? Is it the `crc setup` or `crc start` ?

mohitsuman:
> In Dark theme text in input fields is hardly readable.
> 
> image

Fixed in commit https://github.com/redhat-developer/vscode-openshift-tools/pull/1606/commits/850c905697d51c4ea49836320c80b9be9aa7fe33

Screenshot 2020-08-19 at 5 42 32 PM

mohitsuman:
> I can open as many 'Add OpenShift Cluster' pages as I want.
> 
> image

I think this problem is for all the webview pages, we need to add action for every webview.

mohitsuman:
> On windows it does not work without `-n dns.ip.add.ress/hostname` there is no way to set that through the editor.

As discussed, this was a local issue. So resolving this.

PR #1621, opened 2020-07-03
I1559

dgolovin:


dgolovin:
Replaced with #1622, #1623 and #1625.

PR #1683, opened 2020-08-03
Display the correct component name when deleting URL

jrichter1:


src/openshift/url.ts
```diff
@@ -62,7 +62,7 @@ export class Url extends OpenShiftItem{
         if (url) {
             const value = await window.showWarningMessage(`Do you want to delete URL '${url.getName()}' from Component '${url.getParent().getName()}'?`, 'Yes', 'Cancel');
             if (value === 'Yes') {
-                return Progress.execFunctionWithProgress(`Deleting URL ${url.getName()} from Component ${component.getName()}`, () => Url.odo.deleteURL(url))
+                return Progress.execFunctionWithProgress(`Deleting URL ${url.getName()} from Component ${url.getParent().getName()}`, () => Url.odo.deleteURL(url))
```
mohitsuman, line 65: This should take the name of `${component.getName()}` as component is defined above.

PR #1718, opened 2020-08-29
Fix double refresh when creating a project

dgolovin:
This PR fixes the issue with double refresh when create new project.
Creating new project triggers update of application view because of
changes in kubeconfig. At the same time create new project command
itself triggers tree update. That ends with tree updated, but error
appears in debug output about three item requested to be revealed
does not exists in the tree, because it was refreshed as result of
kubeconfig changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -935,6 +935,11 @@
       ],
       "view/item/context": [
         {
+          "command": "openshift.explorer.refresh",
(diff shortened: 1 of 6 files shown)


mohitsuman:
@dgolovin with latest odo worklfow, I think we do not need this issue anymore. If we don not encounter this, let's close this PR.

PR #1752, opened 2020-09-17
add crc webview check for che/theia

### LINKED ISSUE #1751: Run CRC webview only in vscode and not on Che instance
Add condition to load `Add OpenShift Cluster` for only VSCode and not for Che instance.

mohitsuman:
Signed-off-by: Mohit Suman <mohit.skn@gmail.com>

Fix #1751

### DIFF
--- a/package.json
+++ b/package.json
@@ -803,7 +803,7 @@
         },
         {
           "command": "openshift.explorer.addCluster",
-          "when": "view == openshiftProjectExplorer"
+          "when": "view == openshiftProjectExplorer && isWindows || isLinux || isMac"
         },
         {
           "command": "openshift.explorer.login",
@@ -909,7 +909,7 @@
       "view/title": [
         {
           "command": "openshift.explorer.addCluster",
-          "when": "view == openshiftProjectExplorer",
+          "when": "view == openshiftProjectExplorer && isWindows || isLinux || isMac",
           "group": "navigation@0"
         },
         {
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "bugs": "https://github.com/redhat-developer/vscode-openshift-tools/issues",
   "engines": {
-    "vscode": "^1.30.0"
+    "vscode": "^1.49.0"
   },
   "badges": [
     {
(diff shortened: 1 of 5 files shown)


mohitsuman:
@dgolovin Please review and merge.

package.json
```diff
@@ -13,7 +13,7 @@
   },
   "bugs": "https://github.com/redhat-developer/vscode-openshift-tools/issues",
   "engines": {
-    "vscode": "^1.30.0"
+    "vscode": "^1.49.0"
```
dgolovin, line 16: Not needed, because we are not using any new API introduced after 1.30.0.

package.json
```diff
@@ -105,7 +105,7 @@
     "@types/targz": "^1.0.0",
     "@types/tmp": "0.1.0",
     "@types/validator": "^10.11.3",
-    "@types/vscode": "1.30.0",
+    "@types/vscode": "1.49.0",
```
dgolovin, line 108: The same as above.

src/debug.ts
```diff
@@ -91,8 +91,8 @@ export class DebugSessionsView implements TreeDataProvider<string>, Disposable {
         return undefined;
     }
 
-    refresh(): void {
-        this.onDidChangeTreeDataEmitter.fire();
+    refresh(element?: string): void {
+        this.onDidChangeTreeDataEmitter.fire(element);
```
dgolovin, line 95: not needed, it never updates specific element, because elements in the debug view have no children. It always updates whole view.

src/watch.ts
```diff
@@ -68,7 +68,7 @@ export class WatchSessionsView implements TreeDataProvider<string> {
         return undefined;
     }
 
-    refresh(): void {
-        this.onDidChangeTreeDataEmitter.fire();
+    refresh(element?: string): void {
```
dgolovin, line 71: Not needed for the same reason as debug view.

package.json
dgolovin: should be 1.40.0, where uiKind was introduced

dgolovin (approved):

PR #1805, opened 2020-10-19
Remove `url describe` until it is restored in odo 2

dgolovin:
See https://github.com/openshift/odo/issues/4126 for odo team comment on the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -208,8 +208,6 @@
     "onCommand:openshift.catalog.listComponents",
     "onCommand:openshift.catalog.listServices",
     "onCommand:openshift.url.create",
-    "onCommand:openshift.url.describe",
-    "onCommand:openshift.url.describe.palette",
     "onCommand:openshift.url.delete",
     "onCommand:openshift.url.delete.palette",
     "onCommand:openshift.component.openUrl",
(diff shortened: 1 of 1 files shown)


mohitsuman:
We need to move the command and use `odo url list` as it has all the information needed.
```
$ odo url list
{
	"kind": "List",
	"apiVersion": "odo.dev/v1alpha1",
	"metadata": {},
	"items": [
		{
			"kind": "url",
			"apiVersion": "odo.dev/v1alpha1",
			"metadata": {
				"name": "8080-tcp",
				"creationTimestamp": null
			},
			"spec": {
				"host": "8080-tcp-spring-failure-check-default.apps-crc.testing",
				"protocol": "http",
				"port": 8080,
				"secure": false,
				"kind": "route",
				"path": "/"
			},
			"status": {
				"state": "Pushed"
			}
		}
	]
}

```
And as  per odo team discussion here https://github.com/openshift/odo/pull/2694#issuecomment-601604714

dgolovin:
@mohitsuman that is related to https://github.com/openshift/odo/issues/4126 and there is nothing to do but remove it for now.

PR #2012, opened 2021-03-09
Components View first iteration

### LINKED ISSUE #2008: 'Components View' to show components in current workspace
Currently Applications Explorer View shows only components form current context, but in case there are components in workspace form different project or project from another clusters they are not visible in Application Explorer View. Components view should show all the worksapce folders that contains odo component and let browse and open source files without switching to Explorer view.

dgolovin:
This PR fixes #2008.

This fix adds:
1. Components View (View)  under Application Explorer
2. View shows components discovered in workspace root folders
3. Detection for devfile components based on odo env view, which works
   when cluster is not accessible or user is logged off from cluster
4. Detection for s2i components based on parsing .odo/config.yaml
   (forkaround for odo env view not working for s2i components)
5. Refresh for View based after event about new folder added to
   workspace
6. Refrehs for View after component created form existing workspace
   folder (to avoid using watch for every folder in workspace)
7. Reveal in Explorer for components to activate Explorer view and
   select context folder (not always working correctly first time, have
   no idea why)
8. No support to create component when there is no accessible cluster or
   user is logged off (there should be quick input requests for project
   and application names, there should be qiuickpiks provided that allow
   to select existing and not existing projects/applications)

Not implemented yet:
1. Show folders/files under component node (requires some work to
   incorporate filesystem watching with exlusion for folders like
   'node_modules'
2. Watching for filesystem changes or focus ganed event to detect
   changes done in vscode terminal or external terminal
3. Showing hierarchy Project/Application/Component rather than flat list
4. Support pushing components to cluster from the view, especially for
   components with project that does not exist or currently not active
5. Showing state Pushed/Not pushed for components pushed to current
   cluster/namespace (or current context)
6. No icons yet
7. No view Welcome page with 'Create Component' button

image


jrichter1:
Works in general, there are a few things I think should be improved:
 - deleting a project leaves all associated context folders open, the components won't be removed from the view unless the user manually closes the folders
 - no auto refresh after deleting a component, manual refresh works fine
 - I'd like to see more info about the components, like what template/devfile they are built on, or what state they are in
 - the OpenShift view is getting a bit packed with all the new sections, maybe some of them could be at least collapsed by default? Watch or debug session sections look like they are not really needed until somebody creates such a session.
 - `reveal in explorer` indeed doesn't work on the first try, it seems to focus the currently opened file instead

dgolovin:
>     * deleting a project leaves all associated context folders open, the components won't be removed from the view unless the user manually closes the folders

That is intentional. Imagine you have a project and want to try to deploy it to cluster. Then you delete component and your working project is gone from workspace as well.

>     * no auto refresh after deleting a component, manual refresh works fine

That is a good catch.

> * I'd like to see more info about the components, like what template/devfile they are built on, or what state they are in

We will get to it, for now the purpose is to let to see components without being connected/loggedin to cluster

> * the OpenShift view is getting a bit packed with all the new sections, maybe some of them could be at least collapsed by default? Watch or debug session sections look like they are not really needed until somebody creates such a session.

Not sure if that is possible, will check

> * `reveal in explorer` indeed doesn't work on the first try, it seems to focus the currently opened file instead

I noticed that to, but the same command is used in git sidebar for staged resources, that how I found it accidentally. Will check if calling it twice, solves the problem.

jrichter1:
> > ```
> > * deleting a project leaves all associated context folders open, the components won't be removed from the view unless the user manually closes the folders
> > ```
> 
> That is intentional. Imagine you have a project and want to try to deploy it to cluster. Then you delete component and your working project is gone from workspace as well.

I understand the need to keep the folders open. But the way this is currently implemented, I can delete a project with X components and still be left with those X components in the view, because their folders still contain the `.odo` subfolder. Ideally, the user should be left with the workspace folders open, but no components from a non existent project.

PR #2422, opened 2022-05-13
Detect login status using listing projects to support k8s clusters

### LINKED ISSUE #2393: Using kubernetes context with user with valid certificate data does not show content of cluster in Application Explorer
OpenShift application explorer is empty and shows welcome screen even when docker-descktop context selected that points to correct kubernetes cluster and references user with valid certificates.

It happens because to detect cluster url `odo version` command used. The extension expects `Server: serverurl` string is presented in stdout, but in case of kubernetes cluster it is not presented.

dgolovin:
This PR fixes #2393.


dgolovin:
image

Issue #2415, opened 2022-05-11
Allow users to edit registries added


mohitsuman:
Currently, users can Add and Remove registries. We need to provide an `Edit` option in the context menu to update the name and URL for the registry
![reg-add-remove-edit](https://user-images.githubusercontent.com/620330/169177268-bb4056f8-5762-4656-a4d1-41ab0fd3f41c.gif)
.

PR #2396, opened 2022-04-11
Webview based registry browser

msivasubramaniaan:
Creating webview based devfile registry view

Fixes #2417

PR #2424, opened 2022-05-16
Edit option for registry

msivasubramaniaan:
This PR helps to updating the registry name and URL

Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/2415

### DIFF
--- a/src/componentTypesView.ts
+++ b/src/componentTypesView.ts
@@ -87,8 +87,12 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
     }
 
     public async getRegistries(): Promise<Registry[]> {
-        if (!this.registries) {
-            this.registries = await this.odo.getRegistries();
+        try {
+            if (!this.registries) {
+                this.registries = await this.odo.getRegistries();
+            }
+        } catch (err) {
+            this.registries = [];
         }
         return this.registries;
     }
@@ -175,11 +179,12 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
     @vsCommand('openshift.componentTypesView.registry.add')
     public static async addRegistryCmd(registryContext: Registry): Promise<void> {
         // ask for registry
+        const registries = await ComponentTypesView.instance.getRegistries();
         const regName = await window.showInputBox({
             value: registryContext?.Name,
             prompt: registryContext ? 'Edit registry name' : 'Provide registry name to display in the view',
             placeHolder: 'Registry Name',
-            validateInput: async (value) => {
+            validateInput: (value) => {
                 const trimmedValue = value.trim();
                 if (trimmedValue.length === 0) {
                     return 'Registry name cannot be empty';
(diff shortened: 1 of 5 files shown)


mohitsuman:
I think this PR needs to be rebased after the webview UI PR #2396 is merged to the main.

src/componentTypesView.ts
```diff
@@ -197,15 +198,16 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+            prompt: registryContext ? 'Edit registry URL' : 'Provide registry URL to display in the view',
             placeHolder: 'Registry URL',
             validateInput: async (value) => {
                 const trimmedValue = value.trim();
                 if (!validator.isURL(trimmedValue)) {
                     return 'Entered URL is invalid';
                 }
                 const registries = await ComponentTypesView.instance.getRegistries();
-                if (registries.find((registry) => registry.URL === value)) {
+                if (!registryContext && registries.find((registry) => registry.URL === value)) {
```
dgolovin, line 210: This disables validation for registry with the same URL to avoid validation error for registryContext initial URL and for other registries. The condition should be a part of find method callback instead.

src/componentTypesView.ts
```diff
@@ -173,10 +173,11 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
     }
 
     @vsCommand('openshift.componentTypesView.registry.add')
-    public static async addRegistryCmd(): Promise<void> {
+    public static async addRegistryCmd(registryContext: Registry): Promise<void> {
         // ask for registry
         const regName = await window.showInputBox({
-            prompt: 'Provide registry name to display in the view',
+            value: registryContext?.Name,
+            prompt: registryContext ? 'Edit registry name' : 'Provide registry name to display in the view',
```
dgolovin, line 180: If one wants to change URL, there is no way to get to URL input unless name of the registry changed first.

src/componentTypesView.ts
```diff
@@ -242,6 +244,14 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
         }
     }
 
+    @vsCommand('openshift.componentTypesView.registry.edit')
+    public static editRegistry(registry: Registry): void {
+        vscode.commands.executeCommand('openshift.componentTypesView.registry.closeView');
```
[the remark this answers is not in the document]
dgolovin: In case edit command is canceled, target registry is going to be removed from the list of registries.

dgolovin (changes requested):

src/componentTypesView.ts
msivasubramaniaan: @dgolovin If user wants to edit the registry name alone than URL, it should allow the user to do. So I have made the condition based on **registryContext** value. It has value only at the time of **edit**, suppose if user wants to add new registry by that time **registryContext** value would be undefined and the other condition will check if the URL already exists/not.

src/componentTypesView.ts
msivasubramaniaan: Yes, Since we named the option as "Edit", then user can edit the name/URL. So Only I have provided the placeholder value by default its existing name. Suppose if user don't want to edit the name then he can go the existing name by pressing enter. Otherwise we should provide option to user as edit name/edit URL, based on his input we can proceed. May I know your view about the same?

dgolovin: I am fine with current implementation. Pressing enter does not work for me, I see error like this below
image
so I cannot get to URL input.

src/componentTypesView.ts
dgolovin: Let's do thought experiment. If you have two registries:
1. name: r1, url: url1
2. name: r2, url: url2
Now if you want to edit registry 1, you can call edit and change name to r3 and url to url2 and you wont see a validation error, because validation for url is going to be disabled.

dgolovin (changes requested):
The current approach is good, but issues with validation should be fixed.

dgolovin (changes requested):
Look at https://github.com/msivasubramaniaan/vscode-openshift-tools/pull/2.

PR #2433, opened 2022-05-19
changed create component workflow

msivasubramaniaan:
changed create component workflow and the application name will be considered the starter project name which selected


dgolovin (approved):
LGTM

Issue #2432, opened 2022-05-19
Add telemetry event to capture Registry View UI


mohitsuman:
Add a telemetry event to capture the following user actions:

- Open Registry View 
- Actions associated with the devfiles
   - Name of devfile
   - Create Component Action
   - Clone To Workspace Action
   - Open in Browser Action
   - selected starter project name 
   - Copy devfile.yaml action

dgolovin:
There is already telemetry for all of that. The only issue is there is no way to tell was Create Component Action called form registry viewer or from command palette.

dgolovin:
'New Component Action' is used only from registry viewer. There is nothing to do for this issue it is all already there.

PR #2440, opened 2022-05-30
added telemetry event for copy dev file

msivasubramaniaan:
Added telemetry event for copying the devfile.yaml from card.

Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/2432


src/telemetry.ts
```diff
@@ -47,7 +47,12 @@ export interface NewComponentCommandProps {
     'use_existing_devfile': boolean;
 }
 
-export type TelemetryProps = Partial<CommonCommandProps & NewComponentCommandProps>;
+export interface copyDevFileProps {
```
dgolovin, line 50: No need for new properties, just use component_type above to report devfile name.

src/telemetry.ts
```diff
@@ -47,7 +47,12 @@ export interface NewComponentCommandProps {
...
+    'devFileName': string;
+    'devFileCopied': boolean;
```
dgolovin, line 52: defFileCopied property is not needed because command id itself serves the same purpose.

dgolovin: That means copyDevFileProps interface is not needed at all.

src/webview/devfile-registry/registryViewLoader.ts
```diff
@@ -42,6 +43,15 @@ async function devfileRegistryViewerMessageListener(event: any): Promise<any> {
         case 'openInBrowser':
             vscode.commands.executeCommand('openshift.componentType.openStarterProjectRepository', starterProject);
             break;
+        case 'telemeteryCopyEvent':
+            const devFileName = event.devFile;
+            const copyClicked = event.copyClicked;
+            const telemetryEventCopyDevFile = new ExtCommandTelemetryEvent('openshift.registryView.starterProjects.copyDevFile');
+            telemetryEventCopyDevFile.send({
+                devFileName: devFileName,
```
dgolovin, line 51: use component_type instead

src/webview/devfile-registry/registryViewLoader.ts
```diff
@@ -42,6 +43,15 @@ async function devfileRegistryViewerMessageListener(event: any): Promise<any> {
...
+                devFileCopied: copyClicked
```
dgolovin, line 52: As I said above command's id 'openshift.registryView.starterProjects.copyDevFile' means exactly the same.

dgolovin (changes requested):

dgolovin (approved):

Issue #2459, opened 2022-06-20
Update debug gifs with the latest workflow


mohitsuman:
The gifs mentioned in the readme are old and does not support the latest UI. We need to update them and add higher resolution gifs for the given workflow.

dgolovin:
The only problem with UI I see is that for some reason it does not switch to Debug view after Debug command is executed for component in Application Explorer.

Resolution of gif is calculated to be as wide as page layout permits. There should be a link to FHD video for every gif.

PR #2460, opened 2022-06-20
Updated debug gifs with the latest workflow

msivasubramaniaan:
Updated debug gifs with the latest workflow.

Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/2459

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -842,7 +842,7 @@ export class Component extends OpenShiftItem {
                     request: 'attach',
                     address: 'localhost',
                     localRoot: component.contextPath.fsPath,
-                    remoteRoot: '/project'
+                    remoteRoot: '/projects'
                 });
             }
         } else {
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -766,7 +766,7 @@ export class Component extends OpenShiftItem {
             return result;
         }
         // const components = await Component.odo.getComponentTypes();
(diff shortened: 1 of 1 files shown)


dgolovin (commented):
Why are you replacing remote debugger for local one?

msivasubramaniaan:
> Why are you replacing remote debugger for local one?

changed to remote debugger

dgolovin (changes requested):
1. Lets use the same size for animated gifs that captures full vscode window. I would suggest to use HD resolution width 1280x720 to keep file size under control.
2. NodeJS debug content seems to be fine except the image size is FHD and there is screen recording tool window visible in the end and should be removed
3. Java debug missing the step to show Java extension installation, because it can be not installed by default.
4. The waiting time for debug session start should be reduced to reduce gif size

msivasubramaniaan:
> 1. Lets use the same size for animated gifs that captures full vscode window. I would suggest to use HD resolution width 1280x720 to keep file size under control.
> 2. NodeJS debug content seems to be fine except the image size is FHD and there is screen recording tool window visible in the end and should be removed
> 3. Java debug missing the step to show Java extension installation, because it can be not installed by default.
> 4. The waiting time for debug session start should be reduced to reduce gif size

Addressed all the given points and changed the gif accordingly. Kindly have a look

dgolovin (approved):

PR #2495, opened 2022-07-28
performance improvement of getting devfile registries

### LINKED ISSUE #2496: Improve loading time for registry viewer
Currently to load the registry viewer, it takes some time to get the response and show it on UI. But everytime the `Show View` is closed and again opened, the odo call to get the registries is done and use has to wait for it to load.

To improve the performance of it, the call should be made once and the json be stored. If there is no update on the registry list, the stored json should be shown in the UI.

If there is a change in the registry list, the json should be updated as soon as the list is updated and during the `Show View`, the call will be faster using the stored json, instead of multiple odo calls.

msivasubramaniaan:

Fix #2496


dgolovin (changes requested):
Please set indentation to 4 spaces.

src/extension.ts
```diff
@@ -142,9 +142,11 @@ export async function activate(extensionContext: ExtensionContext): Promise<any>
         OdoImpl.Instance.loadWorkspaceComponents(event);
     });
 
-    OdoImpl.Instance.loadWorkspaceComponents(null);
+    await ComponentTypesView.instance.getAllComponents().then(() => {
```
dgolovin, line 145: activate function should exit fast, there should be no waiting for long running calls like this. getAllComponents call is slow and that is the reason for test failure in CI.

src/extension.ts
```diff
@@ -142,9 +142,11 @@ export async function activate(extensionContext: ExtensionContext): Promise<any>
...
+        OdoImpl.Instance.loadWorkspaceComponents(null);
```
dgolovin, line 146: loadWorkspaceComponents and startTelemetry should not be dependent on loading registry information.

src/registriesView.ts
```diff
@@ -97,6 +102,48 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+
+    public getListOfRegistries(): Registry[] {
+        return this.registries;
+    }
+
+    public async getAllComponents(): Promise<boolean> {
+        return new Promise<boolean>((resolve) => {
+            const compDescs: Set<ComponentTypeDescription> = new Set<ComponentTypeDescription>();
+            void getInstance().getCompTypesJson().then(async (devFileComponentTypes: DevfileComponentType[]) => {
+                const components = new Set<string>();
```
dgolovin, line 117: no need to call getComponentTypesOfJSON this is old function introduced to deal with two king of component types s2i and devfile. Now there is only one king of components - defvile.

src/registriesView.ts
```diff
@@ -97,6 +102,48 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+                getInstance().getComponentTypesOfJSON(devFileComponentTypes).map((comp) => {
+                    components.add(comp.name);
+                });
+                await this.getRegistries();
+                Array.from(components).map((compName: string) => {
```
dgolovin, line 122: just do devFileComponentTypes.forEach here.

src/registriesView.ts
```diff
@@ -97,6 +102,48 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+                            // eslint-disable-next-line max-nested-callbacks
+                            component.Devfile?.starterProjects?.map((starter: StarterProject) => {
+                                starter.typeName = compName;
+                            });
+                            compDescs.add(component);
+                        });
+                        if (devFileComponentTypes.length === compDescs.size) {
+                            this.compDescriptions.clear();
+                            this.compDescriptions = compDescs;
+                            resolve(true);
```
dgolovin, line 136: Is there a reason to return boolean if value is always `true`?

dgolovin (changes requested):

msivasubramaniaan:
> Please set indentation to 4 spaces.

yes I have the same settings only

dgolovin (changes requested):
Better, but still need some work in several places.

1. I agree we can keep component descriptors array in ComponentTypesView class instance.
2. Your implementation works, but it is not easy to understand and I don't think it is good idea to use subscriber to send events from member function to the same class instance when you can just use this.

You have ComponentTypesView instance and RegistryViewLoader instance. I would suggest to get rid of rxjs and use EventEmitter to nodify ComponentTypesView clients (RegistryViewer in this case) about component description update and the need to refresh.

mohitsuman:
LGTM wrt to the UI part of it with loader and icon. Once @dgolovin is fine with the code workflow, we can merge.

mohitsuman (approved):

src/registriesView.ts
```diff
@@ -97,6 +104,50 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+                        this.compDescriptions.add(component);
+                    });
+                    if (devFileComponentTypes.length === this.compDescriptions.size) {
+                        this.subject.next('refresh');
+                    }
+                }).catch(() => {
+                    isError = true;
+                }).finally(() => {
+                    if (isError) {
+                        if (this.subject.closed) {
```
dgolovin, line 139: Where in the code subject gets closed?

src/registriesView.ts
```diff
@@ -97,6 +104,50 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+                            this.subject = new Subject<string>();
+                            this.getAllComponents();
```
dgolovin, line 141: There is a risk of infinite cycle here. What if error persists? Ether exclude component from the list or mark it accordingly in UI.

src/registriesView.ts
dgolovin: It seems never going to happen.

src/registriesView.ts
```diff
@@ -238,17 +289,18 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+            }
+            await vscode.commands.executeCommand('openshift.componentTypesView.registry.remove', registryContext, true);
         }
 
         const newRegistry = await OdoImpl.Instance.addRegistry(regName, regURL, token);
         ComponentTypesView.instance.addRegistry(newRegistry);
 
-        RegistryViewLoader.refresh();
+        ComponentTypesView.instance.getAllComponents();
+        RegistryViewLoader.refresh()
```
dgolovin, line 303: This introduces cycle dependency. RegistryViewLoader depends on ComponentTypesView and vice versa.

dgolovin: RegistryViewLoader is going to be refreshed during `ComponentTypesView.instance.getAllComponents()` call, because it sends 'refresh' event through the subject.

src/registriesView.ts
```diff
@@ -261,7 +313,10 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
         if (yesNo === 'Yes') {
             await OdoImpl.Instance.removeRegistry(registry.Name);
             ComponentTypesView.instance.removeRegistry(registry);
-            RegistryViewLoader.refresh();
+            if (!isEdit) {
+                ComponentTypesView.instance.getAllComponents();
+                RegistryViewLoader.refresh();
```
dgolovin, line 318: Refresh call is not needed.

src/extension.ts
```diff
@@ -142,6 +143,14 @@ export async function activate(extensionContext: ExtensionContext): Promise<any>
         OdoImpl.Instance.loadWorkspaceComponents(event);
     });
 
+    void ComponentTypesView.instance.getAllComponents();
+
+    ComponentTypesView.instance.subject.subscribe((value: string) => {
```
dgolovin, line 148: This code is not needed.

` void ComponentTypesView.instance.getAllComponents();` call will send `refresh` event for the subject. If there is an instance of RegistryViewer it gets 'refresh' event and going to be updated.

dgolovin (changes requested):

src/registriesView.ts
```diff
@@ -97,6 +103,45 @@ export class ComponentTypesView implements TreeDataProvider<ComponentType> {
...
+    }
+
+    public getListOfRegistries(): Registry[] {
+        return this.registries;
+    }
+
+    public getAllComponents(): void {
+        let isError = false;
+        this.compDescriptions.clear();
+        void getInstance().getCompTypesJson().then(async (devFileComponentTypes: DevfileComponentType[]) => {
```
dgolovin, line 117: There should be catch in the chain for getCompTypesJson() call. Catch callback should show error message with info that odo could not pull component types form registry.

dgolovin (changes requested):
One last change is needed.

dgolovin (approved):

Issue #2486, opened 2022-07-20
Fix UI responsiveness for Add Cluster screen


mohitsuman:
Currently the cards present in the `Add Cluster` webview are not fully responsive. We need to update the css and make it responsive.

PR #2501, opened 2022-08-03
fixed ui break through css

msivasubramaniaan:

Fix: https://github.com/redhat-developer/vscode-openshift-tools/issues/2486


src/webview/cluster/app/cluster.tsx
```diff
@@ -16,7 +16,8 @@ import {
     Tooltip,
     Typography
 } from '@material-ui/core';
-
+import clsx from 'clsx';
+import { CardBody, CardFooter } from '@patternfly/react-core';
```
mohitsuman, line 20: @msivasubramaniaan Let's just use material UI and remove the patternfly dependency for now.

mohitsuman (approved):
LGTM

PR #2625, opened 2022-11-04
added switch and delete icons on manage contexts show-pick

msivasubramaniaan:

**Ref:**
image


mohitsuman (dismissed):

package.json
```diff
@@ -364,8 +364,8 @@
 				"category": "OpenShift"
 			},
 			{
-				"command": "openshift.explorer.switchContext",
-				"title": "Switch Contexts",
+				"command": "openshift.explorer.manageContext",
+				"title": "Manage Contexts",
```
dgolovin, line 368: The primary purpose of this feature to switch contexts, so it should be clear what it is and IMO old name should stay.

dgolovin (changes requested):
The same issues explained in  review for https://github.com/redhat-developer/vscode-openshift-tools/pull/2678 should be addressed the same way.

src/openshift/cluster.ts
```diff
@@ -103,20 +107,55 @@ export class Cluster extends OpenShiftItem {
...
+                    })
+            } else {
+                let selection: readonly QuickPickItem[] | undefined;
+                quickPick.onDidChangeSelection((selects) => {
+                    selection = selects;
+                });
+                quickPick.onDidAccept(async () => {
+                    if (selection && selection.length > 0) {
+                        const choice = Array.from(selection).pop();
+                        await Cluster.odo.execute(Command.setOpenshiftContext(choice.label));
```
dgolovin, line 133: add `quickPick.hide();` before command execution to hide quickPickInput, because if won't hide by itself in this case.

src/openshift/cluster.ts
```diff
@@ -103,20 +107,55 @@ export class Cluster extends OpenShiftItem {
...
+                        resolve(`Cluster context is changed to: ${choice.label}`);
+                    }
+                });
+                quickPick.onDidTriggerItemButton(async (event: QuickPickItemButtonEvent<QuickPickItem>) => {
+                    if (event.button === deleteBtn) {
+                        await window.showInformationMessage('Are you sure you want to delete this cluster information from kubeconfig?', 'Yes', 'No')
```
dgolovin, line 139: should assign awaited value

`const answer = await ...`

then you do not need to use `then` callback and code would look much cleaner

src/openshift/cluster.ts
```diff
@@ -103,20 +107,55 @@ export class Cluster extends OpenShiftItem {
...
+                });
+                quickPick.onDidAccept(async () => {
+                    if (selection && selection.length > 0) {
+                        const choice = Array.from(selection).pop();
+                        await Cluster.odo.execute(Command.setOpenshiftContext(choice.label));
+                        resolve(`Cluster context is changed to: ${choice.label}`);
+                    }
+                });
+                quickPick.onDidTriggerItemButton(async (event: QuickPickItemButtonEvent<QuickPickItem>) => {
+                    if (event.button === deleteBtn) {
```
dgolovin, line 138: this `if` is not needed, because there is only one ItemButton.

src/openshift/cluster.ts
```diff
@@ -103,20 +107,55 @@ export class Cluster extends OpenShiftItem {
...
+                        }
+                        resolve(null);
+                    })
+            } else {
+                let selection: readonly QuickPickItem[] | undefined;
+                quickPick.onDidChangeSelection((selects) => {
+                    selection = selects;
+                });
+                quickPick.onDidAccept(async () => {
+                    if (selection && selection.length > 0) {
```
dgolovin, line 131: Not needed if selection default value is empty array.

src/openshift/cluster.ts
```diff
@@ -103,20 +107,55 @@ export class Cluster extends OpenShiftItem {
...
+                        const choice = Array.from(selection).pop();
```
dgolovin, line 132: this could be just `const choice = selection[0]`.

dgolovin (changes requested):

dgolovin:
Squashed and pushed here ahttps://github.com/redhat-developer/vscode-openshift-tools/pull/2684.

PR #2630, opened 2022-11-05
use preference instead of registry list to fetch registries

### LINKED ISSUE #2542: Use odo preference view for registry loading
With odo v3, `odo registry list -o json` has been replaced with `odo preference view -o json`. The webview to load Component Registry View should implement the same.

mohitsuman:
Signed-off-by: Mohit Suman <mohit.skn@gmail.com>

Fix #2542


dgolovin (approved):
Good catch.

Issue #2689, opened 2022-12-02
Update @material-ui/core package


mohitsuman:
v4 is no more supported, update to v5 material UI.

PR #2691, opened 2022-12-02
upstream material-ui to v5

msivasubramaniaan:

Fix: https://github.com/redhat-developer/vscode-openshift-tools/issues/2689


package.json
```diff
@@ -119,15 +119,12 @@
 		"yaml": "^1.10.2"
 	},
 	"devDependencies": {
-		"@emotion/react": "^11.7.1",
-		"@emotion/styled": "^11.6.0",
-		"@material-ui/core": "^4.12.3",
```
mohitsuman, line 124: @msivasubramaniaan As we have removed this, should the reference of this in files be also updated such as here: https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/webview/cluster/app/cluster.style.tsx#L6

We need to be update all the location where the references are added for old dependency.

mohitsuman (changes requested):

msivasubramaniaan:
@mohitsuman as the reference of createStyles from v4 to v5 internally taken care through @mui/systems. Now the build has success

package.json
```diff
@@ -119,15 +119,12 @@
...
-		"@material-ui/icons": "^4.11.3",
-		"@material-ui/lab": "^4.0.0-alpha.60",
+		"@emotion/react": "^11.10.5",
+		"@emotion/styled": "^11.10.5",
 		"@mui/core": "^5.0.0-alpha.54",
 		"@mui/icons-material": "^5.2.5",
 		"@mui/lab": "^5.0.0-alpha.61",
-		"@mui/material": "^5.4.1",
+		"@mui/material": "^5.10.16",
```
mohitsuman, line 127: Let's keep it to the latest version 5.11.0

package.json
```diff
@@ -119,15 +119,12 @@
...
-		"@emotion/react": "^11.7.1",
-		"@emotion/styled": "^11.6.0",
-		"@material-ui/core": "^4.12.3",
-		"@material-ui/icons": "^4.11.3",
-		"@material-ui/lab": "^4.0.0-alpha.60",
+		"@emotion/react": "^11.10.5",
+		"@emotion/styled": "^11.10.5",
 		"@mui/core": "^5.0.0-alpha.54",
 		"@mui/icons-material": "^5.2.5",
 		"@mui/lab": "^5.0.0-alpha.61",
```
mohitsuman, line 126: We should also update the @mui/lab version to latest 5.0.0-alpha.112

package.json
```diff
@@ -119,15 +119,12 @@
...
 	"devDependencies": {
-		"@emotion/react": "^11.7.1",
-		"@emotion/styled": "^11.6.0",
-		"@material-ui/core": "^4.12.3",
-		"@material-ui/icons": "^4.11.3",
-		"@material-ui/lab": "^4.0.0-alpha.60",
+		"@emotion/react": "^11.10.5",
+		"@emotion/styled": "^11.10.5",
 		"@mui/core": "^5.0.0-alpha.54",
 		"@mui/icons-material": "^5.2.5",
```
mohitsuman, line 125: Similarly with the icons-material to latest.

package.json
```diff
@@ -119,15 +119,12 @@
...
 		"@mui/lab": "^5.0.0-alpha.61",
-		"@mui/material": "^5.4.1",
+		"@mui/material": "^5.10.16",
 		"@patternfly/react-core": "^4.198.19",
```
mohitsuman, line 128: @msivasubramaniaan are we using anything patternfly in our codebase ?

package.json
```diff
@@ -119,15 +119,12 @@
...
 		"@rjsf/material-ui": "^3.2.1",
```
mohitsuman, line 129: Similarly are we using this module anywhere ? As this material UI still point to v4 and we have to move to v5. So need to check if this is needed.

mohitsuman (changes requested):

mohitsuman (approved):

Issue #2681, opened 2022-11-28
Cancel git clone process if user exists Import from Git workflow


mohitsuman:
Currently during Import from Git workflow, if the user cancels the import from git, the cloning process keeps on running in the background. For multiple `import from git` processes, this can continue. We should find a better way to handle this workflow so that there is no process running concurrently and consuming memory.

dgolovin:
We need an approach to deal with sudden disposal of webview based editors, because when user closes this king of editors there is no way to stop it. There is no way to cancel closure. All we get is postmortem event didDisposed.

PR #2692, opened 2022-12-02
2681 cancel git clone process if user exits import from git workflow

msivasubramaniaan:


mohitsuman:
This will Fix #2681

src/webview/git-import/gitImportLoader.ts
```diff
@@ -276,18 +292,65 @@ function validateDevFilePath(event: any) {
...
+                        return reject(err);
+                    }
+                    resolve();
+                });
+            }
+        });
+    });
+};
+
+function deleteDirectory(dir) {
```
dgolovin, line 333:
```
const fs = require('fs')

// directory path
const dir = 'temp'

// delete directory recursively
fs.rmdir(dir, { recursive: true }, err => {
  if (err) {
    throw err
  }

  console.log(`${dir} is deleted!`)
})
```

dgolovin (changes requested):

dgolovin (dismissed):

dgolovin (changes requested):
See my comment about deleting directory.

src/webview/git-import/gitImportLoader.ts
msivasubramaniaan: Done with recursive call for removing the directory

dgolovin (changes requested):

dgolovin (approved):

PR #2706, opened 2022-12-13
handled error case if no devfile registry available

msivasubramaniaan:

**Post Fix:**
image

### DIFF
--- a/src/webview/devfile-registry/registryViewLoader.ts
+++ b/src/webview/devfile-registry/registryViewLoader.ts
@@ -132,7 +132,7 @@ export default class RegistryViewLoader {
 
 function getAllComponents(eventActionName: string, url?: string, error?: string) {
     let registries = ComponentTypesView.instance.getListOfRegistries();
-    if (!registries) {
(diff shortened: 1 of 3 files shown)


src/webview/devfile-registry/registryViewLoader.ts
```diff
@@ -132,18 +132,27 @@ export default class RegistryViewLoader {
...
-        {
-            action: eventActionName,
-            compDescriptions: Array.from(componentDescriptions),
-            registries: registries,
-            errorMessage: error
+    if (!registries) {
+        panel?.webview.postMessage(
+            {
+                action: eventActionName,
+                errorMessage: 'No Devfile registries configured'
```
mohitsuman, line 139: @msivasubramaniaan Instead of showing the Devfile Webview with this message, we can just load the Error Notification of VSCode saying `No devfile registry is accessible.`

mohitsuman (commented):
Need to see if vscode error message or webview one makes sense.

dgolovin (approved):

PR #2720, opened 2023-01-03
change syntax highlighting style based on theme selection

msivasubramaniaan:

**Post Fix:**
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -285,7 +285,6 @@
 		"onCommand:clusters.openshift.deployment.openConsole",
 		"onCommand:clusters.openshift.imagestream.openConsole",
 		"onCommand:openshift.componentTypesView.registry.openInView",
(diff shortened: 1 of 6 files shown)


dgolovin (changes requested):
Do not use vscode command for this. The main reason for that is whenever user changes the theme a telemetry event is going to be triggered. This event would not carry any interesting information related to using our tools. If webview based editor required to react to theme change implement it in specific webview based editor for now.

msivasubramaniaan:
> Do not use vscode command for this. The main reason for that is whenever user changes the theme a telemetry event is going to be triggered. This event would not carry any interesting information related to using our tools. If webview based editor required to react to theme change implement it in specific webview based editor for now.

React doesn't have any theme change event as of now. is there any way to restrict telemetry event for an one particular command?

dgolovin (approved):

PR #2545, opened 2022-09-21
Upgrade extension tester to 5.x and clean up ui tests

odockal:
Signed-off-by: Ondrej Dockal <odockal@redhat.com>

Issue #2483, opened 2022-07-20
Add UI tests covering command palette offering 


odockal:

dgolovin:
There is no need for testing if commands are in palette through UI, because they are not there only if you have declared it in package.json and then did not registered a handler for it in extension activation method. 

Unit test [here](https://github.com/redhat-developer/vscode-openshift-tools/blob/d6f7b8ece5cf7409b75f223d4b19a977f4e218a8/test/unit/extension.test.ts#L109) covers those kind of errors.

PR #2721, opened 2023-01-04
Adding first part of command palette test

lgrossma:
Signed-off-by: Lukas Grossmann <lgrossma@redhat.com>


odockal:
@lgrossma after #2545 is merged, UI tests will be reenabled and we can see your tests on GA.

lgrossma:
Looked at that PR. I will need to move the test to a different folder..

odockal:
@lgrossma Ok. I have merged other PR, please rebase. and yes, I have removed duplicated files from smoke directory.

lgrossma:
@odockal should be ready for review. Also updated openshift.ts test which was missing one part that was available in the deleted duplicate

test/ui/suite/command-palette.ts
```diff
@@ -0,0 +1,37 @@
...
+        });
+
+        it('Shows available commands', async () => {
+            expect((await prompt.getQuickPicks()).length > 1);
+        });
+
+        describe('Available Commands', () => {
+            COMMANDS.forEach(command =>
+                it(command, async () => {
+                    expect(await prompt.findQuickPick(command)).not.undefined;
```
dgolovin, line 32: This looks like you are testing if your COMMANDS array is correct. It does not add any value to assure quality. If developers of extension decide to remove one of your COMMANDS form palette, this test fails and it does not mean you found a problem in extension.

What would make sense here is to verify, for example, if a command declared in package.json does not have a actual command handler registered when extension is activated. But that is already covered in UI tests and does not need verification using UI test.

lgrossma: Exactly as you say. Further plan for this was to pick commands from the package.json instead of the COMMANDS array. The solution with the array is only temporary to have at least some test.

dgolovin: Ok, but what are you going to test? If you going to read package.json and analyze if command should appear in palette it is a waste of processing time, because that is supposed to be tested by Microsoft and test is going to work always. If extension is activated all the commands guaranteed to be loaded and appear on pallet if they configured to appear.

It does not mean we do not need UI tests at all, I am not saying that. I am saying we covered tests related to commands verification and you should not waste your time on that.

lgrossma: This PR can be closed then. Can you provide more info about this issue https://github.com/redhat-developer/vscode-openshift-tools/issues/2483 if you already covered tests related to commands verification?

dgolovin: Done [here](https://github.com/redhat-developer/vscode-openshift-tools/issues/2483#issuecomment-1373999005)

Issue #2727, opened 2023-01-06
Registry view shows progress forever after it is opened first time


dgolovin:
I see it every time I open Registry viewer first time.

msivasubramaniaan:
@dgolovin I haven't seen any such delayed while open the Registry view on first time. 
![devfileRegistry-open-first-time](https://user-images.githubusercontent.com/93245779/210963198-fad1485e-c5fc-428c-9155-f083f271f485.gif)

Kindly update your out folder and recheck again and let me know still you face the same issue.

PR #2730, opened 2023-01-06
re initialize the form values when git url text box updated

msivasubramaniaan:


dgolovin (changes requested):
Now 'Create Component' button is always disabled.

dgolovin (approved):
Going to approve, because the issues is seems to be related to #2727

PR #2731, opened 2023-01-06
change button color based on enable/disable

msivasubramaniaan:

**Post Fix:**

- On Disabled
image
- On Enabled
image


dgolovin (changes requested):
image
Disabled button is there, but it always stays disabled.

dgolovin:
the problem is if deviles are not available for some reason from registry 'Create Component' button is never gets enabled and there is no messages what is wrong.

dgolovin (approved):

dgolovin:
Approving this one, because the problem is not related to this change.

PR #2743, opened 2023-01-18
Use proxyURL to log into sandbox using Red Hat auth token

### LINKED ISSUE #2519: Use sandbox proxyURL with idToken to work with sandbox cluster
Sandbox getSignupStatus  request to signup service returns proxyURL among other attributes for sandbox instance. Extension should use this proxy with idToken retrieved with Red Hat Authentication extension to communicate with Sandbox instance.

dgolovin:
Getting resources seems to work fine through the proxy, but there are two problems to solve:
1. idToken lives 900 seconds and then have to be refreshed in kubeconfig (that probably affects live connections through the proxy that already exists at the moment when refresh happens)
2. odo dev cannot sync files through the proxy

Fix #2519.

### DIFF
--- a/src/openshift/sandbox.ts
+++ b/src/openshift/sandbox.ts
@@ -29,6 +29,7 @@ export interface SBSignupResponse {
   givenName: string;
   status: SBStatus;
   username: string;
+  proxyURL: string;
 }
 
 export interface SBResponseData {
--- a/src/webview/cluster/clusterViewLoader.ts
+++ b/src/webview/cluster/clusterViewLoader.ts
@@ -109,7 +109,7 @@ async function clusterEditorMessageListener (event: any ): Promise<any> {
                 } else {
                     if (signupStatus.status.ready) {
                         const oauthInfo = await sandboxAPI.getOauthServerInfo(signupStatus.apiEndpoint);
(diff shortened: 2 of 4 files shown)


src/openshift/sandbox.ts
```diff
@@ -29,6 +29,7 @@ export interface SBSignupResponse {
   givenName: string;
   status: SBStatus;
   username: string;
+  proxyURL: string;
```
dgolovin, line 32: new attribute in reg service response

src/webview/cluster/clusterViewLoader.ts
```diff
@@ -173,7 +173,7 @@ async function clusterEditorMessageListener (event: any ): Promise<any> {
         case 'sandboxLoginUsingDataInClipboard':
             const telemetryEventLoginToSandbox = new ExtCommandTelemetryEvent('openshift.explorer.addCluster.sandboxLoginUsingDataInClipboard');
             try {
-                const result = await Cluster.loginUsingClipboardToken(event.payload.apiEndpointUrl, event.payload.oauthRequestTokenUrl);
+                const result = await Cluster.loginUsingClipboardToken(event.payload.proxyUrl, (sessionCheck as any).idToken);
```
dgolovin, line 176: Login to the sandbox using proxyUrl and token from RedHat SSO. No need to get sandbox token from browser.

vrubezhny:
Incorporated by https://github.com/redhat-developer/vscode-openshift-tools/pull/4439.

PR #2748, opened 2023-01-24
2658 fix npm install warnings for packages and their dependencies

msivasubramaniaan:


rgrunber (commented):
The part of the change that updates `npm-check-updates` looks fine to me.

However it doesn't look like the indentation is done correctly. Try with `"editor.renderWhitespace": "all"` and you'll see spaces still get mixed in.

The file is mostly using tabs with a size of 4 spaces, so that should probably be used.

PR #2749, opened 2023-01-24
added enum for http status code and send corresponding messages based on the code

### LINKED ISSUE #2724: This line give false error 'rate limit is exceeded' in case of connectivitiy issues
https://github.com/redhat-developer/vscode-openshift-tools/blob/d6f7b8ece5cf7409b75f223d4b19a977f4e218a8/src/webview/git-import/gitImportLoader.ts#L112

To test try to unplug the network cable or disable WiFi or both.

msivasubramaniaan:
Fix #2724.

### DIFF
--- a/src/webview/git-import/gitImportLoader.ts
+++ b/src/webview/git-import/gitImportLoader.ts
@@ -34,6 +34,11 @@ interface CloneProcess {
     error: string
 }
 
+enum StatusCode {
+    INTERNAL_SERVER_ERROR = 500,
+    RATE_LIMIT_EXCEEDED = 403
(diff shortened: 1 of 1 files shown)


src/webview/git-import/gitImportLoader.ts
```diff
@@ -34,6 +34,11 @@ interface CloneProcess {
     error: string
 }
 
+enum StatusCode {
```
msivasubramaniaan, line 37: update the enum based on the error code in future

dgolovin (approved):

PR #2750, opened 2023-01-24
Pin 3rd-party actions to SHA1

fbricon:
Hi!

Following the [GH Action Security Hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) guide we should use the commit SHA instead of the `branch` or `tag` for any third-party untrusted action.

This PR was submitted by a script.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -38,7 +38,7 @@ jobs:
 
       # Run tests
       - name: Build and Test
-        uses: GabrielBB/xvfb-action@v1.0
+        uses: GabrielBB/xvfb-action@fe2609f8182a9ed5aee7d53ff3ed04098a904df2 #v1.0
         env:
             NODE_OPTIONS: --max_old_space_size=16384
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -38,23 +38,23 @@ jobs:
...
+        uses: GabrielBB/xvfb-action@fe2609f8182a9ed5aee7d53ff3ed04098a904df2 #v1.0
         env:
             NODE_OPTIONS: --max_old_space_size=16384
         with:
           run: npm run test:coverage --silent
 
       # Upload coverage to codecov.io
       - name: Codecov
-        uses: codecov/codecov-action@v1.0.12
+        uses: codecov/codecov-action@07127fde53bc3ccd346d47ab2f14c390161ad108 #v1.0.12
```
rgrunber, line 49: @fbricon , Minor nitpick but https://github.com/marketplace/actions/codecov is a verified creator. Is this updating the reference regardless of creator status ?

fbricon: > The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

Only `actions` and `docker` are trusted for now. I'll ask around if it's OK to relax the req depending on the publisher status

rgrunber (approved):

rgrunber:
@msivasubramaniaan , if you're fine with this, lets merge it.

PR #2758, opened 2023-01-27
Move Components view at the top of OpenShift activity views

### LINKED ISSUE #2576: Application Explorer options (buttons) are missing and only "loaded-context" item is present when no kubeconfig is set
In a situation when I do not have any kubeconfig on ~/.kube/config or env. var set. the extension shows only "loaded-context" tree item in application explorer instead of Login, Create component, etc.

### LINKED ISSUE #2756: Components view not visible enough
When I open the plugin view, the views on left bar are, in this order:
- Application Explorer
- Components
- Devfile Registries
- Debug Sessions

As a developer, I would expect that the Components view is the first one in the list, as this is the first one I want to focus on: I first need to "Start Dev" from the Components view before there are any information available in the "Application Explorer".

This is especially true when running on podman, where the Application Explorer is not necessary.

I understand that the dev needs to connect to a cluster before to be able to work, but this should, IMHO, be triggered by an alert (pop-pu, etc) and not by the fact that the Application Explorer is first in the list.

dgolovin:
Fix #2756.

### DIFF
--- a/package.json
+++ b/package.json
@@ -829,12 +829,12 @@
 		"views": {
 			"openshiftView": [
 				{
-					"id": "openshiftProjectExplorer",
-					"name": "Application Explorer"
-				},
-				{
 					"id": "openshiftComponentsView",
 					"name": "Components"
+				},
+				{
+					"id": "openshiftProjectExplorer",
+					"name": "Application Explorer"
 				},
 				{
 					"id": "openshiftComponentTypesView",


mohitsuman:
@dgolovin you mean fix for #2756 ?
But do we need to rearrange the order given these view can be replaced by drag and drop. Now for the default view, the Application Explorer view is useful as it will show contents if the cluster already has deployments in it, so it makes sense to be at top for default. We have similar experience in Cloud Run, K8s extensions for reference.

dgolovin:
@mohitsuman Should we just retire Application explorer and redirect to Clusters view instead? We can add 'Show in Cluster Exporer' context menu item for components running in dev mode or something like that.

dgolovin:
@mohitsuman I agree, that the order of views is not really an issue and can be always changed by DnD.

PR #2766, opened 2023-02-16
2592 recommend openshift extension if workspace has devfileyaml file present

msivasubramaniaan:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,7 +11,7 @@ indent_style = space
 indent_size  = 4
 
 # We recommend you to keep these unchanged
-end_of_line              = lf
+end_of_line              = crlf
 charset                  = utf-8
 trim_trailing_whitespace = true
 insert_final_newline     = true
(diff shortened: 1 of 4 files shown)


dgolovin (changes requested):
This is issue is not about creating comands or any new features in extension. This is about how to make vscode to recommend our extension for workspace with devifle.yaml file in. So it most likely related to language extension point, where you can define new language or extends existing language definition.

Try to create 'test.py' file and you will se recommendation message right away.

Not sure if we can force vscode show our extension in recommendations.

rgrunber:
As discussed, we probably want to place the PR in https://github.com/redhat-developer/vscode-yaml since it's a good bet, that would be installed if editing a yaml file.

Otherwise, the only mechanism I know of to recommend an extension, is only for individual workspaces (`.vscode/extensions.json`), or maybe by increasing visibility with the `categories`/`keywords`, fields, https://github.com/redhat-developer/vscode-openshift-tools/blob/3b510ca6e51c375ec2c707bedac6d7c25db5eb3b/package.json#L30-L44

msivasubramaniaan:
As this handled from vscode-yaml extension, closing this PR

PR #2775, opened 2023-02-21
Reduce size of the walkthrough gif videos.

rgrunber:
- gifsicle --resize-fit-height 720 -i [in] -O3 --colors 32 -o [out]
- 30MB -> 11MB

This reduced the overall size of the vsix by about the same. Regular vsix is ~118MB, and this change brings it down to ~100MB.


msivasubramaniaan (approved):
LGTM

Issue #2697, opened 2022-12-06
Create component new WebView workflow


mohitsuman:
Currently, to create components, there are 2 ways:
1. Create Component Button: Opens up a quick pick to select devfile
2. Devfile Webview workflow: Opens up webview to select devfile

We need to figure out a way to merge these 2 workflows together so that for the user, it should be simplified in selecting the devfile registry and proceed forward.

msivasubramaniaan:
@mohitsuman and @dgolovin 

Here I have moved `Create Component` button flow into `Devfile Registry` WebView flow.

Now user can create component with/without starter project initialization through checkbox option. Please have a look into the attached Gifs

- Without Starter Project Initialization

![create-component-new-flow-without-starter-project-demo](https://user-images.githubusercontent.com/93245779/220642109-99b31db2-a97a-436a-9a75-9461ce78b2b1.gif)

- With Starter Project Initialization

![create-component-new-flow-demo](https://user-images.githubusercontent.com/93245779/220642136-81b6e62e-3a31-4c45-a1bd-c46b2ed24e86.gif)

dgolovin:
Where the component is created in local filesystem?

msivasubramaniaan:
> Where the component is created in local filesystem?

That will follow the option to selecting workspace folder. If empty workspace folder and it picks directly otherwise ask user to select through VSCode command.

As discussed earlier WebView doesn't support the folder selection option due to security issue on the browser.

msivasubramaniaan:
@dgolovin Can you please review this PR?

PR #2776, opened 2023-02-22
moved create component into WebView flow

msivasubramaniaan:
Related to https://github.com/redhat-developer/vscode-openshift-tools/issues/2697.

### DIFF
--- a/package.json
+++ b/package.json
@@ -432,7 +432,7 @@
 				}
 			},
 			{
-				"command": "openshift.component.createFromLocal",
+				"command": "openshift.componentTypesView.registry.openInView",
 				"title": "New Component",
 				"category": "OpenShift",
 				"icon": {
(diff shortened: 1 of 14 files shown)


dgolovin (changes requested):
Just redirecting to registry view does not give any clue what to do next. There should be different view that would behave similar to git import and let use 'odo analyze' command as well.

msivasubramaniaan:
> Just redirecting to registry view does not give any clue what to do next. There should be different view that would behave similar to git import and let use 'odo analyze' command as well.

I hope one more view no need. As our moto to make 2 different approach into 1.

If we are plan for new view then we have to call that view on both side. (ie. vscode -> new view, Registry view -> new view). Instead of that I have redirect the vscode approach into Web view(Registry view). From where we can create a component either use of Starter Project/not.

@mohitsuman Please comment on your view here

Issue #2780, opened 2023-03-01
Test case covering creating a component from component's view


odockal:
Create a component using Component View's 'Create a component' button workflow.

rgrunber:
See https://github.com/redhat-developer/vscode-openshift-tools/blob/main/images/walkthrough/createComponent.gif for a sample of this workflow

datho7561:
Wait, I thought it was the button here:
![CreateComponentFromComponentsView](https://user-images.githubusercontent.com/22376627/226671952-fc225214-1477-41ed-890e-b6553d7a2ae6.png)

PR #2803, opened 2023-03-20
Add service to component (`odo add binding`)

### LINKED ISSUE #2668: Support Adding Service to component workflow
Currently, for apps where we need to add `odo add binding`, the extension does NOT support it. We need to add a workflow to support binding if the corresponding devfile has it.

Ref: https://odo.dev/docs/user-guides/advanced/connecting-a-service#step-3-connect-the-application-to-the-mongodb-service

datho7561:
Right click on a component in the components view and select "Bind Service" in order to bind the component to an Operator-backed service.

In order for this to work:
- the service binding operator must be installed on the cluster
- an operator that manages instances of services must be installed (I tested with the RHOAS operator)
- there must be an instance of the service that is managed by the operator in the current project

Other things in this PR:
- Add a walkthrough GIF of using the UI
- Add a smoke test to check that the "Bind Service" context menu item exists
- Fix the smoke tests. This includes fixing the existing test case for https://github.com/redhat-developer/vscode-openshift-tools/issues/2780, even though we are going to rewrite that as a WebView UI soon. I did this, since for the "Bind Service" smoke test, we need a component present in the components view.

Closes #2668


datho7561:
Need to add integration tests for `odo add binding`

msivasubramaniaan:
Here are the comments which needs to be addressed in order to UI

- The webview should align with VSCode theme in terms of button color, font size, font family
image
- Binding name accepts the invalid names as well
image
- In order to showing the error message while loading, we can make the button disabled and show the error message based on user input
- Don't open multiple service binding webview for same component

msivasubramaniaan (changes requested):
Here are the comments which needs to be addressed in order to UI

- The webview should align with VSCode theme in terms of button color, font size, font family
image
- Binding name accepts the invalid names as well
image
- In order to showing the error message while loading, we can make the button disabled and show the error message based on user input
- Don't open multiple service binding webview for same component

src/openshift/component.ts
```diff
@@ -203,6 +205,63 @@ export class Component extends OpenShiftItem {
         return Component.dev(component, 'podman');
     }
 
+    @vsCommand('openshift.component.binding.add')
+    static async addBinding(component: ComponentWorkspaceFolder) {
+        const odo: odo3.Odo3 = odo3.newInstance();
+
+        const services = await odo.getBindableServices();
+        if (!services || services.length === 0) {
+            throw new Error('No bindable services are available');
```
JessicaJHee, line 214: Could we direct the user here in the message to somewhere where they can add a service?

src/webview/add-service-binding/app/addServiceBindingForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                            {!isBindingNameValid(bindingName) &&
+                                                selectedService === '' && <> and </>}
+                                            {!isBindingNameValid(bindingName) && (
+                                                <>set a valid binding name</>
+                                            )}
+                                            .
+                                        </Alert>
+                                    ) : (
+                                        <Button variant="contained" size="large" type="submit">
+                                            Add Service Binding
```
JessicaJHee, line 187: This ends up showing up as all capitalized, maybe change the style to be consistent with the other webviews?

JessicaJHee (commented):
Works well from my testing! Could consider adding a notification when the service was successfully binded (either in the webview and not have it close on it own, or a vscode window popup?), I wasn't sure if everything worked when the webview closed.

Another small thing with the add service binding button, could we keep that disabled but visible when the user hasn't picked the service or service name instead of replacing the error box? But it's not a big deal so if it takes a long time to implement then I don't think it's worth changing.

src/webview/add-service-binding/app/addServiceBindingForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+    };
+
+    return (
+        <ThemeProvider theme={theme}>
+            <Container maxWidth="md">
+                {availableServices && componentName ? (
+                    <form onSubmit={handleSubmit}>
+                        <Stack spacing={3} marginTop={3}>
+                            <FormLabel>
+                                Bind Service to <code>{componentName}</code>
```
msivasubramaniaan, line 116: Instead of `componentName` it showed the **component folder name**
image

src/openshift/component.ts
```diff
@@ -203,6 +205,63 @@ export class Component extends OpenShiftItem {
         return Component.dev(component, 'podman');
     }
 
+    @vsCommand('openshift.component.binding.add')
+    static async addBinding(component: ComponentWorkspaceFolder) {
```
msivasubramaniaan, line 209: Show progress bar with notification until webview opened, currently there were no clue when clicked `Bind Service`

msivasubramaniaan (changes requested):
There was an error in dev console log while loading the Bind service webview
image

src/openshift/component.ts
datho7561: I added a button that links to the operator backed service catalog.

src/openshift/component.ts
datho7561: I misunderstood at first, but it should be fixed now.

rgrunber (approved):
Works for me. We should merge and improve afterwards if there are any other issues.

mohitsuman:
@datho7561 have we added segment event to this command workflow. We need to update around:
- User opens the webview
- Submits the form

datho7561:
I added two new events to track the usage of the "Bind Service" UI: `startAddBindingWizard` and `finishAddBindingWizard`. We also have the telemetry of when the command fails or is run successfully.

datho7561:
@msivasubramaniaan if you think it's good, could you please +1 when you have time (no rush)

datho7561:
@rgrunber let me know if you think it's good. However, I think we are blocked on Muthu giving a +1 in order to merge the PR

rgrunber (approved):

PR #2804, opened 2023-03-21
[WIP] test creating a component from the components view

datho7561:


msivasubramaniaan:
@datho7561 as we are planning to move new webview based UI for component creation and existing approach will be removed. So please hold this test case for create Component.

PR #2829, opened 2023-04-11
added update helm command

msivasubramaniaan:


rgrunber:
I played around with this by clearing the contents of files in `~/.cache/helm/repository/` . Having this ensured they were refreshed appropriately.

PR #2834, opened 2023-04-12
If cluster is not up during login, pick new URL

### LINKED ISSUE #709: `Login to Cluster' command should check if cluster is started
Now it just follows the regular flow that is doomed to fail at the final step.

datho7561:
When the user is attempting to login to a cluster, check if the cluster is accessible before prompting for credentials. To check if the cluster is accessible, send a GET request to the cluster API URL; if the API endpoint cannot be found or returns 5XX, display a warning message and return to the step where the user inputs the cluster URL.

This means that if the user tries to login to a cluster that is not running, they don't have to click through the rest of the inputs before receiving this message. Also, if the cluster is not available since the user mistyped the URL, they have the chance to type it in again.

Fixes #709


msivasubramaniaan:
@datho7561 Can you add the video recording for reference?

src/openshift/cluster.ts
```diff
@@ -275,9 +276,34 @@ export class Cluster extends OpenShiftItem {
...
+                // disable cert checking. crc uses a self-signed cert,
+                // which means this request will always fail on crc unless cert checking is disabled
+                const clusterApiResponse = await fetch(clusterURL, {
+                    strictSSL: false
+                });
+                // don't check for 2XX status, since we are expecting an "unauthorized" response from the server
+                if (500 <= clusterApiResponse.status && clusterApiResponse.status < 600) {
+                    void window.showWarningMessage(
+                        'Received 5XX response (server error) from the cluster API server',
+                    );
```
rgrunber, line 297: Is there documentation somewhere that states a cluster will always return a non-500 error if it is up ? I would think as long as `clusterApiReponse` has any value, it would indicate the cluster is up.

datho7561: Yeah, I haven't been able to replicate a case where the API returns 5XX when it's starting up, so I can probably remove this check

msivasubramaniaan:
@datho7561 if cluster not accessible it good to show the message but I feel it would be better to show some action that make cluster availability again. WDYT?

datho7561:
> show some action that make cluster availability again

Sure, I think it would be nice to present options to get the cluster running. Here's what I think we could show:
- Option to run `crc start` to start OpenShift Local
- Option to open a link to Dev Sandbox

I guess we can figure out from the URL the user entered if either of these apply, then prompt them to take that action. I'll take a look into adding that.

msivasubramaniaan:
> > show some action that make cluster availability again
> 
> Sure, I think it would be nice to present options to get the cluster running. Here's what I think we could show:
> 
> * Option to run `crc start` to start OpenShift Local
> * Option to open a link to Dev Sandbox
> 
> I guess we can figure out from the URL the user entered if either of these apply, then prompt them to take that action. I'll take a look into adding that.

Yeah. Simply we can redirect to the cluster webview page.

datho7561:
Okay cool, I've added a prompts that redirect to the corresponding pages in the "Create Cluster" webview.

msivasubramaniaan (approved):
LGTM

rgrunber:
@datho7561 change looks fine. Only odd thing I noticed were some rather large delays between entering, say `https://api.crc.testing:6443` and the notification asking whether to create the cluster / try another. I would just verify it behaves as expected in those cases.

datho7561:
> I noticed were some rather large delays between entering, say https://api.crc.testing:6443 and the notification asking whether to create the cluster / try another

Yeah, that's expected unfortunately. It takes a while to figure out the URL is unreachable.

PR #2836, opened 2023-04-17
fix updated Node.js 16 in GH workflow actions

msivasubramaniaan:

This PR resolves the following warnings on GH actions
image

### DIFF
--- a/.github/workflows/check-crc.yml
+++ b/.github/workflows/check-crc.yml
@@ -12,7 +12,7 @@ jobs:
             GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         steps:
             - name: Check Out Code
-              uses: actions/checkout@v2
+              uses: actions/checkout@v3
             - name: Get latest CRC version
               run: |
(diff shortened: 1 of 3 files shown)


rgrunber (approved):
LGTM. We seem to mostly use checkout without modifying parameters, so that should probably work the same way.

msivasubramaniaan:
> LGTM. We seem to mostly use checkout without modifying parameters, so that should probably work the same way.

Seems [gabrielbb/xvfb-action](https://github.com/redhat-developer/vscode-openshift-tools/blob/fbe9731e3d78fbbfc6662feca71b1772c73e82d5/.github/workflows/continuous-integration-workflow.yml#L41) was deprecated and used v1 still. Hope we need to update this as well

msivasubramaniaan:
@rgrunber I have moved [coactions/setup-xvfb](https://github.com/coactions/setup-xvfb) and resolved all the existing warnings. Please refer: https://github.com/redhat-developer/vscode-openshift-tools/actions/runs/4731705085. Hence I would merge this PR and close it.

PR #2837, opened 2023-04-17
Switch to first available namespace

### LINKED ISSUE #2180: Namespace non existent should be handled
If the current context is using a non existent namespace on the cluster, the Application Explorer is empty. We should handle this case:
- take the first available namespace
- if no namespaces are available display a link to create a new namespace/project

datho7561:
When the currently configured namespace doesn't exist, switch to the first available namespace.

Closes #2180

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -4,33 +4,33 @@
  *-----------------------------------------------------------------------------------------------*/
 
 import {
-    TreeDataProvider,
-    TreeItem,
+    Disposable,
     Event,
     EventEmitter,
-    Disposable,
+    ThemeIcon,
+    TreeDataProvider,
+    TreeItem,
+    TreeItemCollapsibleState,
     TreeView,
-    window,
+    Uri,
+    commands,
     extensions,
     version,
-    commands,
-    Uri,
-    TreeItemCollapsibleState,
-    ThemeIcon,
+    window,
 } from 'vscode';
 
-import * as path from 'path';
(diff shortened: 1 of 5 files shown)


datho7561:
I honestly have no idea if these integration tests (`npm urn test-integration`) are run anywhere. Most of the suite seems to be failing locally when I use an OpenShift Local (`crc`) instance.

datho7561:
@msivasubramaniaan @rgrunber this PR is ready for review

src/explorer.ts
```diff
@@ -198,22 +198,25 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
             //   * example is sandbox context created when login to sandbox first time
             // (3) there is namespace set in context and namespace exists in the cluster
             // (4) there is namespace set in context and namespace does not exist in the cluster
-            const pOrNs = await this.odo3.getNamespaces();
+            const namespaces = await this.odo3.getNamespaces();
```
rgrunber, line 201: According to the comments above, we're logged into a cluster. Can we ensure `isLoggedIn` is set to `true` as I describe in https://github.com/redhat-developer/vscode-openshift-tools/issues/2848#issuecomment-1520843795  to fix the issue there also ?

datho7561: I think Muthu's fix here: https://github.com/redhat-developer/vscode-openshift-tools/pull/2851 makes more sense

PR #2839, opened 2023-04-18
disable next button until user selects the installed location of local CRC

msivasubramaniaan:

`Next` button will be disabled until user selects the installation file.
image

### DIFF
--- a/src/webview/cluster/app/clusterView.tsx
+++ b/src/webview/cluster/app/clusterView.tsx
@@ -209,7 +209,7 @@ export default function addClusterView(props: ClusterViewProps) {
   };
 
   const handleDisabled = () => {
-    return (activeStep === 1 && fileName === '')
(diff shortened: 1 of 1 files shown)


src/webview/cluster/app/clusterView.tsx
```diff
@@ -209,7 +209,7 @@ export default function addClusterView(props: ClusterViewProps) {
   };
 
   const handleDisabled = () => {
-    return (activeStep === 1 && fileName === '')
+    return ((activeStep === 0 || activeStep === 1) && fileName === '')
         || (activeStep === 2 && pullSecretPath === '');
```
datho7561, line 213: With this PR,  I can skip over selecting the the pull secret file. I think you need to update the check on line `213` to check `activeStep === 1` instead of 2.

datho7561 (changes requested):

PR #2844, opened 2023-04-21
Create component new workflow

msivasubramaniaan:

Fix: https://github.com/redhat-developer/vscode-openshift-tools/issues/2697


datho7561:
A few things that I've noticed so far:
1. There are two "create component" buttons in the components section now; we should remove the old one along with the code associated with it
2. It was unclear whether the component that I was creating was being created in the folder I selected, or in a subfolder (i.e. `./the/folder/i/picked/component-name/devfile.yaml` vs `./the/folder/i/picked/devfile.yaml`). I think we should add a field description that says something like "the devfile and code will be added directly into this folder".
3. A point related to 2: if you are using the `"files.simpleDialog.enable": true,` VS Code setting, there is no option to create folders in the dialog. It might be nice to create a folder named after the component to hold the devfile and sample code, eg. `./the/folder/i/picked/component-name/devfile.yaml`
4. I expected the WebView to close when I click "Create Component", and it doesn't
5. When there are multiple starter examples, such as with Vert.X, I can select the different values (eg. I pick `vertx-crud-example`), but when I click anywhere outside the box in the webview, it resets to `vertx-http-example`. Because of this, I can only set up a project with whatever starter code is initially selected

src/webview/create-component/createComponentLoader.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+function validateComponentName(event: any) {
+    let validationMessage = OpenShiftItem.emptyName(`Required ${event.compName}`, event.compName.trim());
+    if (!validationMessage) validationMessage = OpenShiftItem.validateMatches(`Not a valid ${event.compName}.
+        Please use lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character`, event.compName);
+    if (!validationMessage) validationMessage = OpenShiftItem.lengthName(`${event.compName} should be between 2-63 characters`, event.compName, 0);
+    panel?.webview.postMessage({
+        action: event.action,
+        error: !validationMessage ? false : true,
```
datho7561, line 123: `Boolean(validationMessage)`

msivasubramaniaan:
> A few things that I've noticed so far:
> 
> 1. There are two "create component" buttons in the components section now; we should remove the old one along with the code associated with it
> 2. It was unclear whether the component that I was creating was being created in the folder I selected, or in a subfolder (i.e. `./the/folder/i/picked/component-name/devfile.yaml` vs `./the/folder/i/picked/devfile.yaml`). I think we should add a field description that says something like "the devfile and code will be added directly into this folder".
> 3. A point related to 2: if you are using the `"files.simpleDialog.enable": true,` VS Code setting, there is no option to create folders in the dialog. It might be nice to create a folder named after the component to hold the devfile and sample code, eg. `./the/folder/i/picked/component-name/devfile.yaml`
> 4. I expected the WebView to close when I click "Create Component", and it doesn't
> 5. When there are multiple starter examples, such as with Vert.X, I can select the different values (eg. I pick `vertx-crud-example`), but when I click anywhere outside the box in the webview, it resets to `vertx-http-example`. Because of this, I can only set up a project with whatever starter code is initially selected

Yes I am working on those only. I have added the work-in-progress label already. Anyway thanks for the points added.

datho7561:
> I have added the work-in-progress label already

Okay. Sorry about that. It's looking great so far!

msivasubramaniaan:
> > I have added the work-in-progress label already
> 
> Okay. Sorry about that. It's looking great so far!

NP 😊

msivasubramaniaan:
@datho7561 I have completed the full code flow. Please start reviewing now

datho7561:
Okay, sorry I'm reviewing a bit late. Here's some things I ran into:
- It's not clear if the devfile and sample code will be generated in the folder you select, or if a new subfolder will be generated that contains the devfile and sample code. I think it would be better to generate a new subfolder. eg. if you name your component `my-component` and pick the folder `/my/projects/folder`, create the folder `/my/projects/folder/my-component` and generate the devfile in this folder (`/my/projects/folder/my-component/devfile.yaml`). I think this is better, because the UI to pick folders in VS Code doesn't have the option to create folders, and it won't let you select a folder that doesn't exist:
Screenshot from 2023-04-24 16-30-26
- If I select a devfile that has exactly one starter project, then the dropdown to select the starter project appears, but it's empty. I think we should hide this dropdown if there is only one option.
Screenshot from 2023-04-24 16-34-40
- There is a field for setting the name of the application. However, throughout most of the code base, we assume the name of the application is `app`, since this is what `odo` does. I think it might be better to hardcode the name of the application to `app` and not have this field in the form at all
- I haven't had a chance to look through the source code, I'll get to that tomorrow morning

datho7561:
> If subfolder then what would be the name of that

The name of the subfolder should be the component name

> I don't see any empty of starter-project when only one starter project.

Maybe this only happens with light theme, let me double check...

> If it empty and it default taken as `component-name-app`

Okay, I understand this, but we don't display the app name anywhere or ask for the app name anywhere else, so I don't think it's worth asking for it. If we use a default app name, I think it makes sense to go with the one `odo` is using, which is `app`.

msivasubramaniaan:
> > I don't see any empty of starter-project when only one starter project.
> 
> Yep; this issue is that in the light theme, it's white text on a white background:
> 
> starter-project-name-white-on-white

Thanks for the good catch. Need to concentrate on css based themes

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,427 @@
...
+                                                color: '#EE0000',
+                                                marginTop: '1rem'
+                                            }}>
+                                            Component Name
+                                        </InputLabel>
+                                        <TextField
+                                            defaultValue={component.name}
+                                            error={component.error}
+                                            onChange={(e) => this.validateName(e.target.value, 'Comp')}
+                                            id='bootstrap-input'
```
datho7561, line 242: Correct me if I'm wrong, but I don't think we are using Bootstrap. I think it makes sense to rename the `id` (and the `htmlFor` in the `<InputLabel>`) to something like `component-name-input`

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+    render(): React.ReactNode {
+        const { application, autoSelectDisable, component, wsFolderItems, wsFolderPath, compDescriptions, selectedComponentDesc, selectedStarterProject, showLoadScreen } = this.state;
+        const optionList = compDescriptions.sort(ascName);
+        const filterList = selectedComponentDesc ? optionList.filter((value) => value.devfileData.devfile.metadata.name === selectedComponentDesc.devfileData.devfile.metadata.name &&
+            value.registry.name === selectedComponentDesc.registry.name) : [];
+        const getInputValue = filterList.length > 0 ? filterList[0].devfileData.devfile.metadata.displayName + '/' + filterList[0].registry.name : ''
+        return (
+            <>
+                {compDescriptions.length === 0 ? <LoadScreen title={'Loading the devfiles'} /> :
```
datho7561, line 218: consider turning this into an `if` statement instead of a ternary expression in the returned value. eg.
```tsx
if (compDescriptions.length === 0) {
  return <LoadScreen .../>
} else if (showLoadScreen) {
  return <LoadScreen .../>
} else {
  return <>
    // the main form UI
  </>;
}
```

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,427 @@
...
+                                            </TextField>}
+                                            {wsFolderItems?.length === 0 && <Button variant='contained'
+                                                className='buttonStyle'
+                                                style={{ backgroundColor: '#EE0000', textTransform: 'none', color: 'white' }}
+                                                onClick={() => this.selectFolder()}>
+                                                Select Folder
+                                            </Button>}
+                                        </div>
+                                        {compDescriptions && <>
+                                            <InputLabel required htmlFor='bootstrap-input'
```
datho7561, line 310: `htmlFor='groupedComponents'`

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,427 @@
...
+                                                    <li key={params.key}>
+                                                        <this.GroupHeader>{params.group}</this.GroupHeader>
+                                                        <this.GroupItems>{params.children}</this.GroupItems>
+                                                    </li>
+                                                )}
+                                                onChange={(e, v) => this.handleDevFileChange(e, v)} />
+                                        </>}
+                                        {selectedComponentDesc && selectedComponentDesc.starterProjects?.length > 0 &&
+                                            <div style={{ marginTop: '10px', marginBottom: '10px' }}>
+                                                <InputLabel htmlFor='bootstrap-input'
```
datho7561, line 360: `htmlFor='grouped-starterProjects'`

src/webview/create-component/createComponentLoader.ts
```diff
@@ -0,0 +1,156 @@
...
+    static refresh(): void {
+        if (panel) {
+            panel?.webview.postMessage({ action: 'loadingComponents' });
+        }
+    }
+}
+
+function validateName(event: any) {
+    let validationMessage = event.action !== 'validateAppName' ? OpenShiftItem.emptyName(`Required ${event.name}`, event.name.trim()) : null;
+    if (!validationMessage) validationMessage = OpenShiftItem.validateMatches(`Not a valid ${event.name}.
```
datho7561, line 130: `event.name` refers to the text in the input box. I think it would make more sense if you wrote something like

`${event.name} is not a valid ${validateAppName ? 'app' : 'component'} name.`

datho7561 (changes requested):
Alright. Here are some things I found while looking through the code

msivasubramaniaan:
> > If subfolder then what would be the name of that
> 
> The name of the subfolder should be the component name
> 
> > I don't see any empty of starter-project when only one starter project.
> 
> Maybe this only happens with light theme, let me double check...
> 
> > If it empty and it default taken as `component-name-app`
> 
> Okay, I understand this, but we don't display the app name anywhere or ask for the app name anywhere else, so I don't think it's worth asking for it. If we use a default app name, I think it makes sense to go with the one `odo` is using, which is `app`.

Removed application name field

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,420 @@
...
+                            <div className='title'>
+                                <Typography variant='h5'>Create Component</Typography>
+                            </div>
+                            <div className='subTitle'>
+                                <Typography>Create a component in the local workspace using the selected starter project and associated `devfile.yaml` configuration. Devfile is a manifest file that contains information about various resources (URL, Storage, Services, etc.) that correspond to your component.</Typography>
+                            </div>
+                            <div className='formContainer'>
+                                <div className='form'>
+                                    <div style={{ marginTop: '10px', marginBottom: '10px' }}>
+                                        <InputLabel required htmlFor='bootstrap-input'
```
datho7561, line 219: `htmlFor=component-textField`.

Once you've set this, you should be able to click on the title, and it will put the cursor in the text box

src/webview/create-component/app/component.tsx
```diff
@@ -0,0 +1,420 @@
...
+                                                    backgroundColor: 'var(--vscode-settings-textInputBackground)',
+                                                    WebkitTextFillColor: themeKind <= 1 ? 'black' : 'white',
+                                                    '&:disabled': {
+                                                        WebkitTextFillColor: themeKind <= 1 ? 'black' : 'white'
+                                                    }
+                                                }
+                                            }}
+                                            style={{ width: '30%', paddingTop: '10px' }}
+                                            helperText={component.helpText} />
+                                        <InputLabel required htmlFor='bootstrap-input'
```
datho7561, line 243: `htmlFor=grouped-folder`

datho7561 (changes requested):
Okay, a few more things I noticed

mohitsuman:
We are going to work on new workflow, hence closing this PR.

PR #2857, opened 2023-04-25
[WIP] fix create service webview

datho7561:
Fixes:
- hide the commands from palette (they don't work from palette)
- replace `odo application list` with a version that uses `oc`
- mui5 -ify the create service form
- Add a Patternfly-like dropdown for nested objects, it's collapsed by default.

Still broken:
- When I submit, it adds all the booleans (check boxes) into the formData, even the ones that the user hasn't interacted with. This results in invalid JSON for the custom resource, so the call to create the service fails.
- form could use a few more UI clean ups

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "vscode-openshift-connector",
-	"version": "1.2.0",
+	"version": "1.3.0",
 	"lockfileVersion": 2,
 	"requires": true,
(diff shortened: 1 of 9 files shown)


datho7561:
Broken part might be related to https://github.com/rjsf-team/react-jsonschema-form/issues/675, still need to investigate further

datho7561:
Replaced by #3435

Issue #2841, opened 2023-04-19
Old Service-related commands don't work


datho7561:
The following commands result in an error pop up when run from the command palette:
- "OpenShift: Create Service ..." (`clusters.openshift.csv.create`)
- "OpenShift: New Service" (`openshift.service.create` i think?)
- "OpenShift: Delete Service" (`openshift.service.delete.palette`)

Some of these commands rely on `odo application list`, which no longer exists.

We should either:
- Remove them (it seems like https://github.com/redhat-developer/vscode-openshift-tools/pull/2803 is supposed to replace the functionality of these commands)
- Find a replacement for `odo application list` (I haven't been able to after some searching, if there is one it's not provided by `odo`: https://odo.dev/docs/user-guides/v3-migration-guide/#commands-added-modified-or-removed-in-v3)
- Hide them from the command palette (I think this is a bad option, but it's better than leaving them accessible)

datho7561:
Okay, so after a bit more research, I think they key is hardcoding the application to `'app'` instead of listing all the apps and getting the user to pick an app.

datho7561:
With a bit of hacking, I can get the "Create Service" UI to appear, but it's not very functional:
- The location you open the UI from is not intuitive (I expected it to involve right clicking the cluster or the project, but it's not in that menu)
- The form is very long, with subobjects of subobjects that aren't hidden in any way, so you have to scroll past many unrequired fields to reach the "Create Service" button
- The form incorrectly marks some non-required fields as required
- The form doesn't use MUIv5, it instead uses a bunch of `.scss`

I have a fix for the above that I'll make a PR for, but now the issue I have is that optional boolean parameters in optional objects are assigned a value by the form, even if the user doesn't interact with the corresponding checkbox. So, it will create the optional object no matter what, but only with the boolean fields. This invalidates the form data, because some optional objects have required non-boolean fields. This means that when we try to create the service, we get an error message saying that field values are missing, and the service can't be created.

datho7561:
An update on this issue: most of the old broken commands have been removed. The "Create Service" webview is semi-functional. It now uses the same version of react-jsonschema-form as the rest of the extension. However, it would be nice to use the material ui theme for react-jsonschema-form. I'll create a separate issue for that.

datho7561:
Closed by #2862

Issue #2799, opened 2023-03-18
Integration tests to verify cli tool commands used in extension works with new version released


dgolovin:
To avoid compatibility issues between odo releases there should be integration tests to run:
* with latest nightly oc/odo build to detect problems early
* with latest official oc/odo release to verify compatibility before extension release

datho7561:
I'd be interested in working on this for next sprint. We have `npm urn test-integration`, which runs a bunch of `odo` and `oc` commands, but currently it's broken. I'd like to fix it. It might also involve cleaning up some broken methods in `src/odo.ts`

PR #2862, opened 2023-05-01
Remove the old service create/delete commands

datho7561:
The old commands no longer work. They have been replaced with a webview that provides a nice form to create a service (that also doesn't work but will take a lot of time to fix). Removing this code will result in less to test for #2799.

Part of #2841


datho7561:
@rgrunber if you have time, could you please review this PR?

datho7561:
one second, i located some more dead code I can remove in this PR

package.json
```diff
@@ -265,8 +265,6 @@
 		"onCommand:openshift.component.debug.palette",
 		"onCommand:openshift.resource.load",
 		"onCommand:openshift.resource.unInstall",
-		"onCommand:openshift.service.create",
-		"onCommand:openshift.service.delete",
```
rgrunber, line 269: I see `openshift.service.delete` is still present in odo.test.ts. Should that be removed as well ?

datho7561: Good catch! Yes, I will remove it in this PR. I'll also remove a similar call to `openshift.service.describe`. I am planning on rewriting odo.test.ts in my PR #2799.

rgrunber (approved):

PR #2863, opened 2023-05-01
Replace "delete component" command

### LINKED ISSUE #2825: Can't delete component in components view
Context menu of component in components view is missing delete option

https://user-images.githubusercontent.com/86789336/231115266-86bcd50b-5778-4539-9529-0d08ba6c03d5.mov

datho7561:
The command "delete component" currently doesn't work when run from the command palette and doesn't appear anywhere else.

In this PR, I have replaced it with two commands that you can run though right clicking on a component in the component view.

The first command is "Delete Configuration Files". This command runs `odo delete component --files -f`, which removes the `devfile.yaml` and the `.odo` folder from the user's computer, so that OpenShift Toolkit no longer recognizes the project as an OpenShift component. It doesn't delete any of the source code from the user's computer.

The second command is "Delete Source Folder". This command deletes the folder that the source code for a component is in.

Some of the other features that used to be handled by "delete" can be accomplished using other existing features:
- Stop Dev: if you are running the component in dev mode, `odo` will delete all the cluster resources associated with the component when you stop dev mode.
- Undeploy: if you have deployed the component to the cluster using `odo deploy` or the "Deploy" context menu, then you can use the "Undeploy" context menu to delete the cluster resources associated with the component.

Closes #2825

### DIFF
--- a/package.json
+++ b/package.json
@@ -259,7 +259,6 @@
 		"onCommand:openshift.component.log.palette",
 		"onCommand:openshift.component.followLog",
 		"onCommand:openshift.component.followLog.palette",
-		"onCommand:openshift.component.delete",
 		"onCommand:openshift.componentTypesView.registry.closeView",
 		"onCommand:openshift.component.debug",
 		"onCommand:openshift.component.debug.palette",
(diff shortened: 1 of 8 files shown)


datho7561:
@msivasubramaniaan sorry for working on this issue that you were assigned on, but it's important for #2799, since removing it means less things to test

@rgrunber if you have some time, do you think you could review this PR? Thanks!

datho7561:
@rgrunber Alright, I addressed the [feedback from Lukas](https://github.com/redhat-developer/vscode-openshift-tools/issues/2825#issuecomment-1531049750): I added two new commands to replace the functionality of the old one, and added some unit tests for them (that make heavy use of stubbing)

rgrunber (dismissed):
Change works well for me. I would just rebase and then we can merge.

datho7561:
Okay, I rebased, and it should be working now

PR #2865, opened 2023-05-01
Fix "OpenShift: Create" command

### LINKED ISSUE #2864: "OpenShift: Create" fails with "You need at least one project available"
How to reproduce the error:
1. Login to a cluster
2. Copy https://raw.githubusercontent.com/openshift-labs/starter-guides/ocp-4.6/mongodb-template.yaml into a `.yaml` file and save it
3. With the file opened in the editor and focused, run "OpenShift: Create" from the command palette
4. Even if you have a project set up, you get the error "You need at least one project available"

I think the root cause is that the underlying method that gets the name of the currently active project doesn't work.

datho7561:
Closes #2864

### DIFF
--- a/src/oc.ts
+++ b/src/oc.ts
@@ -4,13 +4,15 @@
  *-----------------------------------------------------------------------------------------------*/
 
 import { window } from 'vscode';
-import { CliChannel } from './cli';
+import { getInstance } from './odo';
+import { Command } from './odo/command';
+import { clusterRequired } from './openshift/openshiftItem';
 import { ToolsConfig } from './tools';
-import OpenShiftItem, { clusterRequired } from './openshift/openshiftItem';
 import { vsCommand } from './vscommand';
-import { OpenShiftProject } from './odo';
 
 export class Oc {
+
+    private static odo = getInstance();
 
     @vsCommand('openshift.create')
     @clusterRequired()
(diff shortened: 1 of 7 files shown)


datho7561:
I re-enabled the unit tests for this command; I still need to fix one of them though

datho7561:
@rgrunber This issue where all projects are detected as active should be fixed now. The solution was to use `odo list project` instead of `oc get project`

datho7561:
I think I should do a refactoring so that we always use this command when we want to list all projects on the cluster

datho7561:
@rgrunber okay, I finished the refactoring. The UI tests have been flaky for me

rgrunber:
Now when I call `OpenShift: Create`, it indicates the resource was successfully created. Checking the cli on the active project I see :

```
$ oc get templates
NAME                DESCRIPTION                                                                        PARAMETERS        OBJECTS
mongodb-ephemeral   MongoDB database service, without persistent storage. For more information ab...   8 (3 generated)   3
```

I think that fixes the issue. @datho7561 I'm not sure if we can do anything else here. According to the docs, once the template is added, it should be possible to call `oc new-app mongodb-ephemeral-template` (refering to the label of the template) but I don't see us using `new-app` anywhere.

PR #2868, opened 2023-05-04
Test case for component creation by importing from git.

### LINKED ISSUE #2779: Test case covering component creation by importing it from a git
Create a component using Component View's import from git button workflow.

rgrunber:
- Fixes #2779
- Add npm-watch as preLaunchTask to ensure test changes get updated
- Create a separate launch config for running the tests quickly with run-tests instead of setup-and-run

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -89,6 +89,28 @@
                 "--mocha_config",
                 "${workspaceFolder}/test/ui/.mocharc.js"
             ],
+            "preLaunchTask": "npm: watch",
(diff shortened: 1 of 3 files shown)


datho7561 (approved):
Works very well, and the shortcut for running the tests faster is really helpful.

I added some comments about using `function () {` instead of `() => {`. This is specifically something you need to be careful about when writing Mocha tests. Feel free to ignore my comments, since we don't do this consistently throughout the tests, and you seem to have found a work around for changing the timeout.

PR #2876, opened 2023-05-16
Increase timeouts for "Import from Git" UI test

datho7561:


datho7561:
How the heck could I possibly be increasing code coverage? That makes no sense.

datho7561:
I double checked, and the coverage is not actually increasing (if I get the coverage from running `main` locally, I get the same numbers as for this PR)

datho7561:
@rgrunber if you have time, could you please take a look at this PR?

rgrunber (approved):
Tested locally on cli and with the launch config in VS Code. Both seem to be passing.

PR #2881, opened 2023-05-19
Add 'Show welcome page on startup' checkbox

### LINKED ISSUE #2552: Add 'Show welcome page on startup' checkbox in OpenShift Welcome page
It should be placed somewhere on weclome page to be visible without scrolling, similar to VSCode welcome page
image

JessicaJHee:
Fixes #2552 

* Also fixes a distorted image in the welcome page

show-welcome-page-check

### DIFF
--- a/src/webview/welcome/app/welcomePage.tsx
+++ b/src/webview/welcome/app/welcomePage.tsx
@@ -4,7 +4,7 @@
  *-----------------------------------------------------------------------------------------------*/
 import React from 'react';
 import { VSCodeMessage } from './vsCodeMessage';
-import { Icon, Stack, Typography } from '@mui/material';
+import { Checkbox, FormControlLabel, Icon, Stack, Typography } from '@mui/material';
 import HelpIcon from '@mui/icons-material/Help';
 import ChatIcon from '@mui/icons-material/Chat';
 import BugReportIcon from '@mui/icons-material/BugReport';
@@ -19,13 +19,15 @@ export interface DefaultProps {
 }
 
 export class Welcome extends React.Component<DefaultProps, {
-    lastRelease: string
+    lastRelease: string,
+    isChecked: boolean
 }> {
 
     constructor(props) {
         super(props);
         this.state = {
-            lastRelease: ''
+            lastRelease: '',
+            isChecked: true
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/webview/welcome/app/welcomePage.tsx
```diff
@@ -216,8 +227,12 @@ export class Welcome extends React.Component<DefaultProps, {
     </footer>
 
     render(): React.ReactNode {
-        const { lastRelease } = this.state;
+        const { lastRelease, isChecked } = this.state;
         return <>
+            <div className='header__logo'>
+                <FormControlLabel control={<Checkbox checked={isChecked} onChange={this.updateShowWelcomePageConfig} sx={{ color: 'white', '&.Mui-checked': { color: 'white' }}} />}
```
datho7561, line 233: When you switch to the light theme, it's hard to see the checkbox. There are a few different ways to handle this:
- you can use the css variables set by VS Code that it uses (eg. setting the colours to `var(--vscode-foreground)` or `var(--vscode-background)`), which are updated automatically when the theme changes. It's probably best to do this, since that's what's done through the rest of the UI.
- you can use [the `<ThemeProvider>` element](https://github.com/redhat-developer/vscode-openshift-tools/pull/2803/files#diff-9979b203d268885e8c70cafa96829141b87e9e033290710c79bd0b44e8f88ad3R87). The idea is that you pass a new message from VS Code to the webview to indicate that the theme has changed. Then, you update a new state variable, `theme`, with the new theme, which you can build by passing the theme kind (light or dark) to `createTheme()`. This will update the colours for all the material ui components that are a child of the `ThemeProvider` component.

datho7561 (changes requested):
Looking good so far, one comment about the styling

mohitsuman:
@JessicaJHee I would move this checkbox to the end of the page below the Documentation section and would also like to capture telemetry of usage of the check/uncheck action. (can do telemetry one as a separate issue).

JessicaJHee:
> One more small thing I noticed; if you have the OpenShift welcome page set to not show on startup, but then open it later through the command (OpenShift: Welcome), then the checkbox is checked, even though the setting is set to not show the page.

Should be fixed now, thanks!

datho7561 (approved):
Looks good and works well for me! Thanks Jessica!

mohitsuman (approved):
LGTM

PR #2905, opened 2023-05-23
fix `npm run test` command that running unit tests without coverage

dgolovin:
It fails right now because redhat-account-extension installation
script is not called before executing the tests

Signed-off-by: Denis Golovin <dgolovin@redhat.com>


mohitsuman:
Thanks for the quick PR @dgolovin !!

PR #2908, opened 2023-05-24
Publish platform-specific VSIXs for OpenVSX.

rgrunber:
If you look at https://download.jboss.org/jbosstools/vscode/stable/vscode-openshift-tools/ , you'll see that the universal vsix we use for OpenVSX is ~259MB (because we include odo, helm, oc for 3 different platforms) while the platform specific bundle vsix are around ~110MB. Since ovsx is now much closer in support to vsce, we should publish the platform specific vsixs.

There's probably more cleanup we can do to eliminate the universal vsix, but for now I see it's used in the smoke test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,15 +86,14 @@ node('rhel8'){
     timeout(time:5, unit:'DAYS') {
       input message:'Approve deployment?', submitter: 'msuman,degolovi,msivasub'
     }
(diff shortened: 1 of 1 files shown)


mohitsuman (approved):
LGTM

msivasubramaniaan (approved):
LGTM

PR #2914, opened 2023-05-26
Switch from webpack to esbuild for bundling webviews

### LINKED ISSUE #2875: Investigate decreasing `npm run compile` times
Currently, running `npm run compile` takes a while (> 1 minute on my computer). This long build time even more noticeable when running the tests, since the script to run the tests always re-compiles/bundles the extension before running the tests. It would be nice to investigate ways to reduce this compile time to make it easier to develop the extension. 

One option worth investigating is [`esbuild`](https://esbuild.github.io/), which claims to be 100x faster than webpack

datho7561:
This PR uses esbuild instead of webpack to bundle the webviews. This dramatically reduces the time it takes to bundle the webviews, from >50 seconds to <1 second.

This PR also modifies the npm scripts and VS Code launch configurations to always recompile the webviews before launching.

I also set up eslint to be run before launching the extension. If the extension seems slower to launch in debug mode, this is why.

Closes #2875


datho7561:
I realized that type checking is not being performed on the webview files. I'm going to make a quick change to fix that.

datho7561:
Just ran into this: https://github.com/microsoft/vscode/issues/70283. Prelaunch tasks that use `dependsOn` with a background task cannot prevent debug from launching. This is a bit of a set back. I think my work around for this will be not running eslint as a part of the prelaunch tasks, and running the rest of the tasks synchronously every time launch is triggered.

rgrunber:
Tried it out and seems to be working well for me!

msivasubramaniaan (dismissed):
LGTM

Issue #2930, opened 2023-06-02
Decrease spacing in between paragraphs on welcome page


datho7561:
There is way too much spacing between these paragraphs on the welcome page:

![welcome-page-spacing-between-paragraphs](https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/0160b9f5-26fe-4dcd-86d8-e4fe17872d17)

PR #2934, opened 2023-06-05
changed height to rem from vh

msivasubramaniaan:

Fix: https://github.com/redhat-developer/vscode-openshift-tools/issues/2930

**Post Fix:**
image

### DIFF
--- a/src/webview/welcome/app/welcomePage.tsx
+++ b/src/webview/welcome/app/welcomePage.tsx
@@ -372,22 +372,22 @@ export class Welcome extends React.Component<DefaultProps, {
                         </div>
                         <section className='sticky-section-content'>
                             <div className='sticky-section-exten'>
-                                <p className='section__header-hint sticky-section-hint'>Allows developers to easily <p className='highlight'>create, deploy and live debug</p> container applications running on OpenShift &#38; Kubernetes. Thus enhancing the development inner loop workflow through One-click actions right from IDE.</p>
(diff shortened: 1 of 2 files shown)


src/webview/welcome/app/welcomePage.tsx
```diff
@@ -372,10 +372,10 @@ export class Welcome extends React.Component<DefaultProps, {
                         </div>
                         <section className='sticky-section-content'>
                             <div className='sticky-section-exten'>
-                                <p className='section__header-hint sticky-section-hint'>Allows developers to easily create, deploy and live debug container applications running on OpenShift &#38; Kubernetes. Thus enhancing the development inner loop workflow through One-click actions right from IDE.</p>
+                                <p className='section__header-hint sticky-section-hint'>Allows developers to easily <p className='highlight'>create, deploy and live debug</p> container applications running on OpenShift &#38; Kubernetes. Thus enhancing the development inner loop workflow through One-click actions right from IDE.</p>
```
datho7561, line 375: This 'highlight' class doesn't seem to change the text that's in the the `<p>` tag

datho7561:
It looks like the div containing the "OpenShift extension" title on the left is a fixed height (based on `vh`), so it scrolls past the paragraphs. Is there a way to set it to scale its height to the height of the paragraphs?

LeftTitleBoxIsFixedLength

msivasubramaniaan:
> It looks like the div containing the "OpenShift extension" title on the left is a fixed height (based on `vh`), so it scrolls past the paragraphs. Is there a way to set it to scale its height to the height of the paragraphs?
> 
> LeftTitleBoxIsFixedLength

May I know your screen-size and the resolution? I couldn't able to reproduce on my system

datho7561 (approved):
Looks good. I think there is a simpler way to accomplish this, but if you want to release soon, I think this can work for the time being

Issue #2939, opened 2023-06-05
Cloning a repo without any code fails with incorrect message


datho7561:
If you clone a repo that doesn't contain a `devfile.yaml` or any files that indicate the language or framework being used, then the form fails with the error message "Issue on Parsing Git URL/devfile".

This is a bad error message, since we know that the issue is not parsing the Git URL. Also, we don't need a devfile in the repo to import the project as a component. For instance, we can clone https://github.com/eclipse/lemminx, and it will automatically detect that a Java devfile is needed.

As well, in the cases that the recommended devfile can't be inferred from the repo content, we should be able to provide a selection drop down to pick the correct devfile (like in https://github.com/redhat-developer/vscode-openshift-tools/pull/2844).

Instead of the error message, we should have a warning message that explains that the language/framework couldn't be detected and what a devfile is, then provide a dropdown to select an appropriate devfile.

msivasubramaniaan:
@datho7561 can you please provide the git URL which showed the above error message.

datho7561:
https://github.com/datho7561/bug-free-meme

PR #2943, opened 2023-06-06
Show all the devfiles when no devfile.yaml in the git repo

msivasubramaniaan:

Fix: https://github.com/redhat-developer/vscode-openshift-tools/issues/2939

When no `devfile.yaml/language` specific code were detects, then the git import showed all the list of `devfiles` from the registry. User can pick any one. Please refer below:
no-devfile-demo

Handling the devfile with dropdown will be part of the https://github.com/redhat-developer/vscode-openshift-tools/issues/2697


src/webview/git-import/app/gitImport.tsx
```diff
@@ -469,7 +469,7 @@ export class GitImport extends React.Component<DefaultProps, {
                                     {!isDevFile &&
                                         <div className='strategyContainer strategyWarning'>
                                             <ErrorIcon color='warning' style={{ fontSize: 20 }} />
-                                            <Typography variant='body2' style={{ margin: '0 5px' }}>Here is the recommended devfile:</Typography>
+                                            <Typography variant='body2' style={{ margin: '0 5px' }}>Here {compDescription?.length > 1 ? 'are' : 'is' } the recommended devfile:</Typography>
```
datho7561, line 472: If there is more than one devfile shown, then the user needs to select a devfile. I think you should add the text "Please select a devfile that corresponds to the language/framework that you are using", or something similar, to indicate this to the user.

datho7561 (changes requested):

datho7561 (approved):
Looks good! Thanks, Muthu!

datho7561:
Unit tests fail (for me at least)

Issue #2965, opened 2023-06-13
Prevent users from creating same URL devfile registry


mohitsuman:
Users should be notified that the URL used to create a new devfile registry is already there. This should prevent any duplicate registry added to the view.

PR #2966, opened 2023-06-13
prevent the duplication of registry URL

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/2965


mohitsuman (approved):
lgtm

PR #2967, opened 2023-06-13
added watch script

msivasubramaniaan:
Added `watch` script to apply ***.ts** file changes(other than the webview folder) without close and open the extension.

**Note:** The changes in the webview folder never applied until run the `compile` script


datho7561 (approved):
I'll open an issue to investigate creating a background task that compiles everything

PR #2969, opened 2023-06-15
Use Describe webview on useWebviewInsteadOfTerminalView setting

### LINKED ISSUE #2910: "Describe Component" context menu always runs the command
To reproduce the issue:
1. Set `"openshiftToolkit.useWebviewInsteadOfTerminalView": true,` in `settings.json`
2. Create a component
3. Right click the component, select "Describe"

I expect the webview to appear, as alluded to in the description of the `openshiftToolkit.useWebviewInsteadOfTerminalView` setting, but instead, the command is run in the terminal.

JessicaJHee:
Fixes #2910

Peek 2023-06-15 15-44

### DIFF
--- a/src/odo/command.ts
+++ b/src/odo/command.ts
@@ -194,7 +194,7 @@ export class Command {
     }
 
     static describeComponent(): CommandText {
-        return new CommandText('odo describe component');
+        return new CommandText('odo', 'describe component');
     }
 
     static describeComponentJson(): CommandText {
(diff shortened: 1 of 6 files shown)


datho7561 (changes requested):
It would be nice if you could select and copy the text output of the command. Take a look if it's easy to do, but if not, don't worry about it.

Otherwise, looks great!

JessicaJHee:
@datho7561 should be working now (made the same change with the log webview), thanks for the review! 

Peek 2023-06-16 13-01

datho7561 (approved):
Looks great! Thanks, Jessica!

PR #2970, opened 2023-06-16
Remove unused libraries from package.json

datho7561:
Some of them are still installed, since they are transitive dependencies


mohitsuman:
@datho7561 If the extension workflow is working better after these unused libraries removed, let's merge this PR.

msivasubramaniaan (approved):
Tested n my local and the extension working fine as expected. LGTM

PR #2975, opened 2023-06-19
Replace @clusterRequired implementation

### LINKED ISSUE #2878: Replace `@clusterRequired`
It would be nice to fix the prompt to log in to a cluster when you perform an operation that requires a cluster but aren't logged into one. Currently, the prompt sometimes appears if you've never logged into a cluster on your computer (i.e. `.kube/config` is empty), but it doesn't count you as logged out if you've logged into a cluster and then logged out (i.e. you still have the cluster set as `current-context`).

We could use `oc whoami` to detect the current user and make sure it isn't `system:anonymous` (the username when not logged in). However, this will only work in OpenShift.

We could look into https://stackoverflow.com/questions/57004093/how-to-retrieve-current-user-granted-rbac-with-kubectl to see if that method is viable to detect if the user is logged on. eg. 'the user is considered logged on if they can create pods in the current namespace' or something similar.

JessicaJHee:
Fixes #2878

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -71,6 +71,7 @@ export interface Odo {
     removeRegistry(name: string): Promise<void>;
     describeComponent(contextPath: string, experimental?: boolean): Promise<ComponentDescription | undefined>;
     analyze(contextPath: string): Promise<AnalyzeResponse[]>;
+    canCreatePod(): Promise<boolean>;
 
     /**
      * Returns the URL of the API of the current active cluster,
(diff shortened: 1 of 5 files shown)


datho7561 (changes requested):
Looks great! Could you please add a test case for `canCreatePod` under `./test/integration/odo.test.ts`?

datho7561 (approved):
Looks good. Thanks, Jessica!

rgrunber:
With this change I see failures begining to occur as early as `npm run test` but without this change this were passing. May need to have a look here..

PR #2980, opened 2023-06-20
esbuild "watch mode"

### LINKED ISSUE #2968: `watch`-style `npm` script to auto compile in the background
Currently, as a part of #2967, we have `npm run watch`, which compiles the main extension code (but not the webviews) in the background.

It would be nice to also compile the webviews and run eslint in the background as a part of this script, and run this task as a prelaunch task in VS Code so that any errors prevent relaunching the extension.

This is not a trivial task, since VS Code doesn't allow running multiple background tasks as prelaunch tasks. Two options that I see are:
1. Make a TypeScript script that combines the different watch commands (`tsc`, `esbuild`, `eslint`) somehow.
2. Use esbuild plugins in order to run `tsc -noEmit` and `eslint` as a part of `esbuild watch`

datho7561:
Create a new script that automates running esbuild and TypeScript as a background task. This means that, while `npm run watch` is running, if there are any changes to the source files, then the source files will be recompiled automatically.

Fixes #2968


datho7561:
The command should be run automatically the first time you debug the extension (in place of the existing `npm run compile`), and will stay running through subsequent launches.

mohitsuman:
@datho7561 Does this cover the scenario that whenever we make a change in the webview tsx file, it automatically recompiles and update the webview ? 

Currently we have to manually run `npm run compile` to get the changes in place.

datho7561:
Based on how we load the HTML into the webview, if the webview is already open, it won't be updated. However, if you close and reopen the webview, it should update. I'll double check this use case.

mohitsuman:
So for this, instead of running `npm run compile`, we need to run `npm run watch` ?

datho7561:
`npm run watch` should be run automatically in the background when you start debugging the extension. The `npm run compile` script still exists, and is used in cases such as before running tests and before packaging the `.vsix`.

msivasubramaniaan (commented):
Tested on my local and LGTM. There was some delay noted each time compiling, will it be possible to make a separate call for the webview compile? If so we can do it on need basis

datho7561:
I enabled the "incremental" option for TypeScript compilation, which might speed up the build a bit.

msivasubramaniaan (approved):
LGTM

mohitsuman (approved):
lgtm

PR #2985, opened 2023-06-21
Improve Login to DevSandbox button workflow

### LINKED ISSUE #2977: Login to Devsandbox button workflow improvement
Screenshot 2023-06-20 at 12 40 26 AM

Currently if user clicks on the `Login to Sandbox` button without having the token copied, it throws error.

We should do a proper error handling on it:

- Detect if clipboard has token copied or not.
- If token present, active the vutton and keep it red color (sam color used for above learn more buttons)
- If no token copied, notify in tooltip that please `Get Token` and copy it and then click on `Login button`.

JessicaJHee:
Fixes #2977 

Peek 2023-06-21 17-05

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -500,6 +500,11 @@ export class Cluster extends OpenShiftItem {
         );
     }
 
+    static validateLoginToken(token: string): boolean {
+        const sha256Regex = new RegExp('^sha256~([A-Za-z0-9_]+)');
+        return sha256Regex.test(token);
+    }
+
     @vsCommand('openshift.explorer.login.clipboard')
     static async loginUsingClipboardToken(apiEndpointUrl: string, oauthRequestTokenUrl: string): Promise<string | null> {
         const clipboard = await Cluster.readFromClipboard();
--- a/src/webview/cluster/app/clusterView.scss
+++ b/src/webview/cluster/app/clusterView.scss
@@ -79,3 +79,27 @@
 .buttonSecondary:disabled {
     opacity: 0.4 !important;
 }
+
+.buttonRed {
+    white-space: nowrap !important;
+    display: inline-block !important;
+    margin-top: 8px !important;
+    margin-right: 8px !important;
+    background-color: #EE0000 !important;
+    color: var(--vscode-button-foreground) !important;
+}
+
+.buttonRed:hover {
+    color: var(--vscode-button-foreground) !important;
+    background-color: #BE0000 !important;
+    cursor: pointer !important;
+}
+
+.buttonRed:focus {
+    background-color: #BE0000 !important;
+    cursor: pointer !important;
+}
+
+.buttonRed:disabled {
+    opacity: 0.4 !important;
+}
--- a/src/webview/cluster/app/sandboxView.tsx
+++ b/src/webview/cluster/app/sandboxView.tsx
@@ -7,7 +7,7 @@ import { Button, CircularProgress, TextField } from '@mui/material';
 import IconButton from '@mui/material/IconButton';
 import Tooltip from '@mui/material/Tooltip';
 import Typography from '@mui/material/Typography';
-import { styled, ThemeProvider } from '@mui/styles';
+import { ThemeProvider, styled } from '@mui/styles';
 import * as React from 'react';
 import PhoneInput from 'react-phone-input-2';
 import 'react-phone-input-2/lib/style.css';
@@ -382,6 +382,9 @@ export default function addSandboxView(props): JSX.Element {
             postMessage('sandboxLoginUsingDataInClipboard', {apiEndpointUrl: currentState.apiEndpoint, oauthRequestTokenUrl: `${currentState.oauthTokenEndpoint}/request`});
         };
 
+        const invalidToken = currentState.errorCode === 'invalidToken';
+        const loginSandboxTitle = !invalidToken ? 'Login to DevSandbox OpenShift cluster with token from clipboard' : 'Token in clipboard is invalid. Select the Get Token option and copy to clipboard';
+
         return (
             <>
                 {( currentState.action === 'sandboxPageProvisioned' ) && (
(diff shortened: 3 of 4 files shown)


src/openshift/cluster.ts
```diff
@@ -500,6 +500,16 @@ export class Cluster extends OpenShiftItem {
         );
     }
 
+    static async validateLoginToken(userClusterUrl: string, userToken: string): Promise<boolean> {
+        try {
+            await Cluster.cli.executeTool(Command.odoLoginWithToken(userClusterUrl, userToken.trim()), undefined, true);
```
JessicaJHee, line 505: I'm not sure if there is a better way to check whether the token is valid or not

src/webview/cluster/app/sandboxView.tsx
```diff
@@ -413,8 +416,8 @@ export default function addSandboxView(props): JSX.Element {
                         <Tooltip title='Copy token from DevSandbox console page in browser' placement='bottom'>
                             <Button variant='contained' className='button' href={`${currentState.oauthTokenEndpoint}/request`}>Get token</Button>
                         </Tooltip>
-                        <Tooltip title='Login to DevSandbox OpenShift cluster with token from clipboard' placement='bottom'>
-                            <Button variant='contained' className='buttonSecondary' onClick={handleLoginButton}>Login to DevSandbox</Button>
+                        <Tooltip title={loginSandboxTitle} placement='bottom'>
+                            <span><Button variant='contained' disabled={invalidToken} className='button' onClick={handleLoginButton}>Login to DevSandbox</Button></span>
```
mohitsuman, line 420: I would recommend having the `Login to DevSandbox` as Red in color and active once there is token in the clipboard.

We can add a check that the button is only active, once the clipboard has entry in it. It might be possible that clipboard has some false data, so that will basically throw an error during login.

We also need to enable as SHA check that token copied is sha compatible, as the sandbox cluster token are SHA based.

src/webview/cluster/app/sandboxView.tsx
```diff
@@ -413,8 +416,8 @@ export default function addSandboxView(props): JSX.Element {
                         <Tooltip title='Copy token from DevSandbox console page in browser' placement='bottom'>
```
mohitsuman, line 416:
```suggestion
                        <Tooltip title='This opens the DevSandbox console page and user can copy the login token' placement='bottom'>
```

src/webview/cluster/clusterViewLoader.ts
```diff
@@ -185,6 +190,22 @@ async function clusterEditorMessageListener (event: any ): Promise<any> {
...
 
+async function pollClipboard(signupStatus) {
+    while (panel) {
+        const previousContent = await vscode.env.clipboard.readText();
+        await new Promise(r => setTimeout(r, 500));
+        const currentContent = await vscode.env.clipboard.readText();
+        if (previousContent && previousContent !== currentContent) {
+            const oauthInfo = await sandboxAPI.getOauthServerInfo(signupStatus.apiEndpoint);
+            let errCode = '';
+            if (!await Cluster.validateLoginToken(signupStatus.apiEndpoint, currentContent)){
```
mohitsuman, line 201: Maybe validate login token can have the token check if it is SHA compatible or not.

mohitsuman (changes requested):
Change around active status of the Login button.

src/webview/cluster/app/sandboxView.tsx
JessicaJHee: At the moment I activate the button only when the token is able to be used to login in without any error. I left a comment on my current approach. Another way to do it would be to just check the format of the clipboard with a regex, then throw an error in a vscode window if the token is invalid?

src/openshift/cluster.ts
datho7561: I think the regular expression check should be enough. I think that logging in and out of the cluster isn't necessary, and might slow down the extension. If the contents of the clipboard match the regex, but the token isn't valid, then we could show a popup when they hit the "Login to DevSandbox" button

src/webview/cluster/app/clusterView.scss
```diff
@@ -79,3 +79,27 @@
 .buttonSecondary:disabled {
     opacity: 0.4 !important;
 }
+
+.buttonRed {
```
datho7561, line 83: instead of modifying the button through the `.scss`, take a look at [`color='error'`](https://mui.com/material-ui/react-button/#color) and see if that's sufficient. That way you don't need to modify the button for all the different states.

src/webview/cluster/app/sandboxView.tsx
mohitsuman: I think doing both checks would make it more solid proof and error handling. The regex approach sounds good to me.

datho7561 (changes requested):

src/webview/cluster/app/sandboxView.tsx
JessicaJHee: Sounds good! The latest change adds both checks

src/webview/cluster/app/clusterView.scss
JessicaJHee: image

It's a bit hard to see when it's disabled, does this work?

datho7561: I think stick with what you had.

If we were managing themes using `<ThemeProvider>` in this webview, I think it could look better, but that would take a lot more refactoring work.

JessicaJHee:
Updated: 

Peek 2023-06-22 16-57

src/openshift/cluster.ts
```diff
@@ -500,6 +500,14 @@ export class Cluster extends OpenShiftItem {
         );
     }
 
+    static validateLoginToken(token: string): boolean {
+        const sha256Regex = new RegExp('^sha256~([A-Za-z0-9_]+)');
+        if (!sha256Regex.test(token)) {
```
datho7561, line 505: `return !sha265Regex.test(...`

src/webview/cluster/clusterViewLoader.ts
```diff
@@ -185,6 +190,22 @@ async function clusterEditorMessageListener (event: any ): Promise<any> {
     }
 }
 
+async function pollClipboard(signupStatus) {
+    while (panel) {
+        const previousContent = await vscode.env.clipboard.readText();
+        await new Promise(r => setTimeout(r, 500));
+        const currentContent = await vscode.env.clipboard.readText();
+        if (previousContent && previousContent !== currentContent) {
+            const oauthInfo = await sandboxAPI.getOauthServerInfo(signupStatus.apiEndpoint);
```
datho7561, line 199: Since sending and receiving the response from the auth server takes some time, the user could copy the token in this time. That means the change in content won't be detected.

I ran into this. As a result, I can't click the button unless I already have the token in my clipboard when I open the cluster view.

datho7561 (changes requested):

mohitsuman (dismissed):
lgtm

datho7561 (approved):
Other than the issue I mentioned with the tooltip, this looks good. Feel free to merge. Thanks, Jessica!

datho7561 (approved):

PR #2987, opened 2023-06-23
Add check for OpenShift cluster when opening developer console

### LINKED ISSUE #2628: Open in Developer console fails for kubernetes applications
The component running on the Kubernetes cluster has an option `Open in Developer Console`. This will fail as there is no Project in Kubernetes.

We need to remove the `Open in Developer Console` command for the component in Application Explorer if it is a k8s cluster.

Error:

1. Connected to k8s cluster
Screenshot 2022-11-05 at 4 24 35 AM

2. Error whe running the action of open in developer console
Screenshot 2022-11-05 at 4 17 50 AM

JessicaJHee:
Fixes #2628

### DIFF
--- a/package.json
+++ b/package.json
@@ -1211,7 +1211,7 @@
 				},
 				{
 					"command": "openshift.open.developerConsole",
-					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext",
+					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext && isOpenshiftCluster",
 					"group": "c0"
 				},
 				{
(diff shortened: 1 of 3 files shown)


JessicaJHee:
@mohitsuman I wasn't able to reproduce the error popup since I believe we no longer use `oc get project`. But I think it still makes sense to not show the option to open the developer console in that case.

mohitsuman:
> @mohitsuman I wasn't able to reproduce the error popup since I believe we no longer use `oc get project`. But I think it still makes sense to not show the option to open the developer console in that case.

Yes we should add a check and only show `Developer Console` for an OpenShift cluster.

msivasubramaniaan (approved):
LGTM

PR #2990, opened 2023-06-26
Throw error when login credentials are incorrect

### LINKED ISSUE #2979: Login with incorrect credentials gives login successful notification
When logging in with incorrect credentials, it should not show that login was successful and instead, notify that the password is incorrect and could not login.

Peek 2023-06-19 17-08

JessicaJHee:
Fixes #2979

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -430,7 +430,7 @@ export class Cluster extends OpenShiftItem {
         try {
             const result = await Progress.execFunctionWithProgress(
                 `Login to the cluster: ${clusterURL}`,
-                () => Cluster.cli.executeTool(Command.odoLoginWithUsernamePassword(clusterURL, username, passwd)));
+                () => Cluster.cli.executeTool(Command.odoLoginWithUsernamePassword(clusterURL, username, passwd), undefined, true));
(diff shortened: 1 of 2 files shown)


src/openshift/cluster.ts
```diff
@@ -430,7 +430,7 @@ export class Cluster extends OpenShiftItem {
         try {
             const result = await Progress.execFunctionWithProgress(
                 `Login to the cluster: ${clusterURL}`,
-                () => Cluster.cli.executeTool(Command.odoLoginWithUsernamePassword(clusterURL, username, passwd)));
+                () => Cluster.cli.executeTool(Command.odoLoginWithUsernamePassword(clusterURL, username, passwd), undefined, true));
```
datho7561, line 433: The error is thrown correctly, but the incorrect password is in the error popup. I think it would be nice to catch the exception, then rethrow with a modified error message that doesn't contain the password. This would be important in the case that someone is doing a screencast of using the extension and types in their password incorrectly, since in that case, it would expose what the user typed in for the password, which is probably close to the actual password.

datho7561 (approved):
Looks good. Thanks, Jessica!

Issue #2915, opened 2023-05-26
Make sure that GitHub URLs only contain first org and repo name in "Git Import"


datho7561:
I entered `https://github.com/redhat-developer/vscode-openshift-tools/issues/2909` as the git URL in the git import workflow, and it's stuck on cloning the repo:

![Screenshot from 2023-05-26 17-08-55](https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/f8a89be6-e9bc-4568-a21a-2ebe64a0bff7)

It lists the repo as valid. Either we should list the repo as invalid, or automatically strip the `/issues/2909` from the URL

Issue #2938, opened 2023-06-05
Trying to clone a private git repo by URL causes UI to freeze


datho7561:
To reproduce this issue:
1. Use SSH keys for authenticating with GitHub instead of username/password (this might be enforced by GitHub nowadays, I'm not sure though)
2. Create a private repo on GitHub
3. Enter the URL of the repo in the "Import from Git" dialog, despite the fact that you need to use `git@github.com/...` to clone repos when you use keys to authenticate with GitHub.
4. The "enter username for github.com" popup appears
5. Close it without entering anything
6. The UI is stuck at the importing from URL step. No progress is indicated.

See https://github.com/redhat-developer/vscode-openshift-tools/issues/2929#issuecomment-1574201423

datho7561:
I double checked, and this is fixed. Either as a part of https://github.com/redhat-developer/vscode-openshift-tools/pull/2991 or the create component redesign

PR #2991, opened 2023-06-27
trim the git URL till repo name

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/2915


src/webview/git-import/gitImportLoader.ts
```diff
@@ -19,7 +19,7 @@ import { selectWorkspaceFolder } from '../../util/workspace';
 import { vsCommand } from '../../vscommand';
 import { loadWebviewHtml } from '../common-ext/utils';
 import { DevfileConverter } from './devfileConverter';
-import GitUrlParse = require('git-url-parse');
+import { gitUrlParse } from './gitParse';
```
datho7561, line 22: If you really want to include the `git-url-parse` source code in our repo, we should remove it as a dependency from the `package.json`, and instead add its dependencies to our `package.json`. This will help us avoid including two copies of the file (one under `src` and one under `node_modules`)

datho7561 (changes requested):

src/webview/git-import/app/gitImport.tsx
```diff
@@ -217,7 +218,7 @@ export class GitImport extends React.Component<DefaultProps, {
                 clone: undefined
             },
             gitURL: {
-                value: value,
+                value: this.getTrimmedURL(value),
```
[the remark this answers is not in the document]
msivasubramaniaan: Trimming the URL will happen once user clicks the `analyze` button now.

src/webview/git-import/gitImportLoader.ts
```diff
@@ -273,7 +273,7 @@ function validateGitURL(event: any) {
         });
     } else {
         try {
-            const parse = GitUrlParse(event.param);
+            const parse = gitUrlParse(event.param);
```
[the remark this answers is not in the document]
msivasubramaniaan: It not only for hostname, Git URL parse will parse the repository URL and provides the detatils like name, repo_name, organization, commit details

src/webview/git-import/gitImportLoader.ts
msivasubramaniaan: Removed `git-url-parse`. But added `git-up` as dependency

src/webview/git-import/app/gitImport.tsx
```diff
@@ -236,6 +237,11 @@ export class GitImport extends React.Component<DefaultProps, {
         });
     }
 
+    getTrimmedURL = (value: string): string => {
+        const parsedURL = gitUrlParse(value);
+        return `${parsedURL.protocol}://${parsedURL.source}/${parsedURL.full_name}`;
```
datho7561, line 242: If I enter an SSH repo like `git@github.com:datho7561/lemminx.git`, then it gets modified to `ssh:////github.com/datho7561/lemminx`. If I click "Analyze" and select a folder to clone into, it gets stuck at "Cloning the repository" with no visible error message. This is because `ssh:////github.com/datho7561/lemminx` is not the same as `git@github.com:datho7561/lemminx.git`. (You will need to use one of your own repos to test this).

datho7561 (changes requested):

src/webview/git-import/gitImportLoader.ts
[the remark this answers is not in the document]
mohitsuman: @msivasubramaniaan we don't need org and commit details for our workflow. The important thing is to have a valid git url and strip out the information after repo name.

src/webview/git-import/app/gitImport.tsx
msivasubramaniaan: Showing error on accessing private repo, Will handle this case on https://github.com/redhat-developer/vscode-openshift-tools/issues/2938
image

src/webview/git-import/gitImportLoader.ts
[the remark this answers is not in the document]
msivasubramaniaan: @mohitsuman Yes we were not implemented that workflow yet. But We have an `Advanced Git Option` in UI, This references like **brange,tag or commit** will helps us by that time. Please create a separate ticket for implementing those options

src/webview/git-import/app/gitImport.tsx
```diff
@@ -236,6 +241,11 @@ export class GitImport extends React.Component<DefaultProps, {
         });
     }
 
+    getTrimmedURL = (value: string): string => {
```
datho7561, line 244: This function doesn't use the component's state, so you can move it out of the class

datho7561 (approved):
Looks good! One small comment, feel free to address it and merge or ignore it and merge. Thanks, Muthu!

Issue #2993, opened 2023-06-28
Implement Serverless Functions into OpenShift Toolkit extension


mohitsuman:
Implement the OpenShift Serverless view into OpenShift Toolkit extension.

- [x] Add a new view in the navigation with name as `Serverless Functions`
- [x] Rewrite the UI in material UI workflow in sync with vscode theme. Remove the old UI elements completely
- [x] Update to latest `func cli` https://github.com/knative/func

rgrunber:
@datho7561 do you see anything in https://github.com/redhat-developer/vscode-knative/ not playing well with esbuild ? Otherwise , size shouldn't be an issue.

datho7561:
Yes, `shelljs` doesn't play well with `esbuild`. We can probably use alternatives to `shelljs`, but that will require a bit more effort.

msivasubramaniaan:
Yes it has. 

But we are going to add one more [CLI](https://mirror.openshift.com/pub/openshift-v4/clients/serverless/latest/) on OpenShift and a Webview on the OpenShift extension. I hope the size shouldn't be a problem.

mohitsuman:
> Does knative have any webviews?

It has webviews, but onty few and the idea is to add more webviews in the future. For dependencies that do not work well, we can update them. Given the repo has not been touched in a while, I would like that when we implement it, we add dependencies as per our toolkit workflow.

PR #3003, opened 2023-07-03
Serverless create, build and run functions in OpenShift

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/2993

Currently This PR covers the Create, Build and Run functionalities of [func cli](https://github.com/knative/func) and the Deploy, invoke function will be covered by https://github.com/redhat-developer/vscode-openshift-tools/issues/3010


datho7561:
If I close the webview without pressing any buttons/entering any information, it still adds a folder to my workspace

datho7561:
I tried creating a new serverless function. I input the name, language, template, and folder, then I clicked the "Create" button. It got stuck on a loading page, and nothing has been added to the folder I specified.

datho7561:
I figured out the actual error: I didn't set the permissions correctly to push the image to quay.io. After I created the repo and added the necessary permissions, it worked as expected and moved on to "Run".

There should be some sort of error message if pushing the built container image to the registry fails.

datho7561:
The "Build" step doesn't seem to push the container image to the container image repository (in my case quay.io/davthomp/my-knative-function)

datho7561:
I can't seem to open the UI for an existing KNative function.

datho7561:
I'm deleting my previous comments to make my feedback more concise, and since some of the things I said were because I didn't know what I was doing.

After playing around with the feature a bit more, I figured out the following things, which helped me get through the "Create", "Build", and "Run" steps of the UI:
- You need to be signed into a image registry, such as quay.io or dockerhub, in order for this UI to work
- You need to set up `docker-credential-pass` with your image registry username and password, since `kn`/`func` cannot use the fact that you are logged in with `docker`/`podman` to access the registry
- The image is not pushed to the image registry during the build step, it's built locally and tagged with an appropriate name
- The image is not pulled from the specified image registry during the "Run", it uses the local build of the image. By default, the image is run using `podman`/`docker` and is accessible through `localhost`.

Things we can improve:
- We need to provide an error popup if the build fails. There was no indication that the build failed and I was confused why the wizard wouldn't advance to step 3 after the build finished.
- If I load a project folder for an existing serverless function, it won't appear in the tree view. It should be fairly simple to detect if a folder is a serverless function folder, since it has a `func.yaml` file at the root of the workspace
- If I create a new serverless function, then the source folder isn't added to the current VS Code workspace until I close the wizard. I think it should be added right after you click "Create" in the first step of the wizard.
- If I close the wizard before I finish creating a new serverless function, then it attempts to add the folder `/new` to the workspace. We shouldn't add any folders to the workspace if the user didn't finish filling out the first step of the wizard.
- After the serverless function is set up, it should be possible to build and run the function without opening the Webview. i.e. the context menu items "Build" and "Run" should perform the operation without opening the Webview.

msivasubramaniaan:
> I'm deleting my previous comments to make my feedback more concise, and since some of the things I said were because I didn't know what I was doing.
> 
> After playing around with the feature a bit more, I figured out the following things, which helped me get through the "Create", "Build", and "Run" steps of the UI:
> 
> * You need to be signed into a image registry, such as quay.io or dockerhub, in order for this UI to work
> * You need to set up `docker-credential-pass` with your image registry username and password, since `kn`/`func` cannot use the fact that you are logged in with `docker`/`podman` to access the registry
> * The image is not pushed to the image registry during the build step, it's built locally and tagged with an appropriate name
> * The image is not pulled from the specified image registry during the "Run", it uses the local build of the image. By default, the image is run using `podman`/`docker` and is accessible through `localhost`.
> 
> Things we can improve:
> 
> * We need to provide an error popup if the build fails. There was no indication that the build failed and I was confused why the wizard wouldn't advance to step 3 after the build finished. - **Addressed and I will push the code**
> * If I load a project folder for an existing serverless function, it won't appear in the tree view. It should be fairly simple to detect if a folder is a serverless function folder, since it has a `func.yaml` file at the root of the workspace - **Tested in my local and it is working fine on both ways 1) Open Folder 2) Add to workspace. Let me know how you tested?**
> * If I create a new serverless function, then the source folder isn't added to the current VS Code workspace until I close the wizard. I think it should be added right after you click "Create" in the first step of the wizard. - **If we add after create then the extension was restarts, we lost the WebView control, So provided "Finish" option on each level, if user wants to finish then it will add it to workspace, or if user closed it will does.**
> * If I close the wizard before I finish creating a new serverless function, then it attempts to add the folder `/new` to the workspace. We shouldn't add any folders to the workspace if the user didn't finish filling out the first step of the wizard. - **Will work on this**
> * After the serverless function is set up, it should be possible to build and run the function without opening the Webview. i.e. the context menu items "Build" and "Run" should perform the operation without opening the Webview. - **If so we need to use vscode open dialog for getting build name and other required details, To overcome the vscode dialogs, open it on webview. So the flow was uniform across**

datho7561:
> Tested in my local and it is working fine on both ways 1) Open Folder 2) Add to workspace. Let me know how you tested?

I can't reproduce this issue, my bad

> If so we need to use vscode open dialog for getting build name and other required details, To overcome the vscode dialogs, open it on webview. So the flow was uniform across

You can run `func build` and `func run` on the command line without passing these parameters. I think that you only need to specify the parameters the first time that you run the Knative function.

datho7561:
> Regarding the error dialog

"Build" doesn't seem to be working any more, there is no output in the terminal. It works from the command line (`func build`) but not in the UI.

> and empty workspace added I will push the code. Please have a look

Okay, the issue where it adds `/new` to the workspace if I close the wizard before filling it out is gone.

> It is not only on the first time, user can update the build image if needed while build

I don't think that changing the name of the container image is that common, since if other people are pulling the image from a container registry, they expect the name of the container image to remain the same. I think that `func` keeps track of the name of the built image under the `image` field in the `func.yaml` file. What is your opinion on setting this value in the first step of the wizard, then providing an "Edit" wizard where the user can edit these values?

> Also there was an option run,

What do think about having a context menu item "Build and Run"? So we could have these context items:
- Build
- Build and Run
- Run without Rebuilding

msivasubramaniaan:
> > Regarding the error dialog
> 
> "Build" doesn't seem to be working any more, there is no output in the terminal. It works from the command line (`func build`) but not in the UI.
> 
> > and empty workspace added I will push the code. Please have a look
> 
> Okay, the issue where it adds `/new` to the workspace if I close the wizard before filling it out is gone.
> 
> > It is not only on the first time, user can update the build image if needed while build
> 
> I don't think that changing the name of the container image is that common, since if other people are pulling the image from a container registry, they expect the name of the container image to remain the same. I think that `func` keeps track of the name of the built image under the `image` field in the `func.yaml` file. What is your opinion on setting this value in the first step of the wizard, then providing an "Edit" wizard where the user can edit these values?
> 
> > Also there was an option run,
> 
> What do think about having a context menu item "Build and Run"? So we could have these context items:
> 
> * Build
> * Build and Run
> * Run without Rebuilding

@datho7561 Moved to context menu. Please have a look

src/serveressFunction/build-run-deploy.ts
```diff
@@ -0,0 +1,219 @@
...
+            await this.buildProcess(functionUri, functionName);
+        }
+    }
+
+    private async buildProcess(functionUri: Uri, functionName: string) {
+        const clusterVersion: ClusterVersion | null = await this.checkOpenShiftCluster();
+        const buildImage = await this.getImage(functionUri);
+        const outputEmitter = new EventEmitter<string>();
+        let devProcess: ChildProcess;
+        let terminal = window.createTerminal({
```
datho7561, line 92: See `./src/webview/openshift-terminal/openShiftTerminal.ts`, `OpenShiftTerminalManager.createTerminal()` in order to integrate into the separate OpenShift Terminal. If there is something missing in the API, please let me know so that I can fix it.

package.json
```diff
@@ -803,6 +794,43 @@
...
+			},
+			{
+				"command": "openshift.Serverless.run",
+				"title": "Run",
+				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.Serverless.stopRun",
+				"title": "Stop",
+				"category": "OpenShift"
```
datho7561, line 833: Make sure to hide these from the command palette

datho7561 (changes requested):
Seems to be working pretty well! I don't have the issue with `build` any more.

The context menu for running the function without building appears even when the function hasn't been built yet. I don't think this is that bad, since the command immediately produces an error and explains what went wrong. Also, we would need to run docker/podman commands in order to figure out if the function has been built, and we don't have an easy way to do that.

src/serveressFunction/build-run-deploy.ts
msivasubramaniaan: The integration currently blocked by https://github.com/redhat-developer/vscode-openshift-tools/issues/3031

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,332 @@
...
+                                }
+                            }
+                        }}
+                        helperText={functionData.helpText}
+                    />
+                </FormControl>
+                <FormControl sx={{ margin: '2rem 0 0 2rem', width: 'auto' }}>
+                    <Autocomplete
+                        value={imageData.name}
+                        disabled={functionData.error || functionData.name.length === 0}
```
mohitsuman, line 191: Why do we need this check ? A user can first opt to provide the build image and then provide the name of function. Both actions are mutually exclusive. We should remove the check of function name length.

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,332 @@
...
+                        onChange={(_event, newValue: string) => {
+                            if (newValue) {
+                                const value = newValue.replace('Add', '').trim();
+                                if (!imageRegex.test(value)) {
+                                    this.setState({
+                                        imageData: {
+                                            name: '',
+                                            error: true,
+                                            helpText: 'Provide full image name in the form [registry]/[namespace]/[name]:[tag]'
```
mohitsuman, line 200: Where are we using this helptext. I see the drop down automatically fetched quay and docker.io image names, which can be completely incorrect.

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,332 @@
...
+                            }
+                        }}
+                        filterOptions={(options, params) => {
+                            const filtered = filter(options, params);
+
+                            const { inputValue } = params;
+                            // Suggest the creation of a new value
+                            const isExisting = options.some((option) => inputValue === option);
+                            if (inputValue !== '' && !isExisting) {
+                                filtered.push(`Add ${inputValue}`);
```
mohitsuman, line 224: If user can `Add` a new image name, there should be a provision to remove that added name also. Right now we can only add, but cannot remove that.

If possible, add a delete icon at the end of the select option in the dropdown for the specific added build image.

mohitsuman: Screenshot 2023-07-19 at 2 36 50 AM

For example, delete icon at the right of the last row (as that was added)

mohitsuman (changes requested):
Screenshot 2023-07-19 at 2 40 51 AM

I would like the Build Image input box to be same length as fn name. And similarly increase the sizes of the input box for Language and Template so that Language, Template and Select Folder together fall in the same length.

mohitsuman:
We need to Disable the Template Drop down till the Language is not selected. Once Language is selected, the Template will be in Active state. This will help user to understand the workflow.

mohitsuman:
Increase the height of the input box by a small height.

Screenshot 2023-07-19 at 2 44 51 AM

src/webview/serverless-function/app/createFunction.tsx
msivasubramaniaan: When user entered invalid image name then help text will appear
help-text

src/webview/serverless-function/app/createFunction.tsx
msivasubramaniaan: The delete icon was not possible on the `options`

src/serveressFunction/build-run-deploy.ts
datho7561: Closing this comment. See https://github.com/redhat-developer/vscode-openshift-tools/pull/3041, which reverts the OpenShift Terminal

datho7561 (dismissed):
Looks good to me. Thanks, Muthu!

package.json
```diff
@@ -803,6 +794,48 @@
 				"command": "openshift.sandbox.signup",
 				"title": "Provision OpenShift Developer Sandbox",
 				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.Serverless.openFunction",
+				"title": "Open Function",
```
mohitsuman, line 800: @mkuznyetsov What does Open Function command do ?

src/tools.json
```diff
@@ -59,6 +59,43 @@
             }
         }
     },
+    "func": {
+        "description": "Function CLI tool",
+        "vendor": "Red Hat, Inc.",
+        "name": "func",
+        "version": "1.10.0",
```
mohitsuman, line 66: @msivasubramaniaan Bump the version to latest func 1.10.1 https://github.com/knative/func/releases/tag/knative-v1.10.1

src/util/workspace.ts
```diff
@@ -71,14 +80,19 @@ export async function selectWorkspaceFolder(skipWindowPick = false): Promise<Uri
...
         });
         if (!selectedFolders) return null;
         if (fs.existsSync(path.join(selectedFolders[0].fsPath, '.odo', 'config.yaml'))) {
             void window.showInformationMessage(
                 'The selected folder already contains a component. Please select a different folder.',
             );
             return selectWorkspaceFolder(skipWindowPick);
+        } else if(folderName && fs.existsSync(path.join(selectedFolders[0].fsPath, folderName))) {
+            void window.showInformationMessage(
+                `The folder ${folderName} already exists,  Please select a different folder.`,
```
mohitsuman, line 93:
```suggestion
                `The folder ${folderName} already exists. Please select a different folder.`,
```

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,327 @@
...
+        const filter = createFilterOptions<string>();
+        const imageRegex = RegExp('[^/]+\\.[^/.]+\\/([^/.]+)(?:\\/[\\w\\s._-]*([\\w\\s._-]))*(?::[a-z0-9\\.-]+)?$');
+        return (
+            <>
+                <FormControl sx={{ margin: '2rem 0 0 2rem', flexFlow: 'row', width: 650, gap: 0.2 }}>
+                    <FormControl sx={{ width: 180 }}>
+                        <Button variant='contained'
+                            disabled={true}
+                            className='labelStyle'>
+                            Function Name*
```
mohitsuman, line 170: Let's keep the star with space between the text. And keep it consistent across Labels.

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,327 @@
...
+                        }}
+                        id='image-dropdown'
+                        options={images}
+                        renderOption={(props, option) => <li {...props}>{option}</li>}
+                        clearOnBlur
+                        disableClearable
+                        fullWidth
+                        renderInput={(params) => (
+                            <TextField {...params}
+                                placeholder='Provide image name (podman, docker, quay)' error={imageData.error} helperText={imageData.helpText} />
```
mohitsuman, line 246:
```suggestion
                                placeholder='Provide full image name (podman, docker, quay)' error={imageData.error} helperText={imageData.helpText} />
```

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,327 @@
...
+                    <Autocomplete
+                        value={language}
+                        id='language-dropdown'
+                        options={languages}
+                        onChange={(e, v) => this.handleDropDownChange(e, v, true)}
+                        renderOption={(props, option) => <li {...props}>{option}</li>}
+                        fullWidth
+                        disableClearable
+                        renderInput={(params) => (
+                            <TextField {...params} placeholder='Select the language' />
```
mohitsuman, line 267:
```suggestion
                            <TextField {...params} placeholder='Select the Language Runtime' />
```

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,327 @@
...
+                        value={template}
+                        id='template-dropdown'
+                        options={templates}
+                        disabled={language?.length === 0}
+                        onChange={(e, v) => this.handleDropDownChange(e, v)}
+                        renderOption={(props, option) => <li {...props}>{option}</li>}
+                        fullWidth
+                        disableClearable
+                        renderInput={(params) => (
+                            <TextField {...params} placeholder='Select the template based on the language selected' />
```
mohitsuman, line 290:
```suggestion
                            <TextField {...params} placeholder='Select the Function template' />
```

src/webview/serverless-function/app/createFunction.tsx
```diff
@@ -0,0 +1,327 @@
...
+                    <Autocomplete
+                        value={wsFolderPath ? wsFolderPath.fsPath : ''}
+                        id='folder-dropdown'
+                        options={folders}
+                        onChange={(e, v) => this.handleWsFolderDropDownChange(e, v)}
+                        renderOption={(props, option) => <li {...props}>{typeof option === 'string' ? option : option.fsPath}</li>}
+                        fullWidth
+                        disableClearable
+                        renderInput={(params) => (
+                            <TextField {...params} placeholder='Provide the folder to add the function files' />
```
mohitsuman, line 311:
```suggestion
                            <TextField {...params} placeholder='Select the folder to initialise the function at that path' />
```

src/webview/serverless-function/app/home.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    render(): React.ReactNode {
+        return (
+            <div className='mainContainer margin'>
+                <div className='title'>
+                    <Typography variant='h5'>Serverless Function</Typography>
+                </div>
+                <div className='subTitle'>
+                    <Typography>The OpenShift Serverless Functions support enables developers to create, build, run, invoke and deploy serverless functions on OpenShift, providing a seamless development experience with the latest kn and func CLI tool integrated.</Typography>
```
mohitsuman, line 41:
```suggestion
                    <Typography>The OpenShift Serverless Functions enables developers to easily create, build, run, invoke and deploy [stateless](https://www.techtarget.com/whatis/definition/stateless-app), [event-driven](https://en.wikipedia.org/wiki/Event-driven_programming) functions as Knative Services through the guided workflow via the extension</Typography>
```

mohitsuman (changes requested):
Reviewed the wordings and UX changes.

package.json
msivasubramaniaan: It opens the `func.yaml` in editor

datho7561:
- Clicking the "Language" option in the drop down causes the page to go blank, so I can't create a new function
- In my opinion, the UI elements, in particular the text field labels, no longer look like Material (see [material design text field](https://m2.material.io/components/text-fields))

datho7561:
oh

msivasubramaniaan:
> * Clicking the "Language" option in the drop down causes the page to go blank, so I can't create a new function
> * In my opinion, the UI elements, in particular the text field labels, no longer look like Material (see [material design text field](https://m2.material.io/components/text-fields))

https://github.com/redhat-developer/vscode-openshift-tools/pull/3049 will fix the mentioned issue

PR #3005, opened 2023-07-05
Fix cluster-dependent UI test suite

### LINKED ISSUE #2777: Test case covering logging into existing cluster
We would like to cover basic workflow when user is logging into a cluster.

datho7561:
With a running `crc` cluster, run `npm run cluster-ui-test` in order to run the cluster-dependent UI test suite.

- renamed `smoke-test` to `cluster-ui-test` and removed some of the tests that overlap between the suites
- Most of this work is also done in #2803, so I just copied it from over there
- Temporarily addresses #2780, we will need to rewrite it once the new create component workflow is in place

Fixes #2777


lgrossma (approved):
LGTM

msivasubramaniaan (approved):
LGTM

PR #3008, opened 2023-07-06
Fix integration tests on internal Jenkins

### LINKED ISSUE #3006: Make integration test suite more robust so that it can work on Jenkins
I tried to run the integration tests on our internal Jenkins, and they failed. It would be nice to address these failures:
- The cluster API URL might have a `/` added or removed from the end in `odo`/`oc` commands that return the URL, which causes those tests to fail
- The user `kubeadmin` gets changed to `kube:admin` when listing the current user using `odo`/`oc`
- The service binding operator is not installed on the cluster I was using. It would be nice to install it as a part of the suite, perhaps in the Jenkinsfile
- I don't know why yet, but the `loadItems` test timed out.

datho7561:
- load the kube config after the first login to the cluster

Fixes #3006

### DIFF
--- a/test/integration/command.test.ts
+++ b/test/integration/command.test.ts
@@ -20,8 +20,6 @@ import { Command } from '../../src/odo/command';
 const ODO = getInstance();
 const kc = new KubeConfig();
 
-kc.loadFromDefault();
-
 const newProjectName = `project${Math.round(Math.random() * 1000)}`;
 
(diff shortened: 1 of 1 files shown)


datho7561:
To test this PR, delete your `~/.kube/config` (*make sure nothing important is in it first*) before running the integration test suite.

PR #3011, opened 2023-07-10
Add create component functionality

JessicaJHee:
Peek 2023-07-10 12-30

### DIFF
--- a/src/webview/common/componentNameInput.tsx
+++ b/src/webview/common/componentNameInput.tsx
@@ -4,11 +4,11 @@
  *-----------------------------------------------------------------------------------------------*/
 import { TextField } from '@mui/material';
 import * as React from 'react';
-import { ComponentNameState } from '../create-component/pages/fromLocalCodebase';
 
 export type ComponentNameInputProps = {
-    componentName: ComponentNameState;
-    setComponentName: React.Dispatch<React.SetStateAction<ComponentNameState>>;
+    isComponentNameFieldValid: boolean,
+    componentNameErrorMessage: string,
+    setComponentName: React.Dispatch<React.SetStateAction<string>>
 };
 
 export function ComponentNameInput(props: ComponentNameInputProps) {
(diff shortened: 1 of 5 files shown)


JessicaJHee:
https://github.com/redhat-developer/vscode-openshift-tools/assets/73968480/4e75cd00-809b-49b4-8976-d2a3d8f94c75

A few more things that are missing:

- [ ]  implement select different devfile option

- [ ]  implement advanced git option
- [ ]  add loading icon when creating component?

datho7561:
A few things:
- while cloning the repository, it would be nice to disable the "Back" and "Next" buttons
- for the "import form git" workflow, it would be nice to specify in the "pick folder" dialog whether the source code will be added to the folder specified or a new subfolder

datho7561:
Some other things:
- some of the text inputs do not have spacing between them. This might be because of a change I did. [Check out the `spacing` attribute of the `<Stack>` component](https://mui.com/material-ui/react-stack/#direction)

datho7561:
Looking great! Two things:
- For the "Import from git" workflow, the helper text mentions the folder the project will be clone into, but the repo content is actually cloned into the parent folder (i.e. the folder you specify in the file picker)
- For the "From existing local codebase" workflow, after you have scanned the devfile, the screen looks like this:
component-and-folder-should-be-disabled
  I can't tell if you are already doing this, but is there some sort of way to visually indicate that you can't modify the component name and folder until you click back? I know `<Button>` has the `disabled` property, maybe the input fields have something similar
- Also for the "From existing local codebase", when you select a folder that isn't in the list, the contents of the input field doesn't change. Is it possible to put the path to the selected folder there?

datho7561:
> But I can also disable that as well is you think that makes more sense?

Okay, I think it's fine to leave it enabled

datho7561 (approved):
Looks great! Thanks, Jessica!

PR #3014, opened 2023-07-11
Implement "from template project" workflow

datho7561:
When you use the "Create Component" command, the "From Template Project" workflow should be fully functional and allow you to create a new component from a template project that's defined in a devfile from one of the devfile registries that are configured.

### DIFF
--- a/src/odo.ts
+++ b/src/odo.ts
@@ -95,6 +95,17 @@ export interface Odo {
      * @param componentPath the path to the component
      */
     deleteComponentConfiguration(componentPath: string): Promise<void>;
+
+    /**
+     * Create a component from the given devfile template project.
+     *
+     * @param componentPath the folder in which to create the project
+     * @param componentName the name of the component
+     * @param devfileName the name of the devfile to use
+     * @param registryName the name of the devfile registry that the devfile comes from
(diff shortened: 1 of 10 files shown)


JessicaJHee (commented):
Looking great! I haven't looked at the code yet but I have a few comments/questions:
* Personally I prefer the buttons such as `open project in browser` and `next` in the template project selection card and the final `create component` to be visible but disabled when the inputs have an error, instead of the error alert covering it. Imo the helpText under the textField/select is enough, what do you think?
* Would it be possible for the `use different template project` button to take the user back to the select template project popup? (1 step back instead of 2)
* Could we add a `x` button on the top right corner on the select template project popup? It would be a nice addition to just clicking outside of the popup

datho7561:
> Looking great! I haven't looked at the code yet but I have a few comments/questions:

I'll address these in a new PR, so that you can work with the "Set name and folder" view from this PR in the mean time.

datho7561:
> - Would it be possible for the use different template project button to take the user back to the select template project popup? (1 step back instead of 2)

I kind of don't want to do this:
- It's difficult given the design of the component
- Most of the devfiles only have 1 template project, so in order to change template projects, you will need to exit to the devfile search

PR #3017, opened 2023-07-12
Address comments on "from template project" workflow

datho7561:
- add a close button for the modal
- always show the "open in browser" and "next" buttons in the modal and enable/disable them depending on if a project is selected. show the validation message as helper text if the user interacts with the "select project" dropdown but doesn't select a project


datho7561:
@JessicaJHee feel free to focus on your workflow for now instead of reviewing this

JessicaJHee (commented):
Looks great thank you! Do you mind doing the same thing for this case where you leave the buttons disabled as well?

image

datho7561:
> Do you mind doing the same thing for this case...

Alright, should be done

JessicaJHee (approved):
LGTM thanks very much!

PR #3022, opened 2023-07-13
"Delete project" warns about deleting project contents

### LINKED ISSUE #2023: Request if components should be deleted when deleting Application or Project from Application explorer
As @jrichter1 again pointed out current deletion command behavior for Project or Application is inconsistent. There should be common approach when deleting resources with multiple nested resources in it. There should be confirmation from user requested if all the components within project or application should be deleted as well.

datho7561:
- Add some integration tests for the `oc` commands we call in the file

Closes #2023


msivasubramaniaan (approved):
LGTM

PR #3024, opened 2023-07-14
UI consistency (see notes)

datho7561:
In order for all pages of the wizard to be more consistent:
- use `<Container />` as parent for all components, so that all the pages are the same width
- all page titles are left justified
- `5em` of margin at the top of each page


JessicaJHee (approved):
LGTM!

PR #3030, opened 2023-07-17
Require 'git-up' as a runtime dependency

### LINKED ISSUE #3029: Extension doesn't launch when run as `.vsix`
In an empty workspace, while logged into a cluster, right click on the cluster, then select "Logout". I get the following output

```
Error running command openshift.explorer.logout: command 'openshift.explorer.logout' not found. This is likely caused by the extension that contributes openshift.explorer.logout.
```

datho7561:
Fixes #3029


msivasubramaniaan (approved):
LGTM

PR #3032, opened 2023-07-18
Remove clusterRequired annotation for dev on podman command

### LINKED ISSUE #2992: Running "Start Dev on podman" should not require being logged into a cluster
When you attempt to run a component in dev mode but are not connected to a Kubernetes cluster, you get a popup that says "Login in to a Cluster to run this command.".

This is because in https://github.com/redhat-developer/vscode-openshift-tools/commit/9815ad261c98781bb236dfc2610096bc929059be we annotate the function that defines the "Start Dev" command with `@clusterRequired`, and this function is reused for "Start Dev on podman".

However, "Start Dev on podman" doesn't require a cluster to work. We should avoid checking for a cluster for the "Start Dev on podman" case.

JessicaJHee:
Fixes #2992

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -204,7 +204,7 @@ export class Component extends OpenShiftItem {
 
     @vsCommand('openshift.component.dev.onPodman')
     static async devOnPodman(component: ComponentWorkspaceFolder) {
-        return Component.dev(component, 'podman');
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
Looks good to me. Thanks, Jessica!

PR #3033, opened 2023-07-18
Add advanced git option functionality and error check

JessicaJHee:


datho7561:
If I hit the back button after the clone fails, it returns to the "cloning in progress" screen.

If I enter anything into the "Context Dir" field, the clone fails. I think we can remove this field, since it's not clear what it's for.

datho7561:
My bad, I meant to review this earlier today. I'll take another look now.

datho7561:
One more small change: "Select a different devfile" is still clickable if the git clone fails. Other than that, this looks good!

JessicaJHee:
@datho7561 thanks for the review! Could I make that change in the next PR? I already made some changes on this branch to address the feedback from the demo

datho7561:
Okay, sounds good! I'll merge this for now

PR #3037, opened 2023-07-19
Check podman installation for start dev on podman

### LINKED ISSUE #2808: Add installation of podman binary if not present in the system
When running odo dev on podman, if we detect that podman is not running, the extension should prompt installation of podman binary and then proceed with the workflow.

JessicaJHee:
Fixes #2808


datho7561 (approved):
Looks good! Thanks, Jessica!

PR #3038, opened 2023-07-19
Delete existing 'create component' parts

datho7561:
- Delete the quick pick based create component
- Delete the 'git import' workflow
- Replace the devfile registry viewer with the one contributed in create component
- Fix a few bugs I found:
  - If you add the staging devfile registry, and enable and disable it repeatedly in the devfile search, the list of items will grow
  - Maybe some others I forgot


datho7561:
> the create component button in both the application explorer and the components window?

That's a good point. I had it there since there were already buttons for create component and import from git, but I think it makes sense to remove completely. My push didn't address this yet.

>  I also get an error with devfileRegistry webview: Could not resolve "../../common/vscode-theme".

I pushed a commit which should fix this.

Issue #3016, opened 2023-07-12
Investigate and fix the reason why the import-from-git, devfile-registries and openshift-view tests are unstable


lgrossma:
These are results of 3 test runs of the same commit - each with different results.
![Screenshot 2023-07-12 at 11 19 23](https://github.com/redhat-developer/vscode-openshift-tools/assets/86789336/540e5d65-07e6-457f-838e-4d70b57e9127)
![Screenshot 2023-07-12 at 11 19 04](https://github.com/redhat-developer/vscode-openshift-tools/assets/86789336/15ea34ac-7049-40a3-9066-dc1afecc1987)
![Screenshot 2023-07-12 at 11 18 50](https://github.com/redhat-developer/vscode-openshift-tools/assets/86789336/764a1873-66fd-4c43-a0ca-19752d696e09)

PR #3043, opened 2023-07-21
Stabilizing UI tests

lgrossma:
Stabilizing devfile registries tests, import from git tests and extension tests
See: https://github.com/redhat-developer/vscode-openshift-tools/issues/3016

### DIFF
--- a/test/ui/suite/devfileRegistries.ts
+++ b/test/ui/suite/devfileRegistries.ts
@@ -15,21 +15,21 @@ export function testDevfileRegistries() {
         let driver: WebDriver;
 
         before(async function context() {
-            this.timeout(10000);
+            this.timeout(10_000);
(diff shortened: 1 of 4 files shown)


lgrossma:
Not sure if I can do much more right now. The most problematic tests seem fixed for now, but sometimes some new failure appears that has not appeared in last 10 runs..

datho7561 (approved):
Yeah, I still see failures in the before-all hook for the git import test, which I think is new.

4/5 runs I did failed, although for two of the runs all the tests passed except git import.

I don't think it's worth putting much more effort into stabilizing git import. We are hoping to merge the rewrite of that interface early next week and release. I'll help rewrite the tests for that UI.

I'm good to merge this PR as is.

PR #3045, opened 2023-07-21
Address some of the feedback on 'Create Component' (see notes)

datho7561:
- Clear field button for search field in devfile search
- Select the template project by default if there is only one
  - Handle the case that there are no template projects (you can test with the Universal Developer Image devfile)
- Change icon buttons to use secondary text colour
- Prepopulate the component name field based on the git url or template project name
- Rename `fromExistingGitRepo.tsx` to address typo
- Add a copy button for devfile YAML
- Focus existing Create Component webview instead of opening a second instance (There are issues with how message passing is handled if there are multiple instances of Create Component running)
- Make devfile search scale better when zooming (there is no horizontal scrolling up to zoom level 4 on my 1440p monitor)

### DIFF
--- a/src/webview/common/devfileListItem.tsx
+++ b/src/webview/common/devfileListItem.tsx
@@ -21,68 +21,76 @@ export function DevfileListItem(props: DevfileListItemProps) {
             sx={{ width: 'calc(100% - 16px)' }}
             alignItems="center"
             paddingX={1}
-            spacing={3}
+            justifyContent="space-between"
         >
-            <Box
-                sx={{
-                    display: 'flex',
-                    width: '7em',
-                    height: '7em',
-                    bgcolor: 'white',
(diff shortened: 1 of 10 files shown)


JessicaJHee (commented):
Looking really nice! A few things I noticed:
* It would be nice to fix the width of the search bar while typing
* There is some extra spacing between the template project drop down and the `Open Project in Browser` and `Next` buttons
* When the component name is auto filled, there is still the `please enter a component name` helper text under the text field

datho7561:
> There is some extra spacing between the template project drop down and the Open Project in Browser and Next buttons

I reworked the layout a bit to reduce the spacing, let me know what you think.

JessicaJHee (approved):
LGTM, thanks!

PR #3047, opened 2023-07-21
Make all the buttons use the filled variant

datho7561:


JessicaJHee:
@datho7561 I think we are good to close this

datho7561:
Good catch!

PR #3048, opened 2023-07-24
Replace use devfile button with clickable devfile list items

JessicaJHee:
Peek 2023-07-21 17-49


datho7561 (changes requested):
I like this change, I think it's nearly ready for merging. However:
- Please address the merge conflicts
- After the rebase, it should be possible to simplify the component by removing one of the horizontal stacks I created, since I grouped things like this (pseudocode):

```xml
<stack horizontal>
  <stack horizontal>
    <devfileLogo />
    <stack vertical>
      <name />
      <description />
      <tags />
    </stack>
  </stack>
  <button />
</stack>
```

(you should be able to get away with only one horizontal stack)

PR #3049, opened 2023-07-24
Fix SVG component

msivasubramaniaan:


datho7561:
The welcome page still doesn't load, and I still experience https://github.com/redhat-developer/vscode-openshift-tools/pull/3003#issuecomment-1648063844.

msivasubramaniaan:
> The welcome page still doesn't load, and I still experience [#3003 (comment)](https://github.com/redhat-developer/vscode-openshift-tools/pull/3003#issuecomment-1648063844).

Pushed the code. Can you try it now?

PR #3050, opened 2023-07-24
Telemetry for create component wizard

datho7561:


JessicaJHee (commented):
Looks good! I wonder if we should include the error message when the create component fails? It would also be nice to send a telemetry when the user exits/cancels out of the workflow but I think we can leave that for the future?

datho7561:
I added the error message to the error telemetry and an event for when the wizard closes. The wizard close event will trigger regardless of if the user closes early or as a part of completing the wizard.

USAGE_DATA.md
```diff
@@ -21,14 +21,15 @@ When a command contributed by extension is executed, telemetry event sent with f
...
-* component_type - name of the component type from devfile registry or catalog
-* component_version - version (for S2I components)
-* starter_project - name of a starter project used (for devfile components)
-* use_existing_devfile - boolean attribute to indicate if an existing devfile was used rather than one from registry
+In addition to generic command's usage data (see above) `New Component` command and wizard also reports the following:
+
+* `newComponentCreated` - sent when a new component is created, with the following properties:
+    * `strategy` - indicates whether the component was created from an existing local codebase, an existing git repository, or a template project
+    * `component_type` - name of the devfile used
+    * `starter_project` - name of the template project used (for components created from a template project)
```
mohitsuman, line 29: Remove bracket.

mohitsuman (changes requested):
minor typo

PR #3051, opened 2023-07-24
Remove broken coverage badge from README

### LINKED ISSUE #2906: Coverage badge in README.md is broken
The coverage reported in README.md is 65%, however the tests report ~35% coverage. According to https://shields.io/category/coverage, the codecov badge should only be used in private repositories, since it requires passing a token in the URL to work (this is BAD). We should consider removing the badge, since as it is, it's misleading.

datho7561:
Fixes #2906


mohitsuman (approved):
lgtm

PR #3052, opened 2023-07-24
Delete cloned repository on exit/cancel

JessicaJHee:


src/webview/create-component/createComponentLoader.ts
```diff
@@ -232,8 +235,8 @@ export default class CreateComponentLoader {
                             // move the cloned git repo to selected project path
                             componentFolder = path.join(message.data.gitDestinationPath, componentName);
                             await fs.mkdir(componentFolder);
-                            await fse.copy(message.data.tmpDirUri.fsPath, componentFolder);
-                            await fs.rm(message.data.tmpDirUri.fsPath, { force: true, recursive: true });
+                            await fse.copy(tmpFolder.fsPath, componentFolder);
+                            await fs.rm(tmpFolder.fsPath, { force: true, recursive: true });
```
datho7561, line 239: I get the following in the extension host logs if I step through all the "import from git" steps:

```
rejected promise not handled within 1 second: Error: ENOENT: no such file or directory, stat '/tmp/tmp-23687-2JNsu65ekALv'
```

I think delete is being triggered twice on the folder, once when you move the contents and another time when the folder is emptied. The second time fails with an error message in the console, since the directory doesn't exist.

I think it would be nice to avoid this, although it's not a huge deal, since it's hidden from the user (no pop ups)

datho7561 (changes requested):
One small comment, but otherwise, works very well!

datho7561 (changes requested):
After a fresh launch, if I close the webview without selecting a strategy:

```
rejected promise not handled within 1 second: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined
extensionHostProcess.js:105
stack trace: TypeError: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined
```

datho7561 (approved):
Looks good!

PR #3064, opened 2023-07-26
New create component UI

datho7561:
- Combine 'Create component from existing local codebase', 'Import from git', and 'Create component from template project' into one webview
- Replace the devfile registry view with the one written for the new 'Create component' webview
- Disable some of the UI tests related to creating a component, for example "import from git", as they will need to be rewritten for the new UI


datho7561:
Demo walking through the UI:

https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/989a3b56-24ff-49f4-9d9e-3c96098cd080

msivasubramaniaan:
Good to see the new UI of create component. kudos to @datho7561 and @JessicaJHee

Here are my observation:

1. Open Devfile registry view
    Edit the name of any one of the registry from tree view
    The opened devfile registry view not updated

2. There was an inconsistent behavior on search
issue-search

3. It would good to skip the folders(if already devfile.yaml) from dropdown on existing local code base workflow
image

4. Disable `Next` button if URL not reached
image

5. Don't show the `temp folder` creation to the user
image

6.When search text ends with space
image

7. Open Project in browser not working
issue-open-browser

8. Why there was an `...` chip?
image

msivasubramaniaan (changes requested):

datho7561:
Thanks so much for the helpful feedback! I'm working on addressing the comments, but have a few questions about these two:

> 6. When search text ends with space

What do you think I should do in this case? Right now I'm doing a simple substring search. Another approach I could take is split the input into words and only show the result if it contains each of the words. Does this approach sound good, or did you have something else in mind?

> 8. Why there was an ... chip?

I'm only displaying the first 4 tags. Some of the devfiles (like Wildfly and Universal Developer Image) have more than that (they have 7 and 10 tags, respectively). I was hoping to indicate to the user that there were more tags that aren't displayed. Do you think having the `...` tag makes sense, or should I just show the first 4 tags without the `...` tag?

msivasubramaniaan:
> Thanks so much for the helpful feedback! I'm working on addressing the comments, but have a few questions about these two:
> 
> > 6. When search text ends with space
> 
> What do you think I should do in this case? Right now I'm doing a simple substring search, so the name or one of the tags needs to contain the input exactly. Another approach I could take is split the input into words and only show the result if it contains each of the words. Does this approach sound good, or did you have something else in mind?
>
>  _I would suggest to split the inputs and search accordingly, if input is empty just omit it._
> > 8. Why there was an ... chip?
> 
> I'm only displaying the first 4 tags. Some of the devfiles (like Wildfly and Universal Developer Image) have more than that (they have 7 and 10 tags, respectively). I was hoping to indicate to the user that there were more tags that aren't displayed. Do you think having the `...` tag makes sense, or should I just show the first 4 tags without the `...` tag?
>
> _I think better you don't show the `...` tag. If you want to show then on hover just display the list of remaining tags name._

datho7561:
I added hover text for the `...` chip, and adjusted how the search works using the algorithm I described.

datho7561:
Feedback from Mohit, along with progress:
 - [ ] Description for the first page in Create Component
 - [ ] Make most buttons use the filled variant
 - [ ] Replace the "From Template" icon
 - [ ] Replace the "Create Component" panel button icon with a `+`
 - [ ] Update panel welcome content to mention that you can import from git, use a local project, or use a template project
 - [ ] Add a check box on the final step of the wizard to indicate if the component should be added to the user's workspace. The checkbox should be enabled by default
 - [ ] Investigate getting the Quarkus starter projects to work (they are a zipped folder)
 - [ ] Always select the first template project by default, even when there are multiple
 - [ ] Rearrange the display order of the devfiles so that Quarkus is first, followed by the devfiles with debug support, followed by the rest of the devfiles
 - [ ] Fix importing a local empty folder
 - [ ] Add telemetry event for copy devfile to clipboard
 - [ ] Add telemetry event for open in browser
 - [ ] Add telemetry event for selecting a strategy
 - [ ] Add telemetry event for crash in webview

datho7561:
Alright. All the feedback from Mohit is addressed. I also found and fixed some bugs. Here's what I put down for the description on the first page of the wizard:

AddDescriptionToCreateComponent

@mohitsuman @JessicaJHee let me know if it's good.

msivasubramaniaan:
Hello @datho7561,

I can confirm that the feedback from @mohitsuman works as expected in the latest code. Also found some issues, it would good if we fix this below issues as well.

1. Duplicate entry in Folder Selection
image
2. Once user selects the folder from file-picker then the corresponding path should default in dropdown
3. Also noticed file path starts with `/` `(eg: /c://<folder_name>/<file_name>)`
4. Open Project in Browser of Quarkus shows below error message
image

msivasubramaniaan (changes requested):
Mentioned the issues in comment

datho7561:
I can't seem to reproduce 1). I think it's a Windows-specific issue that has the same root cause as 3). I will put up a PR that addresses the other issues soon.

rgrunber (approved):

Issue #3028, opened 2023-07-17
The new "OpenShift Terminal" doesn't decrease size correctly


datho7561:
If you increase then decrease the size of the OpenShift Terminal by expanding the panel upwards and then downwards, the last lines of output of the terminal are hidden under the bottom of the webview:

https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/eb45a9c4-e78f-44b1-8d9a-635b9d9c731f

Ideally, the "Building your application" line should not be cut off when the panel is resized.

datho7561:
Don't know when this got fixed, but I can't reproduce it any more.

PR #3065, opened 2023-07-26
Re-enable separate OpenShift Terminal

datho7561:
Revert the revert commit, and address the bug preventing the terminal from working on Windows. The resolution was to wrap all program executions in a Windows command line shell. I chose `cmd.exe`, although I expect PowerShell would also have worked.

WIP:
- [x] Fix it on Windows again (it's not working any more)
- [x] Fix resizing (see https://github.com/redhat-developer/vscode-openshift-tools/issues/3028)


datho7561:
@msivasubramaniaan I still need to fix a few places where commands are still being run in the default VS Code terminal, but features like 'Describe' and 'Dev' should work now. If you have time at some point, could you please confirm that this works for you?

datho7561:
I added support for the serverless commands and fixed a race condition that I ran into.

datho7561:
@msivasubramaniaan I've updated this PR to fix it on Windows and take into account the Serverless commands (build, run, deploy). If you have time could you please let me know if it's working for you?

datho7561:
rebased on Muthu's and Victor's PRs:
- invoke should run in the separate terminal
- `odo run` should run in the separate terminal if `"openshiftToolkit.useWebviewInsteadOfTerminalView": false`

src/webview/openshift-terminal/openShiftTerminal.ts
```diff
@@ -0,0 +1,588 @@
...
+        this._disposables.push(this._termSerializer);
+    }
+
+    startPty() {
+        if (platform() === 'win32') {
+            const escapedArgs = Array.isArray(this._args)
+                ? this._args.join(' ')
+                : this._args;
+            this._pty = ptyInstance.spawn(
+                'C:\\WINDOWS\\system32\\cmd.EXE',
```
vrubezhny, line 157: I'm not sure, but is it possible to use `COMSPEC` System Environment Variable here, so it  correctly finds `cmd.exe`?

> C:\Windows\System32\cmd.exe or if running a 32 bit [WOW](https://ss64.com/nt/syntax-64bit.html) - C:\Windows\SysWOW64\cmd.exe
> Comspec is used whenever the command shell spawns a new process, e.g. the [FOR](https://ss64.com/nt/for.html) command, if comspec is not defined those processes will fail to launch.

vrubezhny (commented):
Please look at the question in-line

src/webview/openshift-terminal/openShiftTerminal.ts
datho7561: I'll try it out...

vrubezhny:
The most annoying problems (running on linux) I found are:
- OS Terminal doesn't have a ruler at right to make possible to scroll the terminal contents by mouse (a "plain" vscode terminal has such a ruler)
- OS Terminal has an issue that does not allow to scroll down to the very bottom line - sometimes it works, but after switching tabs or doing some actions  (like trying to create a serverless function) and switching back to OS Terminal  tab, it doesn't allow to see the bottom line (or even lines) - the bottom line(-s) are "cut" or even not visible at all.
- A new OS Terminal tab is opened for every run of the same Component Command (personally I think we should reuse tabs at least if we run the same command multiple times and probably to have a possibility to clear the contents of a terminal tab)
- No "Clear Tab Contents"-like action available
- There are `Cut/Copy/Paste` actions, but there is no `Select All` action available

datho7561:
Thanks for the feedback! I think I will need to go back to this feature and do more work on it, so we should avoid trying to get it in the next release.

vrubezhny:
> Thanks for the feedback! I think I will need to go back to this feature and do more work on it, so we should avoid trying to get it in the next release.

I agree, probably we should postpone it for a while. 
But my overall impression is good and, when resizing/scrolling issues are fixed, I'd like to see it included, so we can open the issues  for the rest of the problems and solve them one by one.

src/webview/openshift-terminal/openShiftTerminal.ts
datho7561: looks like the library I'm using contains a mechanism to resolve the location of cmd.exe, so I don't need to provide an absolute path

datho7561:
> Pressing Ctrl-X/Ctrl-C/Ctrl-V closes the OS Terminal Tab instead of performing a Cut/Copy/Paste action (we're in VSCode, not in a *nix console, right?)

This is the same behaviour as the regular VS Code Terminal. In order to cut/copy/paste using the keyboard, use `Ctrl+Shift+X/C/V`

datho7561:
Hello Victor. If you have time, could you please re-review this PR? I addressed the first two of your suggestions. For the other ones:
- *A new OS Terminal tab is opened for every run of the same Component Command*

I think we should handle this in a separate PR.

- *No "Clear Tab Contents"-like action available*
- *There are Cut/Copy/Paste actions, but there is no Select All action available*

The default VS Code Terminal has buttons in the top right:

button-gutter-openshift-terminal

What do you think about me adding buttons for "select all" and "clear tab contents" here? I don't think the extension API gives access to add a "select all" context menu.

- *Pressing Ctrl-X/Ctrl-C/Ctrl-V closes the OS Terminal Tab instead of performing a Cut/Copy/Paste action (we're in VSCode, not in a \*nix console, right?)*

The terminal behaves like a *nix or Windows terminal, since in come cases you need to send `Ctrl+C` to the process (eg. to stop component dev mode). In order to copy, you can use `Ctrl+Shift+C`. This is the default behaviour of the library I'm using, which is the same library and behaviour as VS Code's built-in terminal.

vrubezhny (dismissed):
Looks good, works smooth. The main problem, when it was impossible to scroll to the very beginning/very end of terminal text, is definitely solved.

I'm :+1:  for it to be merged once successfully rebased.

All the other issues and improvements may be discussed/solved in separate issues.

datho7561:
Annoying bug on Windows: Right click -> copy doesn't work, and `Ctrl+Shift+C` is apparently a keybinding to open the Windows Terminal app, so there is no way to copy text.

datho7561:
The keybinding for copy on Windows is `Ctrl+Ins`

PR #3071, opened 2023-07-28
Fix uninstalling Helm charts

### LINKED ISSUE #3021: Uninstalling Helm releases only works when the name of the release ends with the name of the chart
Let's say you create a Helm release called `my-janus`, based on the `backstage` Helm chart from the OpenShift Helm repository:

helm-release-uninstall-pt1

It will appear in the "Application Explorer" panel as `my-janus-backstage`:

helm-release-uninstall-pt2

If you attempt to use the context menu to uninstall the Helm release:

helm-release-uninstall-pt3

Then nothing happens.

This is because the command is trying to uninstall the release `my-janus-backstage`, but the release is called `my-janus`. However, if you name the release `my-janus-backstage`, then the uninstall works.

The root cause is that the item that appears in the "Application Explorer" isn't the Helm release, it's a deployment created by the Helm chart, which is only one of the components making up the Helm release. In order to address this issue, we should list Helm releases in the "Application Explorer" tree view.

You can run `helm list -o json` to get a JSON list of the Helm releases for the current namespace.

datho7561:
- Show Helm releases in the project explorer using results from `helm list`
- Add description text for items in the application explorer to help distinguish them
- Refactor Helm commands into their own folder
- Add integration tests for Helm commands

Fixes #3021


datho7561:
Here's what the Application Explorer UI looks like after this change:

ShowHelmChartsInApplicationExplorer2

In this demo, I have two charts installed, Redis and Red Hat Developer Hub.

msivasubramaniaan (approved):
LGTM

PR #3073, opened 2023-07-31
deploy and un deploy functionality

msivasubramaniaan:
Related to https://github.com/redhat-developer/vscode-openshift-tools/issues/3010


datho7561:
I tested deploying onto the sandbox cluster, since I couldn't figure out how to set up Knative serverless on `crc`.

Here are a few things I ran into:
- When I run `func deploy` on the command line it gives information on the progress and at the end it outputs the URL of the function. I didn't see any of this information when using the "Deploy" context menu. Is this intentional? I think it would be nice to display the URL somewhere, either by showing the command output, or having some sort of UI, maybe a button to open/copy the URL.
- If I have deployed the function, but then make a change to the source code, then I can update the version running on the cluster by rerunning `func deploy`. It would be nice to keep the deploy option in the context menu, even if the function is currently deployed

msivasubramaniaan:
> I tested deploying onto the sandbox cluster, since I couldn't figure out how to set up Knative serverless on `crc`.
> 
> Here are a few things I ran into:
> 
> * When I run `func deploy` on the command line it gives information on the progress and at the end it outputs the URL of the function. I didn't see any of this information when using the "Deploy" context menu. Is this intentional? I think it would be nice to display the URL somewhere, either by showing the command output, or having some sort of UI, maybe a button to open/copy the URL.
> * If I have deployed the function, but then make a change to the source code, then I can update the version running on the cluster by rerunning `func deploy`. It would be nice to keep the deploy option in the context menu, even if the function is currently deployed

@datho7561 addressed the feedback that you provided. Please have a look

PR #3080, opened 2023-08-03
Add context menu to create Routes

### LINKED ISSUE #3072: Context menu to expose a `Deployment` as a `Route` or `Ingress`.
It would be nice to expose a running service to the internet using a `Route` (in the case of OpenShift) or `Ingress` (in the case of other Kubernetes distributions).

Here's how to create a `Route` in OpenShift, given that `$NAME_OF_DEPLOYMENT` is a deployment that's in the current name:

`oc expose deployment $NAME_OF_DEPLOYMENT`

Here's an example of an `Ingress` `yaml` in Kubernetes:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: wildwest-backend-ingress
spec:
  defaultBackend:
    service:
      name: wildwest-backend-app
      port:
        number: 8080
```

The values we will have to change depending on the deployment are:
- `metadata.name` to the name of the deployment but with the suffix `-ingress` to help distinguish it
- `spec.defaultBackend.service.name` to the name of the deployment
- `spec.defaultBackend.service.port.number` to the port that the deployment exposes its service on

In order to create the ingress, you will need to run `oc apply -f ingress-definition.yaml`.

A few final thoughts:
1. It might also be possible to include the route or ingress in the devfile by embedding the YAML using the additional Kubernetes resources feature. For instance, this is how odo creates `ServiceBinding`s:
```yaml
// ...
components:
- kubernetes:
    inlined: |
      apiVersion: binding.operators.coreos.com/v1alpha1
      kind: ServiceBinding
      metadata:
        creationTimestamp: null
        name: my-binding
      spec:
        application:
          group: apps
          kind: Deployment
          name: using-go-app
          version: v1
        bindAsFiles: true
        detectBindingResources: true
        services:
        - group: postgres-operator.crunchydata.com
          id: my-binding
          kind: PostgresCluster
          name: example-postgres
          namespace: default
          resource: postgresclusters
          version: v1beta1
      status:
        secret: ""
  name: my-binding
// ...
```
  The benefit of this approach is that, if I understand correctly, `odo` will manage the lifecycle of this Kubernetes object by creating it when dev is started and deleting it when dev is stopped

2. A final aspect of this change is that we should alert the user that this will expose the service to the public if the cluster is capable of exposing services to the public, and require an additional confirmation in order to do this. We shouldn't try to detect if the cluster is capable of exposing public endpoints, since that is really difficult. Instead we should always show this warning.

3. Not all Kubernetes clusters are capable of creating an ingress. They require an ingress controller. On `minikube` the ingress controller that is automatically set up is the `nginx` one. Maybe we should hide the UI option if the cluster is not OpenShift and doesn't have an ingress controller set up?

datho7561:
- Add a context menu item in order to add a route for a given service
- Add a context menu item in order to delete the route for a given service
- Add a button when there is a route to open the route in the browser
- Add a button when there is a route to copy the route URL

I think that we should wait on working similar funcitonality for Ingresses in a future PR (see [my comment on the issue](https://github.com/redhat-developer/vscode-openshift-tools/issues/3072#issuecomment-1664563162)).

Closes #3072

### DIFF
--- a/package.json
+++ b/package.json
@@ -827,6 +827,28 @@
 				"command": "openshift.Serverless.stopRun",
 				"title": "Stop",
 				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.resource.exposeAsRoute",
+				"title": "Expose service as Route",
+				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.resource.unexpose",
+				"title": "Remove Route for service",
+				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.resource.openUrl",
+				"title": "Open Service Url",
+				"category": "OpenShift",
+				"icon": "$(link-external)"
+			},
+			{
+				"command": "openshift.resource.copyUrl",
+				"title": "Copy Service Url",
+				"category": "OpenShift",
+				"icon": "$(link)"
 			}
 		],
 		"keybindings": [
@@ -1153,8 +1175,23 @@
 				{
 					"command": "openshift.Serverless.stopRun",
 					"when": "false"
+				},
+				{
+					"command": "openshift.resource.exposeAsRoute",
+					"when": "false"
+				},
+				{
+					"command": "openshift.resource.unexpose",
+					"when": "false"
+				},
+				{
+					"command": "openshift.resource.openUrl",
+					"when": "false"
+				},
+				{
+					"command": "openshift.resource.copyUrl",
+					"when": "false"
 				}
-
 			],
 			"view/title": [
 				{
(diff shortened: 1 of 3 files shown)


datho7561:
I think we should hold off on merging this for now. Tim deBoer mentioned it would be better to get things working across all types of clusters instead of depending on OpenShift features.

datho7561:
After I did some playing around and learning, I realized we don't need this to demo the Wild West application. I think that this is a bad idea, since it only works on OpenShift. Also, the cluster might be configured such that creating a `Route` exposes the app that the user is actively developing to the public, which isn't good.

PR #3084, opened 2023-08-04
Fix Create Component performance issues

### LINKED ISSUE #3083: Performance issues for new "Create Component"
Roland mentioned that his fan started spinning audibly faster when the Create Component or Devfile Registry view was opened.

This is notably different from the performance of the old Devfile Registry.

I investigated a bit, and found that something was causing the render function for the view to run on a loop.

datho7561:
- Use `ErrorBoundary` component from a library instead of my hand coded one
- Use an empty array in the `init` `React.useEffect` calls (omitting the array causes the render function to loop)

Closes #3083


rgrunber (approved):
Thanks for handling this.

Here's the before/after for anyone interested.

**Before**
Screenshot from 2023-08-04 14-26-20

where each of those "purple" segments was :

Screenshot from 2023-08-04 12-15-51

**After**
Screenshot from 2023-08-04 14-23-44

PR #3085, opened 2023-08-04
Add  Create Component From Folder action to workspace folder

### LINKED ISSUE #3039: Add `Deploy to OpenShift/Podman` action to folder in workspace
When using VSCode to create a new component in a project, the "Create Component" functionality currently requires selecting a workspace for the component. 

 This issue proposes the addition of an "Deploy to OpenShift/Podman" option to streamline the process and eliminate the need to navigate through files and directories repeatedly.

**Acceptance Criteria:**
- [ ] The "Create Component" functionality should be enhanced with an additional option named "Deploy to OpenShift/Podman"
- [ ] When the user selects the "Deploy to OpenShift/Podman" option, the component creation process should automatically use the currently opened workspace directory as the location for the new component.
- [ ] when the current workspace is not a valid project directory, an informative error message should be displayed, indicating that a valid workspace must be open for this option to work.
- [ ] The "Deploy to OpenShift/Podman" option should be displayed prominently in the Folder right click context menu, ensuring its visibility and accessibility to users.
- [ ] If the user chooses to create a component using the traditional method of selecting a workspace manually, the existing functionality should remain unaffected.
- [ ] Proper error handling should be implemented to gracefully handle any unexpected scenarios or errors that may arise during the process.
- [ ] The new feature should be well-documented in the Welcome and Walkthrough documentation, including clear instructions on how to use the "Deploy to OpenShift/Podman" option.

**Additional Notes:**
This enhancement would greatly improve the user experience for developers working with VSCode, making the process of creating components within an existing workspace more efficient and user-friendly. The implementation should be carefully reviewed and tested to ensure its seamless integration with the existing functionality and to avoid introducing any regressions.

JessicaJHee:
Fixes #3039 

createComponentFromWorkspaceFolder


src/webview/create-component/createComponentLoader.ts
```diff
@@ -92,6 +92,12 @@ export default class CreateComponentLoader {
         );
         panel.webview.html = await loadWebviewHtml('createComponentViewer', panel);
         CreateComponentLoader.panel = panel;
+
+        if (folderPath) {
```
datho7561, line 96: There are a few options to make this code simpler:
- If the `folderPath` isn't set, don't send the message
- Send the message regardless of the value of `folderPath`. Instead of using an if statement, let the webview determine if it's necessary to change views based on if `folderPath` contains a path

src/webview/create-component/pages/createComponent.tsx
```diff
@@ -100,7 +107,9 @@ export default function CreateComponent() {
             case 'fromLocalCodeBase':
                 return (
                     <div style={{ marginRight: '5em' }}>
-                        <FromLocalCodebase setCurrentView={setCurrentView} />
+                        {folderPath != undefined &&
```
datho7561, line 110: `!folderPath &&`

src/webview/create-component/pages/createComponent.tsx
```diff
@@ -131,7 +140,9 @@ export default function CreateComponent() {
                     paddingBottom: '16px',
                 }}
             >
-                {renderComponent()}
+                {folderPath != undefined &&
```
datho7561, line 143: `!folderPath &&`

datho7561 (changes requested):
A few small changes, but it appears to be working pretty well.

src/webview/create-component/createComponentLoader.ts
```diff
@@ -92,6 +92,9 @@ export default class CreateComponentLoader {
         );
         panel.webview.html = await loadWebviewHtml('createComponentViewer', panel);
         CreateComponentLoader.panel = panel;
+
+        panel.webview.postMessage({ action: 'initFromRootFolder', rootFolder: folderPath });
```
datho7561, line 96: It's possible that the webview hasn't loaded properly when this message is set, so it's unable to process this message, and the webview will be stuck on the loading package.

In practice, I sometimes get this the first time that I launch the webview.

This could be fixed by sending the message in the `'init'` message handler

PR #3090, opened 2023-08-09
[WIP] refactor get all components and perform in the background

datho7561:


datho7561:
Still need to think this through more, but I managed to shave of ~3 seconds off the extension start time on Windows by not blocking on `getAllComponents()`. This is important, since the extension startup on Windows is noticably slower than Windows. However, this introduces the problem that if a user immediately goes to create component form template project, it might not have the devfiles loaded at first.

datho7561:
I don't think this is needed any more. I think the slowdown on Windows that I experienced was because my kubecontext was pointing to an invalid cluster, which is a different bug that's been resolved.

Issue #3096, opened 2023-08-13
Create function workflow is broken


mohitsuman:
<img width="1283" alt="Screenshot 2023-08-14 at 2 24 59 AM" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/3133356/082e4ca9-fb3f-483f-b114-844fbb0a114b">

PR #3098, opened 2023-08-14
replaced the regx instead of string

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3096


datho7561 (approved):
Looks good and fixes the use case for me

Issue #3097, opened 2023-08-13
Fix the editor title for functions


mohitsuman:
<img width="308" alt="Screenshot 2023-08-14 at 2 27 22 AM" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/3133356/0bf5e038-6fae-462d-84a8-af10b416e1b1">

Remove the word wizard from the function create page in the editor heading.

datho7561:
I can't seem to reproduce this issue

PR #3099, opened 2023-08-14
title changes removed

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3097

### DIFF
--- a/src/webview/serverless-function/serverlessFunctionLoader.ts
+++ b/src/webview/serverless-function/serverlessFunctionLoader.ts
@@ -47,7 +47,6 @@ async function gitImportMessageListener(panel: vscode.WebviewPanel, event: any):
             break;
         case 'createFunction':
             const selctedFolder: vscode.Uri = vscode.Uri.file(path.join(functionPath.fsPath, functionName));
-            panel.title = `Serverless Function - Create - ${functionName}`;
             await Progress.execFunctionWithProgress(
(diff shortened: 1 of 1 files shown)


src/webview/serverless-function/serverlessFunctionLoader.ts
```diff
@@ -47,7 +47,6 @@ async function gitImportMessageListener(panel: vscode.WebviewPanel, event: any):
             break;
         case 'createFunction':
             const selctedFolder: vscode.Uri = vscode.Uri.file(path.join(functionPath.fsPath, functionName));
-            panel.title = `Serverless Function - Create - ${functionName}`;
```
mohitsuman, line 50: @msivasubramaniaan There should be a title with `Serverless Function - Create`.

mohitsuman (changes requested):

src/webview/serverless-function/serverlessFunctionLoader.ts
msivasubramaniaan: yes it will get the title from https://github.com/redhat-developer/vscode-openshift-tools/blob/55431dd0575860b32b2f412ade2132341f8fd3e2/src/serverlessFunction/view.ts#L203C9-L203C9

datho7561 (approved):
LGTM! Thanks, Muthu!

Issue #3075, opened 2023-08-01
`listServiceInstances` is only used in integration tests


datho7561:
See https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/odo/command.ts#L134

We should remove this function and the associated integration test.

This function is only invoked in tests. There is a similar function `getBindableServices`, which is the one that we actually use for the bind service feature.

PR #3107, opened 2023-08-16
listServiceInstances is only used in integration tests #3075

vrubezhny:
The unused method and according integration test are removed

Fixes: #3075


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3109, opened 2023-08-17
Improve contributing documentation

datho7561:
I added some things in the contributing documentation that I realized we didn't cover well after we onboarded Victor to the project. Notably:
- Note that the Red Hat Authentication extension is required by VS Code OpenShift Toolkit
- Explain that different commands are used to run the different suites, and explain why the suites are split up
- List the different types of clusters that have been well tested with the extension, and the tradeoffs of using them


vrubezhny (approved):
It's great to have this PR merged!

datho7561:
Thanks for the review, Victor!

Issue #3076, opened 2023-08-01
`listCatalogComponents` is only ever invoked in `listCatalogComponentsJson`


datho7561:
Refactor the code so that we only have `listCatalogComponentsJson`. Remove the associated integration test for `listCatalogComponents` (keep the one for `listCatalogComponentsJson`).

PR #3110, opened 2023-08-17
listCatalogComponents is only ever invoked in listCatalogComponentsJs…

vrubezhny:
…on #3076

The listCatalogComponents is incorporated to listCatalogComponentsJson method, listCatalogComponents method and according integration test are removed

Fixes: #3076


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3111, opened 2023-08-17
Handle importing git repos that already have a Devfile

### LINKED ISSUE #3104: Import from git fails if there is an existing devfile
eg. Go through "Import from Git" workflow, using the following repository: https://github.com/openshift-evangelists/Wild-West-Frontend

https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/59979fbd-7feb-4b99-9611-07bdcfed1a4a

The expected behaviour is that you should be able to clone the repository if it contains an existing devfile, and that the existing devfile should be preserved when the repository is cloned.

datho7561:
Fixes #3104

### DIFF
--- a/src/webview/create-component/createComponentLoader.ts
+++ b/src/webview/create-component/createComponentLoader.ts
@@ -235,6 +235,10 @@ export default class CreateComponentLoader {
                         action: 'cloneFailed',
                     });
                 } else {
+                    void CreateComponentLoader.panel.webview.postMessage({
+                        action: 'devfileExists',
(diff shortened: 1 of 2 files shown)


datho7561:
@vrubezhny If you have time at some point, could you please review this PR? Let me know if you have any questions about it. Hopefully the video in the issue explains the use case well enough.

vrubezhny (approved):
Looks good to me

rgrunber (approved):
Same. This fixes the issue for me.

Issue #2811, opened 2023-03-23
Add tree elements including the devfile commands in component views


mohitsuman:
Currently the components views only mentions the component name with the status if dev mode is running or not. We need to add a tree view to every component mentioning the devfile commands present.

Reference image attached:

![component-devfile](https://user-images.githubusercontent.com/3133356/227239685-d753a89a-c4d3-4e1d-a80c-f8b221647103.png)

vrubezhny:
@mohitsuman @datho7561 I cannot find an icon for "Run Component Command"-like menu item. Do we have any `svg`-image  suitable  for this kind of command?

datho7561:
Don't know if you've already seen this, but here is the list of the built in icons you can use without needing to add an SVG to the extension: https://code.visualstudio.com/api/references/icons-in-labels. I think the `debug-start` or `notebook-execute` icons are close enough to what we need. What do you think Mohit?

vrubezhny:
I found that other icons like "Refresh" or "Plus" are taken from the https://mui.com/material-ui/material-icons/ pack and I found the  https://mui.com/material-ui/material-icons/?query=start&selected=PlayCircleOutline icon there which looks exactly as the issue image shows.

But I  have a little problem while I'm trying to use this icon from this icon pack - it's not shown when added to a command on the tree item - there is a placeholder, I can click and  run the  associated command - but he icon itself is not displayed. An SVG image (when  I tried to use it) was working fine.

vrubezhny:
@datho7561 I wasn't including any new svg icons, I just took the existing one for testing purpose and used it in `package.json` as the following:
```
{
	"command": "openshift.component.commands.command.run",
	"title": "Run Command",
	"category": "OpenShift",
	"icon": {
		"light": "images/title/light/icon-login.svg",
		"dark": "images/title/dark/icon-login.svg"
	}
},
```
and this was working fine until I changed `"icon" : ...` to:
````
{
	...
	"icon": "${PlayCircleOutline}"
},
```

vrubezhny:
I found that I was wrongly referencing the icons and icons are not  look like taken from https://mui.com/material-ui/material-icons/

So, `$(debug-start)` works for me but... it's green:

![Screenshot from 2023-08-21 17-01-07](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/9bf36add-c7b8-40d8-9d66-53dc5ceb7208)

datho7561:
[...]

PR #3114, opened 2023-08-21
Add tree elements including the devfile commands in component views #…

vrubezhny:
…2811

Fixes: #2811


datho7561:
Works pretty well!

Two things I noticed:
- All of the items in the Components view (including the word "Commands" and each of the commands) all have the same icon. I think it would be nice if each of the commands had an icon that represents a command (maybe terminal-view-icon ?). I don't think that the word "Commands" needs an icon.
- If I modify the `devfile.yaml` for the component by adding a new command (add this under `.commands`):

```yaml
- exec:
    commandLine: echo "hello, world!"
    component: tools
  id: hello-world
```

Then I refresh the components view, the component isn't updated to have the new command. I have to restart VS Code for it to appear. I think we should reread the content of the file when the user manually refresh.

vrubezhny:
>  * All of the items in the Components view (including the word "Commands" and each of the commands) all have the same icon. I think it would be nice if each of the commands had an icon that represents a command (maybe terminal-view-icon ?). I don't think that the word "Commands" needs an icon.

Yes, that needs to be fixed. 

> * If I modify the devfile.yaml for the component by adding a new command (add this under .commands):
> Then I refresh the components view, the component isn't updated to have the new command. 

The commands (and actually as well as  the list of components) are cached.  And this is still a question for me - do we have an event or something like this that I can listen to in order to invalidate the caches when a component is added/removed/changed?

vrubezhny:
@datho7561 Fixed the PR as follows:
- Changed the icons (no more icon for `Components`, used `terminal-view-icon` icon for a `Command`)
- `Refresh Components View`  now clears the cached components thus forcing the Components to be reloaded
[Screencast from 2023-08-22 03-00-23.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/1fc3ca9c-a110-4484-908f-97851500bfe3)

datho7561:
Here are some resources for writing the UI tests:

We have an existing tests for the DevfileRegistries section, which is a TreeView: `./test/ui/suite/devfileRegistries.ts`

The UI testing framework has the following API to add folders to the workspace: https://github.com/redhat-developer/vscode-extension-tester/wiki#opening-files-and-folders.

We have some existing components created under `./test/fixtures/components`. Unfortunately, it looks like VS Code's test fixture mechanism doesn't work with the UI tests, so you will need to add the folder using the API I mentioned.

For the integration tests, in `./src/test/integration/command.test.ts` there is the method `text('dev()', ...` that starts a component in dev mode. You can use a similar method in a `setup` step, then run a test case that runs `odo run` and checks that it works.

vrubezhny:
> Here are some resources for writing the UI tests:
> ...

Thanks. I'll take a look at these tests

> For the integration tests, in `./src/test/integration/command.test.ts` there is the method `text('dev()', ...` that starts a component in dev mode. You can use a similar method in a `setup` step, then run a test case that runs `odo run` and checks that it works.

This is almost exactly what I'm trying to do (I''m just not added it to `setup` step, but tried to run `dev()` right inside of my test case. 
The test suite creates a local component from devfile and uses its location as `componentLocation`... Normally OS Tools runs `dev()` for a component in a terminal started with "current working directory" pointing to component path in workspace. Test doesn't specify any CWD when starting a terminal.  And nothing changes when I try using the component's `componentLocation` I always get:
```
Error: Command failed: "/home/jeremy/projects/vscode-openshift-tools/source/vscode-openshift-tools/out/tools/linux/odo" run build-image
 ✗  unable to get pod for component my-test-component: pod not found for the selector: component=my-test-component. Please check the command 'odo dev' is runnin
 ```
 It looks like pod is really cannot be identified by this location - neither the project  nor component names present in component location path - the server just has nothing related to this `componentLocation` path...

datho7561 (approved):
Looks good and works well. Thanks, Victor!

PR #3115, opened 2023-08-21
3010 create UI for serverless invoke function

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3010

invoke


datho7561:
Seems pretty good so far! A few quick comments, I'll take a look at the code as well.
- Displaying the result as a popup message isn't that helpful, since the content gets cut off if its long. Also, the message is hidden within a few seconds. I think it would be nice to put the result in a new text editor, similar to what "Load" does for Deployments.
- In the case that the function invocation fails, for example when the function isn't running, there is no error message
- In the case of `crc`, we cannot invoke the function, since the certificate isn't trusted, since it's self-signed. It would be nice to configure the certificate so that it's trusted, or at least show an error message as to why it fails

msivasubramaniaan:
> Seems pretty good so far! A few quick comments, I'll take a look at the code as well.
> 
> * Displaying the result as a popup message isn't that helpful, since the content gets cut off if its long. Also, the message is hidden within a few seconds. I think it would be nice to put the result in a new text editor, similar to what "Load" does for Deployments.
> * In the case that the function invocation fails, for example when the function isn't running, there is no error message
> * In the case of `crc`, we cannot invoke the function, since the certificate isn't trusted, since it's self-signed. It would be nice to configure the certificate so that it's trusted, or at least show an error message as to why it fails

Addressed the mentioned comments. Please have a look

src/serverlessFunction/functionImpl.ts
```diff
@@ -126,19 +127,19 @@ export class ServerlessFunctionImpl implements ServerlessFunction {
         return functionList;
     }
 
-    getFunctionStatus(funcData: FunctionContent, deployedFunctions: FunctionObject[]): FunctionStatus {
+    getFunctionStatus(funcData: FunctionContent, deployedFunctions: FunctionObject[]): [FunctionStatus, string] {
```
datho7561, line 130: Add the URL as an optional field of `FunctionStatus` instead of a separate return item

src/webview/serverless-function/app/home.tsx
```diff
@@ -31,28 +58,67 @@ export class ServerlessFunction extends React.Component<DefaultProps, {
...
+                        color: '#99CCF3'
+                    }}>
+                        <Box
+                            display='flex'
+                            flexDirection={'column'}
+                        >
+                            <InvokeFunction instance={this.state.instance} uri={this.state.folderURI} name={this.state.name}
+                                invokeURL={this.state.invokeURL} id={this.state.id} onInvokeSubmit={this.handleInvokeSubmit} />
+                        </Box>
+                    </Container>
```
datho7561, line 99: Move the code surrounding the invoke component and the create function component into their respective components. eg.:

```tsx
this.state.invoke ?
<InvokeFunction />
: <CreateFunction />
```

Then put the title and container into the `InvokeFunction` and `CreateFunction`

datho7561 (changes requested):
The invoke functionality seems to be working. Here are a few things I found in the code.

package.json
```diff
@@ -187,6 +188,7 @@
 		"tslint": "^5.20.1",
 		"typescript": "^4.5.2",
 		"utf-8-validate": "^5.0.7",
+		"uuid": "^9.0.0",
```
datho7561, line 191: There is a built-in library for this in NodeJS: https://nodejs.org/docs/latest-v16.x/api/crypto.html#cryptorandomuuidoptions.

src/serverlessFunction/functionImpl.ts
```diff
@@ -126,19 +127,19 @@ export class ServerlessFunctionImpl implements ServerlessFunction {
         return functionList;
     }
 
-    getFunctionStatus(funcData: FunctionContent, deployedFunctions: FunctionObject[]): FunctionStatus {
+    getCommanFunction(funcData: FunctionContent, deployedFunctions: FunctionObject[]): DeployedFunction {
```
datho7561, line 130: getDeployedFunction

datho7561 (changes requested):
Two more quick comments

datho7561 (approved):
Looks good. Thanks, Muthu!

datho7561:
There are merge conflicts, I can't merge yet

PR #3116, opened 2023-08-22
Run integration tests on KinD

### LINKED ISSUE #2871: Investigate using KinD to run integration tests in GH Actions
intellij-openshift runs integration tests in GitHub Actions by using a KinD cluster. Investigate if something similar can be done for OpenShift Toolkit, since then we can report if these tests are passing on GitHub instead of through the internal Jenkins

datho7561:
- Introduce another environment variable in the tests, `IS_OPENSHIFT`, which indicates if the cluster is OpenShift or another k8s
- Set `IS_OPENSHIFT` to true when running on Jenkins
- Rewrite the tests to skip tests of OpenShift-specific features when the cluster isn't OpenShift
- Modify the CI GitHub Action to run the integration tests on KinD on Linux

*minor annoyance:*

When deleting a namespace in the teardown step of tests, the command wouldn't return when running on a non-OpenShift Kubernetes cluster. (Note that it works fine when running the command on CLI). In order to prevent this from causing timeout errors, I don't `await` these calls.

Closes #2871

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -44,13 +44,6 @@ jobs:
         with:
           run: npm run test:coverage --silent
 
-      # Upload coverage to codecov.io
-      - name: Codecov
-        uses: codecov/codecov-action@40a12dcee2df644d47232dde008099a3e9e4f865 #v3.1.2
-        if: runner.os == 'Linux'
-        with:
-          file: ./out/coverage/coverage-final.json
-
       # UI tests fail under linux
       # Run UI tests
       - name: Run UI Tests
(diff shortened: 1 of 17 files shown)


datho7561:
The project test coverage should have increased, instead it decreased. This likely means that only the coverage of the last suite is being uploaded to codecov, and since the last suite changed to the integration test suite, that number changed dramatically. I'll investigate fixing this.

datho7561:
Looks like operatorhub.io is down ;_;

mohitsuman:
@datho7561 any update on this ?

datho7561:
Rebased, it's been ready for review for a while

coverage.json
```diff
@@ -0,0 +1 @@
+{}
```
datho7561, line 1: I don't think I need this file I'll remove it

vrubezhny (dismissed):
Looks good to me.

Without IS_OPENSHIIFT set:
  51 passing (5m)
  21 pending
  
With IS_OPENSHIFT set:
  57 passing (6m)
  15 pending

72 tests in total in both cases

PR #3118, opened 2023-08-23
Updated CONTRIBUTING docs with info on microshift and k3s

datho7561:
After talking with Armel from `odo` and testing a bit more myself, I figured out how to get MicroShift and k3s working with VS Code OpenShift toolkit. The key is to avoid using the default namespace.


msivasubramaniaan (approved):
LGTM

PR #3125, opened 2023-08-25
[test] Add integration test suite and test for running a component co…

vrubezhny:
…mmand

Issue: #2811

### DIFF
--- a/test/integration/command.test.ts
+++ b/test/integration/command.test.ts
@@ -3,7 +3,9 @@
  *  Licensed under the MIT License. See LICENSE file in the project root for license information.
  *-----------------------------------------------------------------------------------------------*/
 
+import { V220Devfile, V220DevfileCommandsItemsExecGroup } from '@devfile/api';
 import { KubeConfig } from '@kubernetes/client-node';
+import { fail } from 'assert';
 import { assert, expect } from 'chai';
 import { ChildProcess } from 'child_process';
 import * as fs from 'fs/promises';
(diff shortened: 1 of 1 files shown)


datho7561:
I get `AssertionError [ERR_ASSERTION]: Start Dev failed: I0825 11:12:19.168688   77369 starterserver.go:123] API Server started at localhost:20000/api/v1` for the test. I guess odo using stderr for some logging, which is causing the test to detect odo dev failing, which is causing the test to fail.

vrubezhny:
> I get `AssertionError [ERR_ASSERTION]: Start Dev failed: I0825 11:12:19.168688 77369 starterserver.go:123] API Server started at localhost:20000/api/v1` for the test. I guess odo using stderr for some logging, which is causing the test to detect odo dev failing, which is causing the test to fail.

This most likely means that "Start Dev" hasn't finished in 60 seconds.
When starting Dev it sends some messages like `I0825 11:12:19.168688 77369 starterserver.go:123] API Server started at localhost:20000/api/v1` to the error stream for some reason - while this is not an error at all imho, but it's reported to the error stream.
But starting  Dev takes some time and when 60 seconds timeout is reached I do report the received errors as `fail(...)` message.
So actually the reason was a timeout happened.

datho7561:
I'm running the suite a few times to see how stable it is, and it looks like it was just a one-off failure

datho7561 (approved):
Looks good to me. Thanks, Victor!

PR #3130, opened 2023-08-28
Don't show card if existing Devfile is used

### LINKED ISSUE #3129: If an existing devfile is used, don't show the card with the recommended devfile
image

This is the page after the page that informs there is an existing devfile that will be used. I assume that the devfile item being shown is the recommended one from the previous page ? This might be a bit confusing given that when a devfile exists, it's ignored.

_Originally posted by @rgrunber in https://github.com/redhat-developer/vscode-openshift-tools/issues/3104#issuecomment-1696060842_

datho7561:
Instead, show an information message indicating that the existing devfile will be used.

Fixes #3129


rgrunber (approved):
Works for me!

datho7561:
Thanks for the review!

PR #3133, opened 2023-08-29
Background colour for "Create Function" wizard should respect theme.

rgrunber:
Prior to this change on a light theme:

image

I'm actually not sure what `color` does in this instance as I wasn't able to see it shown at all, but I've also changed that to what seems like a better value.


datho7561 (approved):
Looks good. Thanks, Roland!

FYI, there's a method of handling light theme vs dark theme [built-in to Material UI that we used for create component](https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/webview/common/vscode-theme.ts#L20) that I would like to use for all webviews in the future, but as a quick fix this works well.

PR #3140, opened 2023-08-29
[test] A test suite and test for verifying a component command tree

vrubezhny:
Also:
- removed unused code from componentsView.ts
- partially restores Component.describe test suite

Issue: #2811


datho7561 (approved):
Looks good. Thanks, Victor!

Issue #3121, opened 2023-08-24
Node 16 is EOL soon


datho7561:
Node 16 goes out of maintenance on [September 11, 2023](https://github.com/nodejs/release#release-schedule).

We should update the version of node in the `.nvmrc` and in CI/CD to use Node 18.

PR #3141, opened 2023-08-30
[build] Use Node.js v18.17.1 (LTS)

vrubezhny:
Fixes: #3121


mohitsuman (approved):
lgtm

.nvmrc
```diff
@@ -1 +1 @@
-v16.17.1
+v18.17.1
```
datho7561, line 1: Just as a heads up, we can't use any nodejs 18 specific features until VS Code itself runs our extension using NodeJS 18, which will probably land in the September update (which will probably be out mid September)

vrubezhny: Changing the Node.js version here doesn't automatically leads to build (or whatever else) to immediately start using the new Node.js version.  The version that is currently configured will still be used independently of what you have written in `.nvmrc` until you apply it by `nvm use`, won't it"

rgrunber:
Works well for me. package-lock.json didnt get modified in any way and didn't get any compilation issues.

I would note that even with Node 16, if I remove the `package-lock.json` and attempt an install from scratch I get an error. But I think this is more related to a newer version of esbuild-plugin-svgr potentially causing a conflict (if/when we updated), so not related.

```
$ npm install
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: vscode-openshift-connector@1.7.0
npm ERR! Found: esbuild@0.17.19
npm ERR! node_modules/esbuild
npm ERR!   dev esbuild@"^0.17.19" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer esbuild@"^0.19.1" from esbuild-plugin-svgr@2.1.0
npm ERR! node_modules/esbuild-plugin-svgr
npm ERR!   dev esbuild-plugin-svgr@"^2.0.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
```

PR #3144, opened 2023-08-30
[tests] Use Node.js v18 images for component containers

vrubezhny:
Related to: #3121


vrubezhny:
@mohitsuman @datho7561 These are the container images I was talking about. 
Changes to use Node.js v18 images, as far as I understand, shouldn't break or make tests to work differently until a test really starts component in Dev ...  I mean run component command integration test - I don't know if any other tests can be harmed.

datho7561 (approved):
Looks good. I double checked that the referenced container image exists. I don't think that we run these devfiles in dev mode, but in the case we do in the future, this will be helpful

PR #3148, opened 2023-08-31
[tests] Fix File Watcher Utility test - close FileWatcher

vrubezhny:
Also, avoided to watch the whole temp directory because of its possibly very big size in order to not get 'too many open files error


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3149, opened 2023-08-31
Use correct folder during create component context menu button

### LINKED ISSUE #3122: Create Component From Folder always uses first open folder, regardless of which one was right clicked on
To reproduce the issue:
- Create two empty folders
- Open the folders in a workspace in VS Code
- Right click on the second folder and select "Create Component From Folder"

*Expected*: the path and name of the second folder is autofilled in the form that appears
*Actual*: the path and name of the first folder is autofilled

datho7561:
Use the selected folder when activating create component through the context menu instead of always using the first workspace folder.

Fixes #3122


msivasubramaniaan (approved):
LGTM

Issue #3123, opened 2023-08-24
Deduplicate YAML libraries


datho7561:
We depend on two YAML libraries:
- `yaml`, `1.10.2` and `2.2.2`
- `js-yaml`, `3.14.1` and `4.1.0`

It might be nice to use `js-yaml` as our direct dependency instead of `yaml`, since many of our transitive dependencies use `js-yaml`. This would reduce the number of YAML dependencies in our dependency tree by 1.

vrubezhny:
I'd suggest trying to dedupe the NPM dependency versions bu updating to the latest versions. GitHub dependabot helps to automatically create the update PRs: #3155

vrubezhny:
Switching `vscode-openshift-tools` to use `js-yaml` doesn't reduce the number of dependencies, because we still have indirect dependencies to `yaml` -  both (`js-yaml` and `yaml`) are the indirect dependencies of `vscode-openshift-connector@1.7.0`:

```
$ npm list js-yaml
vscode-openshift-connector@1.7.0 ~/projects/vscode-openshift-tools/source/vscode-openshift-tools
├─┬ @kubernetes/client-node@0.16.3
│ └── js-yaml@4.1.0 deduped
├─┬ @svgr/plugin-jsx@8.0.1
│ └─┬ @svgr/core@8.0.0
│   └─┬ cosmiconfig@8.2.0
│     └── js-yaml@4.1.0 deduped
├─┬ codecov@3.8.3
│ └── js-yaml@3.14.1
├─┬ eslint@8.48.0
│ ├─┬ @eslint/eslintrc@2.1.2
│ │ └── js-yaml@4.1.0 deduped
│ └── js-yaml@4.1.0 deduped
├─┬ istanbul@0.4.5
│ └── js-yaml@3.14.1
├── js-yaml@4.1.0
├─┬ mocha@9.2.2
│ └── js-yaml@4.1.0 deduped
├─┬ npm-check-updates@16.8.2
│ └─┬ rc-config-loader@4.1.2
│   └── js-yaml@4.1.0 deduped
├─┬ tslint@5.20.1
│ └── js-yaml@3.14.1
└─┬ vscode-extension-tester@5.5.1
  └── js-yaml@4.1.0 deduped

$ npm list yaml
vscode-openshift-connector@1.7.0 ~/projects/vscode-openshift-tools/source/vscode-openshift-tools
├─┬ @mui/lab@5.0.0-alpha.124
[...]

PR #3180, opened 2023-09-02
Deduplicate YAML libraries #3123

vrubezhny:
Use JS-YAML instead of YAML

 Fixes: #3123


datho7561:
Found two more uses of `'yaml'`:
```
❯ rg "'yaml'"
src/serverlessFunction/commands.ts
8:import { parse } from 'yaml';

src/serverlessFunction/functionImpl.ts
10:import { stringify } from 'yaml';
```

datho7561 (approved):
Looks good. Thanks, Victor!

PR #3188, opened 2023-09-05
Remove `utf-8-validate`

datho7561:

Supercedes #3178


vrubezhny (approved):
Looks good to me

PR #3198, opened 2023-09-05
Remove 'css-loader' dependency as unused

vrubezhny:

Supercedes #3182


datho7561:
Thanks, Victor!

PR #3199, opened 2023-09-05
Remove ts-loader as a direct dependency

vrubezhny:
Supercedes #3173


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3201, opened 2023-09-06
Remove 'style-loader' dependency as unused

vrubezhny:


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3212, opened 2023-09-06
[build] Try limiting the supported Node.js version range

vrubezhny:
Try limiting Node.js to [v16, v18] (inclusively) to use only the versions supported by VSCode

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -118,6 +118,8 @@
 				"vscode-extension-tester": "^5.2.0"
 			},
 			"engines": {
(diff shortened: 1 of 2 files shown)


vrubezhny:
@datho7561 I'd like to try merging these limitations to see if `@dependabot` will react and start suggesting to update modules not to just `latest`, but to `latest compatible` versions. Could you please review if the limits in `engines` and in `@types/node` won't break anything?

datho7561 (approved):
Looks good! Doesn't prevent building testing on Node 18, although there is a warning

PR #3213, opened 2023-09-06
Remove 'npm-check-updates' dependency as unused

vrubezhny:
Supercedes #3210


datho7561:
Also remove the `update-deps` scripts in the `package.json`, which references `ncu`, the binary that is installed with `npm-check-updates`

datho7561 (approved):
Looks good. Thanks, Victor!

datho7561:
As a result of updating vscode-extension-tester, one of the type assertions that we perform in the tests is redundant. This is unrelated to this PR, but we should fix it soon so that the tests can run. I'll merge this PR for now.

PR #3214, opened 2023-09-06
Upversion to 1.8.0

datho7561:


datho7561:
see https://github.com/redhat-developer/vscode-openshift-tools/pull/3216

PR #3223, opened 2023-09-07
Fix build errors after db update got 13.0.0

vrubezhny:


vrubezhny:
Due to API changes in `got` the following test suite doesn't work as expected  anymore and needs to be changed:
```
  Download Util
      1) reports download progress
  Suite duration: 0.028 s, Tests: 1
```

PR #3237, opened 2023-09-08
Update Node.js/Npm engines to v18.x / v9.x

vrubezhny:
As VSCode 1.182.0 (including support for Node 18.x) is released we can use  Node.js v18 / NPM v9


datho7561:
If you perform `npm i`, it also modifies the `package-lock.json`. It would be nice to also commit those changes in this PR.

vrubezhny:
> If you perform `npm i`, it also modifies the `package-lock.json`. It would be nice to also commit those changes in this PR.

It;s better to do it right before merge in order to not override the other changes/not to have conflicts).
But, ye. I'll do this

datho7561 (approved):
Looks good. Thanks, Victor!

Issue #3145, opened 2023-08-30
Tree item in the Serverless Functions view keeps name of the latest used project


lgrossma:
Tree item in the Serverless Functions view changes its name depending on the name of the project that is currently used. After deleting all projects on cluster or loggin off the cluster, the tree item in Serverless functions view keeps the name of the latest used project.
![Screenshot 2023-08-30 at 14 58 58](https://github.com/redhat-developer/vscode-openshift-tools/assets/86789336/5a8d91dc-88ee-4c78-bbe1-a61d969f3298)

msivasubramaniaan:
@lgrossma 
Yes it is always pointed to latest namespace based on `.kube` config file. When user switches to different namespace then tree automatically get refreshed.

lgrossma:
Sounds as a correct behaviour when user switches namespaces, but I was pointing to a case when user loggs off or deletes all namespaces - in this case, shouldn't the tree item be set to it's default value?

msivasubramaniaan:
As part of serverless function workflow, There are three types:

1. Local
2. Cluster
3. Local/Cluster

We have to provide an option to create function even though there were no cluster/namespaces. So for incorporating the above local workflow we are showing the latest namespace. And user can create a function under on it.

lgrossma:
I understand that workflow. It just does not make sense to me to use the latest namespace, when there is none. For me as a user, it kinda seemed as a bug that after deleting all projects or logging out the tree item label did not switch back to it's default value..

datho7561:
@rgrunber mentioned that we should be able to create a serverless function without providing a namespace, so we shouldn't use the last available namespace to create the serverless function.

Do we need to show the current namespace in the Serverless Functions view at all? We are already showing it in the application explorer view.

msivasubramaniaan:
> @rgrunber mentioned that we should be able to create a serverless function without providing a namespace, so we shouldn't use the last available namespace to create the serverless function.
> 
> Do we need to show the current namespace in the Serverless Functions view at all? We are already showing it in the application explorer view.

I think we can removed the current namespace from Serverless Functions view.

PR #3241, opened 2023-09-08
Removed current namespace from serverless function tree view

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3145

### DIFF
--- a/src/serverlessFunction/view.ts
+++ b/src/serverlessFunction/view.ts
@@ -169,16 +169,7 @@ export class ServerlessFunctionView implements TreeDataProvider<ExplorerItem>, D
     async getChildren(element?: ExplorerItem): Promise<ExplorerItem[]> {
         let result: ExplorerItem[] = [];
         if (!element) {
-            result = [{
-                kind: 'project',
-                metadata: {
-                    name: this.kubeContext?.namespace,
-                },
-            } as KubernetesObject];
-            this.setCurrentNameSpace(this.kubeContext?.namespace);
-        } else if ('kind' in element) {
-            if (element.kind === 'project') {
-                result = [...await this.serverlessFunction.getLocalFunctions()]
+            result = [...await this.serverlessFunction.getLocalFunctions()]
                 if (result.length === 0) {
                     const functionNode: FunctionObject = {
                         name: 'No Available Functions',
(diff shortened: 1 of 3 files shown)


src/serverlessFunction/view.ts
```diff
@@ -169,24 +169,15 @@ export class ServerlessFunctionView implements TreeDataProvider<ExplorerItem>, D
...
+            result = [...await this.serverlessFunction.getLocalFunctions()]
                 if (result.length === 0) {
                     const functionNode: FunctionObject = {
                         name: 'No Available Functions',
                         context: FunctionStatus.NONE
                     }
                     result = [functionNode]
                 }
-            }
+            this.setCurrentNameSpace(this.kubeContext?.namespace);
```
datho7561, line 180: Is this still needed? I think you can get rid of this as well as the other code that watches for namespace changes, such as the file watcher on the `~/.kube/config`, since you no longer need to display the up to date namespace.

datho7561 (changes requested):

src/serverlessFunction/view.ts
msivasubramaniaan: > Is this still needed? I think you can get rid of this as well as the other code that watches for namespace changes, such as the file watcher on the `~/.kube/config`, since you no longer need to display the up to date namespace.

The latest namespace needed on deploy command. There was a parameter we need to pass

datho7561: As a part of the "remove cycles" PR I made the following change: https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/serverlessFunction/functions.ts#L242

msivasubramaniaan: Removed :)

datho7561:
There is also this file watcher, which I think is only used to track the current namespace: https://github.com/redhat-developer/vscode-openshift-tools/pull/3241/files#diff-7b252f53014e39d3101cc51d5a916d9da0cddcc8467c63aa901db00a85ee2c61R68

datho7561 (approved):
Looks good. Thanks, Muthu!

datho7561:
(waiting for CI)

PR #3242, opened 2023-09-08
Add `./src/tools.json` to distribution

### LINKED ISSUE #3154: Error when opening "Create OpenShift Local Cluster"
```
ENOENT: no such file or directory, open '/home/davthomp/.vscode/extensions/redhat.vscode-openshift-connector-1.7.0/src/tools.json'
```

datho7561:
To verify this change,
- build and run from the .vsix
- remove `crc` from PATH if you have it (or rename it)
- log out of crc
- select "+" in the application explorer in the OpenShift sidebar
- select OpenShift Local in the wizard that appears
  - There should be no error popup
- Click reset then download crc
  - it should take you inside the folder for the latest version instead of the parent folder with all the versions listed

Fixes #3154

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -8,6 +8,7 @@ out/tools-cache/**
 out/**/*.map
 out/src-orig/**
 src/**
+!src/tools.json
 doc/**
 .gitignore
 .github


rgrunber:
This is a fairly recent issue right ? I just checked `openshift-toolkit-1.7.0-2-linux-x64.vsix` and it contains `extension/out/src/tools.json`.

msivasubramaniaan (approved):
LGTM

PR #3250, opened 2023-09-11
moved serverless function walkthrough gifs

msivasubramaniaan:


rgrunber (approved):
Definitely improves the responsiveness when opening those pages

Screenshot from 2023-09-05 08-44-13

Only current complaint would be the gifs are a bit difficult to read when embedded in the page. We can always fix this later.

PR #3269, opened 2023-09-14
[build] Install extension dependencies defined in package.json

vrubezhny:
Make the list of extensions to be installed to be read from `extensionDependencies` key of `package.json`  instead of duplicating the same list in a constant.


datho7561 (approved):
Looks good! Thanks, Victor!

PR #3270, opened 2023-09-14
Extension wouldn't start due to missing shelljs

datho7561:
Move the shelljs dependency from development to production, since we use it in the module to download tool binaries.


vrubezhny (approved):
Looks good to me

rgrunber:
Good catch. I can reproduce :

```
2023-09-14 13:15:54.086 [error] Activating extension redhat.vscode-openshift-connector failed due to an error:
2023-09-14 13:15:54.086 [error] Error: Cannot find module 'shelljs'
```

rgrunber (approved):

vrubezhny:
Yes, and it looks like we couldn't catch it in our tests because we're not installing from  VSIX during the tests, but:
```
2023-09-14T10:39:26.3041413Z Loading development extension at /home/runner/work/vscode-openshift-tools/vscode-openshift-tools
 ```
 ... which is built using whole set of dependencies (`dependencies` + `devDependencies`)

PR #3310, opened 2023-09-21
Do not autoclose terminals on exit

### LINKED ISSUE #3309: Avoid autoclosing tabs after the command exits in OpenShift Terminal
In the OpenShift Terminal PR, I auto close some tabs, like `odo dev`, after they exit.

Victor pointed out this isn't a good idea, since the user may want to copy text from these commands after they exit.

In order to reduce the number of tabs that are open, we can investigate other strategies like reusing tabs after the command exits.

datho7561:
Fixes #3309

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -309,7 +309,6 @@ export class Component extends OpenShiftItem {
                 `odo dev: ${component.component.devfileData.devfile.metadata.name}`,
                 component.contextPath,
                 process.env,
-                true,
                 {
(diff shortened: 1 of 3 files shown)


datho7561:
Don't bother reviewing yet, I think this is broken

vrubezhny (approved):
Looks good to me. 
Terminal tab stays open and shows the output for a component command  and/or a  serverless function after its execution is complete. So user can continue  using the information logged out to the terminal tab.

PR #3312, opened 2023-09-22
Refactor code that interacts with `odo`

### LINKED ISSUE #3078: Refactor `odo` related code in anticipation of redhat-developer/odo#6990
We have two files which handle invoking the `odo` cli:
- `./src/odo.ts`
- `./src/odo/command.ts`

However, there are a few problems with this setup:
- Some of the functions in these files interact with `oc` instead of `odo`
- `./src/odo.ts` is highly coupled to `./src/odo/command.ts`
  - i.e. many of the functions in `./src/odo/command.ts` are only called directly in `./src/odo.ts`
  - The purpose of `./src/odo/command.ts` is to hide the name of the `odo` command that is called to perform a given action. However, this assumes that `./src/odo.ts` will implement it's functionality by calling an `odo` command, which won't be true once we use redhat-developer/odo#6990

It's important to refactor these files before we start integrating redhat-developer/odo#6990. Here are some things that we can do:
- Separate the `odo` and `oc` functionality into two files. I've started this process by moving the `helm` commands out of `./src/odo.ts` into their own module.
- Get rid of `./src/odo/command.ts`. Add the functions that return the `CommandText` as private functions in `./src/odo.ts` or the file that is created for the `oc` commands
- Make sure that there are not any duplicate instances of the `oc` commands
  - It seems like the code that extends the Kubernetes tree view has it's own wrapper functions for the `oc` commands that it invokes
- Rewrite the unit tests taking the above into account so that they pass
  - This is tricky, since we sometimes directly stub the function that invokes the command instead of the wrapper function that invokes the command and processes the output

datho7561:
- Split `odo.ts` up, so that commands that interact with `odo` go into `odoWrapper.ts` and commands that interact with `oc` go into `ocWrapper.ts`
- Remove most methods from `./src/odo/command.ts`, and instead create the `CommandText` objects directly in the corresponding method in `odoWrapper.ts`
- Adapt the integration tests to the change, and add a few more integration tests
- Delete some dead code

Closes #3078


vrubezhny (approved):
Looks good to me

Issue #3316, opened 2023-09-24
Invoke a function using "OpenShift Toolkit" extension doesn't work


danieloh30:
After I created a new function using the OpenShift Toolkit extension on VS Code, I tried to invoke the function using the extension GUI. But it failed with the following error:

Invoking 'cloudevent' function at http://localhost:8080/
Invoking 'cloudevent' function using 'cloud events' format
Error: format 'cloud events' not supported.

<img width="1107" alt="image" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/14066977/48adfeb9-88b0-4276-9842-348c09894a83">

My func.yaml looks like

specVersion: 0.35.0
name: quarkus-awesome
runtime: quarkus
registry: quay.io/danieloh30
image: quay.io/danieloh/quarkus-awesome:latest
created: 2023-09-24T01:53:14.752Z
invoke: cloudevent
build:
  builder: pack
  buildEnvs:
  - name: BP_NATIVE_IMAGE
    value: "false"
  - name: BP_MAVEN_BUILT_ARTIFACT
    value: func.yaml target/quarkus-app/lib/ target/quarkus-app/*.jar target/quarkus-app/app/
      target/quarkus-app/quarkus/
  - name: BP_MAVEN_BUILD_ARGUMENTS
    value: package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=fast-jar
  - name: MAVEN_S2I_ARTIFACT_DIRS
    value: target/quarkus-app
  - name: S2I_SOURCE_DEPLOYMENTS_FILTER
    value: lib quarkus-run.jar app quarkus
  pvcSize: 256Mi

The Quarkus app runs well.
Adding 137 container CA certificates to JVM truststore
Picked up JAVA_TOOL_OPTIONS: -Djava.security.properties=/layers/paketo-buildpacks_bellsoft-liberica/java-security-properties/java-security.properties -XX:+ExitOnOutOfMemoryError -XX:ActiveProcessorCount=6 -XX:MaxDirectMemorySize=10M -Xmx10302293K -XX:MaxMetaspaceSize=71230K -XX:ReservedCodeCacheSize=240M -Xss1M -XX:+UnlockDiagnosticVMOptions -XX:NativeMemoryTracking=summary -XX:+PrintNMTStatistics
__  ____  __  _____   ___  __ ____  ______ 
[...]

PR #3319, opened 2023-09-25
getting serverless function invoke templates from CLI

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3316

**Post Fix:**
image


datho7561:
Something I noticed is that only the first word of the "Data" gets sent to the serverless function. eg. in the response, you can see `Hello`, but not `World`. This is a separate issue that I'll file.

datho7561 (approved):

datho7561:
Thanks, Muthu!

PR #3343, opened 2023-09-26
fix: downgrade make-fetch-happen to 12.0.0 to fix sandbox detection

### LINKED ISSUE #3342: Sandbox page in `Add OpenShift Cluster` editor stays in 'Detecting' status forever
Current master version of 'Add OpenShift Cluster' editor stays in detection status forever. The reason is the error in agent-base module (see screenshot below). Apparently during extension packaging agent-base module used in make-fetch-happen@13.0.0 is not included, because vscode has its own verison included and it does not work well with make-fetch-happen@13.0.0. 

image

The fix is to downgrade make-fetch-happen to 12.0.0

dgolovin:
Fix #3342.

Signed-off-by: Denis Golovin <dgolovin@redhat.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
 				"js-yaml": "^4.1.0",
 				"json-schema": "^0.4.0",
 				"lodash": "^4.17.21",
-				"make-fetch-happen": "^13.0.0",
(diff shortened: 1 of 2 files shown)


vrubezhny (approved):
Looks good to me.

PR #3360, opened 2023-09-27
Terminal: keybinding and context menu for copy & select all

### LINKED ISSUE #3137: "Select All" functionality for OpenShift Terminal
See [this comment from Victor](https://github.com/redhat-developer/vscode-openshift-tools/pull/3065#issuecomment-1693730583).

There are context menu items for cut, copy and paste, but not "Select All".

I don't know if it's feasible to add another context menu, but it would be nice to have some method of doing "Select All" even if it's not in a context menu.

### LINKED ISSUE #3353: Copying terminal text doesn't work on Linux
The expected key combo for copying text (`Ctrl+Shift+C`) doesn't work (at least on Linux).

The context menu option to copy text also doesn't seem to work.

datho7561:
Adds the following keybindings to the OpenShift terminal:
- `Ctrl+Shift+C`: copy selection
- `Ctrl+Shift+A`: select all text

Hide the existing broken context menu (provided by VS Code) and replace it with one with "Select All" and "Copy" items that actually work. This context menu imitates the style of VS Code's context menu.

Closes #3353, closes #3137

### DIFF
--- a/src/webview/openshift-terminal/app/terminalInstance.tsx
+++ b/src/webview/openshift-terminal/app/terminalInstance.tsx
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See LICENSE file in the project root for license information.
  *-----------------------------------------------------------------------------------------------*/
 
-import { Box } from '@mui/material';
+import { Box, Button, Paper, Stack, Typography } from '@mui/material';
 import React from 'react';
 import { VSCodeMessage } from './vscodeMessage';
 import { Terminal, ITheme } from 'xterm';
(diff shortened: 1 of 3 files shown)


src/webview/openshift-terminal/app/terminalInstance.tsx
```diff
@@ -35,9 +132,32 @@ export const TerminalInstance = (props: {
         newTerm.loadAddon(new WebLinksAddon());
         newTerm.loadAddon(new WebglAddon());
         newTerm.loadAddon(fitAddon);
+        newTerm.attachCustomKeyEventHandler((keyboardEvent: KeyboardEvent) => {
+            // Copy/Paste/Select All keybinding handlers
+            if (keyboardEvent.shiftKey && keyboardEvent.ctrlKey) {
+                // https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values
+                if (keyboardEvent.code === 'KeyC' && term.hasSelection) {
+                    // Ctrl+Shift+C copies
+                    void navigator.clipboard.writeText(term.getSelection());
```
vrubezhny, line 141: The problem here is that the KeyboardEvent processing is not stopped after the `Ctrl-Shift-C`/`Ctrl-Shift-A` hot keys are processed here. For example, on my Fedora,  pressing `Ctrl-Shift-C`  leads into copying the selected text into Clipboard (this handler) plus opening a System Terminal.
 
Everything else works fine,

vrubezhny (approved):
Looks good to me,

Please see the comment inline

src/webview/openshift-terminal/app/terminalInstance.tsx
datho7561: I'm not too sure what I can do to prevent this. I tried adding `preventDefault()` in the key event handler, but that didn't fix it.

datho7561: Alright I pushed the fix, let me know if it works for you

vrubezhny (approved):
Look good and perfectly works!

Thanks!

datho7561:
Yay! Thanks for the review, Victor!

Issue #3339, opened 2023-09-26
Use VSCode SecretStorage API instead of 'keytar'


vrubezhny:
Due to `keytar` [is officially archived and no longer be supported](https://github.com/microsoft/vscode-discussions/discussions/662) it's suggested to start using [VSCode SecretStorage API](https://code.visualstudio.com/api/references/vscode-api#SecretStorage) instead

PR #3368, opened 2023-09-28
Use VSCode SecretStorage API instead of 'keytar' #3339

vrubezhny:
Fixes: #3339


datho7561 (approved):
Looks good, and works as expected. Thanks, Victor!

PR #3370, opened 2023-09-28
If a command is rerun after exiting, close the old tab

### LINKED ISSUE #3138: Investigate reusing terminals to avoid too many tabs open
See [Victor's comments](https://github.com/redhat-developer/vscode-openshift-tools/pull/3065#issuecomment-1693730583)

> A new OS Terminal tab is opened for every run of the same Component Command (personally I think we should reuse tabs at least if we run the same command multiple times and probably to have a possibility to clear the contents of a terminal tab)

It would be nice to reuse the same tab for commands that immediately exit.

datho7561:
When you start a new command in the terminal, check if there is a tab open where that command was run and has exited. If there is, close that tab before opening an new tab to run the rerun of the command. Not quite the same as recycling the tab, but it has a much simpler implementation.

Closes #3138

### DIFF
--- a/src/webview/openshift-terminal/openShiftTerminal.ts
+++ b/src/webview/openshift-terminal/openShiftTerminal.ts
@@ -74,6 +74,7 @@ class OpenShiftTerminal {
     private _file: string;
     private _args: string | string[];
     private _options;
+    private _name: string;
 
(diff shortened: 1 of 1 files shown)


vrubezhny (approved):
Looks good to me.

Thanks!

Issue #3040, opened 2023-07-19
Display Progress Information during Import Git Clone Process


mohitsuman:
**Issue Description:**
When cloning a Git repository , the process lacks a progress indicator, leaving users unaware of the current status and completion time. This can be frustrating, especially when cloning large repositories, as there's no visibility into the progress. This issue proposes the addition of a progress information display during the Git clone process to enhance the user experience and provide valuable feedback.

**Acceptance Criteria:**
- [ ] Implement a progress indicator during the Git clone process to show the percentage of completion 
- [ ] The progress information should be displayed in the progress bar MUI
- [ ] If the clone process completes successfully, the progress indicator should be replaced with a success message indicating the completion.
- [ ] If the clone process encounters an error, the progress indicator should be replaced with an appropriate error message, providing details about the issue.
- [ ] Thoroughly test the progress indicator under different scenarios, including large repository clones, slow network connections, and error-handling scenarios.
- [ ] Verify that the feature works seamlessly across different operating systems

**Additional Notes:**
This enhancement would significantly improve the user experience for developers who perfoms Import From Git repositories. The progress information display will provide users with valuable feedback on the ongoing process and alleviate frustration caused by the lack of visibility into the cloning progress. The feature should be thoroughly tested and reviewed to ensure its stability and compatibility with various Git repository configurations and network conditions.

PR #3371, opened 2023-09-28
added percentage of completion on progress bar

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3040

**Post Fix:**
complete-percentage

### DIFF
--- a/src/webview/create-component/createComponentLoader.ts
+++ b/src/webview/create-component/createComponentLoader.ts
@@ -425,6 +425,12 @@ export default class CreateComponentLoader {
         let analyzeRes: AnalyzeResponse[] = [];
         let compDescriptions: ComponentTypeDescription[] = [];
         try {
+            void CreateComponentLoader.panel.webview.postMessage({
+                action: 'getRecommendedDevfile',
+                data: {
+                    value: 60
+                }
+            });
             analyzeRes = await OdoImpl.Instance.analyze(uri.fsPath);
             compDescriptions = getCompDescription(analyzeRes);
         } catch (error) {
@@ -464,6 +470,12 @@ export default class CreateComponentLoader {
                 }
             }
         } finally {
+            void CreateComponentLoader.panel.webview.postMessage({
+                action: 'getRecommendedDevfile',
+                data: {
+                    value: 90
+                }
+            });
             const devfileRegistry: DevfileRegistry[] = getDevfileRegistries();
             const allDevfiles: Devfile[] = devfileRegistry.flatMap((registry) => registry.devfiles);
             const devfile: Devfile =
@@ -530,6 +542,12 @@ function clone(url: string, location: string, branch?: string): Promise<ClonePro
     const git = gitExtension.getAPI(1).git.path;
     let command = `${git} clone ${url} ${location}`;
     command = branch ? `${command} --branch ${branch}` : command;
+    void CreateComponentLoader.panel.webview.postMessage({
+        action: 'cloneExecution',
+        data: {
+            value: 25
+        }
+    });
     // run 'git clone url location' as external process and return location
     return new Promise((resolve, reject) =>
         cp.exec(command, (error: cp.ExecException) => {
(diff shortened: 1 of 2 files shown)


src/webview/create-component/pages/fromExistingGitRepo.tsx
```diff
@@ -229,7 +250,27 @@ export function FromExistingGitRepo({ setCurrentView }) {
                                                 spacing={2}
                                                 alignItems="center"
                                             >
-                                                <CircularProgress />
+                                                <Box sx={{ position: 'relative', display: 'inline-flex' }}>
+                                                    <CircularProgress />
```
datho7561, line 254: It looks like there is a pre-built component for this: https://mui.com/material-ui/react-progress/#circular-with-label, I think it would be better to use the pre-built component

src/webview/create-component/createComponentLoader.ts
```diff
@@ -530,6 +539,9 @@ function clone(url: string, location: string, branch?: string): Promise<ClonePro
     const git = gitExtension.getAPI(1).git.path;
     let command = `${git} clone ${url} ${location}`;
     command = branch ? `${command} --branch ${branch}` : command;
```
datho7561, line 541: Ask Roland about this, there should be a way to get json output from git clone so that you can get precise progress numbers. If you run git using `spawn` from `node:child_process`, then you should be able to attach a callback that runs whenever something is printed to the terminal, and you can use this to send a message to the webview that updates the progress.

src/webview/create-component/pages/fromExistingGitRepo.tsx
```diff
@@ -77,12 +79,17 @@ export function FromExistingGitRepo({ setCurrentView }) {
...
+                        ...prevState,
+                        completionValue: 100,
+                    }));
                     setRecommendedDevfile((prevState) => ({
                         ...prevState,
                         devfile: message.data.devfile,
                     }));
                     setRecommendedDevfile((prevState) => ({
                         ...prevState,
+                        completionValue: 100,
```
datho7561, line 92: You don't need to set this again here, it's already be set on line 84

datho7561 (changes requested):

src/webview/create-component/createComponentLoader.ts
datho7561: The progress is reported onto `stderr` if you pass the `--progress` option to `git`, I asked Roland and it seems there is not a nice way to get JSON.

src/webview/create-component/pages/fromExistingGitRepo.tsx
msivasubramaniaan: I think there were no pre-build component, internally they were used [CircularProgress](https://mui.com/material-ui/react-progress/#circular). I haven't find any specific imports as well

src/webview/create-component/createComponentLoader.ts
msivasubramaniaan: Here the percentage of completion not only for clone, It was from clone to analyze flow. So I segregated the completion percentage on each level.

src/webview/create-component/pages/fromExistingGitRepo.tsx
datho7561: You're right. Wow, the MUI documentation is pretty misleading.

src/webview/create-component/createComponentLoader.ts
datho7561: Okay, this makes sense to me.

src/webview/create-component/pages/fromExistingGitRepo.tsx
```diff
@@ -229,7 +249,27 @@ export function FromExistingGitRepo({ setCurrentView }) {
                                                 spacing={2}
                                                 alignItems="center"
                                             >
-                                                <CircularProgress />
+                                                <Box sx={{ position: 'relative', display: 'inline-flex' }}>
+                                                    <CircularProgress />
```
datho7561, line 253: If you use the determinant variant and pass the value then it will show another visual indicator of progress: https://mui.com/material-ui/react-progress/#circular-determinate

datho7561 (approved):
Works great as it currently is, but I think there is a small visual improvement you can make. Feel free to merge after making the change.

Issue #3318, opened 2023-09-25
Components tree view not get refreshed


msivasubramaniaan:
Step to re-produce:

- Add a component folder into workspace
- The component will be shown into the tree view
- Delete `devfile.yaml` or `.odo` folder

Error: The component still showing in the view
Assert: The component should not show

vrubezhny:
@msivasubramaniaan We do not need to watch for `.odo` removal here. Existing  `devfile.yaml` is enough for a component to be functional and `.odo` will be created by `odo` for example if you execute `Start Dev` on a component.

A PR to fix  is added here: #3377

PR #3377, opened 2023-09-29
Components tree view not get refreshed #3318

vrubezhny:
Fixes: #3318

### DIFF
--- a/src/util/workspace.ts
+++ b/src/util/workspace.ts
@@ -117,7 +117,8 @@ async function updateDevfileContext(_: unknown) {
         }),
     );
     const foldersArray = devfileFolders.map(folder => folder.name);
-    void commands.executeCommand('setContext', 'ext.folderContainsDevfile', foldersArray);
+    void commands.executeCommand('setContext', 'ext.folderContainsDevfile', foldersArray)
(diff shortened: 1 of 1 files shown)


datho7561 (changes requested):
This doesn't seem to work for me. I put a breakpoint on the added line, and it's never hit when I delete or move the Devfile.

vrubezhny:
Hm...
screencast-component-view-refresh-on-devfile-delete

vrubezhny:
@datho7561 I've updated the PR, could you please  give it another try?

datho7561:
I think that what I was actually running into was that any newly added folders don't get the listener installed. I'll do a bit more testing.

datho7561:
I can't replicate what happened before.

datho7561 (approved):
Works for me. Thanks, Victor!

vrubezhny:
> I think that what I was actually running into was that any newly added folders don't get the listener installed. I'll do a bit more testing.

This purely can be a reason for not working - but this is another issue - with this PR the removal of a component  `devfile.yaml` was fixed.

Reacting on a  new component folder is added (and a `devfile.yaml` placed into that folder) - is a different story.

PR #3378, opened 2023-09-29
Context menu item to clear terminal

### LINKED ISSUE #3135: Button to clear contents of terminal in OpenShift Terminal view
See [this comment from Victor](https://github.com/redhat-developer/vscode-openshift-tools/pull/3065#issuecomment-1693730583)

datho7561:
Closes #3135

### DIFF
--- a/src/webview/openshift-terminal/app/terminalInstance.tsx
+++ b/src/webview/openshift-terminal/app/terminalInstance.tsx
@@ -14,9 +14,10 @@ import 'xterm/css/xterm.css';
 import '../../common/scrollbar.scss';
 
 /**
- * Clone of VS Code's context menu with "Copy" and "Select All" items.
+ * Clone of VS Code's context menu with "Copy", "Select All", and "Clear" items.
  */
 const TerminalContextMenu = (props: {
+    onClearHandler: React.MouseEventHandler<HTMLButtonElement>;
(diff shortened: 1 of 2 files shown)


vrubezhny:
Perfectly works... It's just that the only one thing confuses me... See the different context menus we have (in Dark Theme):

context-menues

datho7561:
I have the `"window.titleBarStyle": "custom"` setting enabled. This setting makes all the context menus look like the one I added.

I think we might need to detect if this setting is enabled and switch between the two styles of context menu. The only issue is that I think the appearance varies between operating systems when `"window.titleBarStyle": "custom"` is disabled.

vrubezhny:
Yes, I had `"window.titleBarStyle" === "native"`... With `"custom"` all the other menus are looking  "dark" exactly as the one from the OS Terminal.

I don't know, but probably it's better if all the application menus look the same way -  color scheme/fonts/etc.

vrubezhny (approved):
Looks good to me.

The context menu coloring issue - a color scheme on OS Terminal context menu looks different compared to other app context menus when Dark Theme and  `"window.titleBarStyle" === "native"` preference are set - is not related to the `Clear` OS Terminal context menu action itself. We should probably open a different issue for this or leave as is.

Issue #3386, opened 2023-10-02
Component create from Git fails if a devfile exists in git repo


vrubezhny:
`Create Component` view cannot be opened when creating a new component from a git, using  https://github.com/openshift-evangelists/Wild-West-Frontend app repo:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'port')
    at SetNameAndFolder (index.js:131409:73)
    at renderWithHooks (index.js:13229:28)
    at mountIndeterminateComponent (index.js:15405:23)
    at beginWork (index.js:16234:24)
    at HTMLUnknownElement.callCallback2 (index.js:5636:24)
    at Object.invokeGuardedCallbackDev (index.js:5661:26)
    at invokeGuardedCallback (index.js:5695:41)
    at beginWork$1 (index.js:19395:17)
    at performUnitOfWork (index.js:18621:23)
    at workLoopSync (index.js:18572:15)
index.js:16871 The above error occurred in the <SetNameAndFolder> component:

    at SetNameAndFolder (https://file+.vscode-resource.vscode-cdn.net/home/jeremy/projects/vscode-openshift-tools/source/vscode-openshift-tools/out/createComponentViewer/index.js:131408:71)
    at FromExistingGitRepo (https://file+.vscode-resource.vscode-cdn.net/home/jeremy/projects/vscode-openshift-tools/source/vscode-openshift-tools/out/createComponentViewer/index.js:131795:34)
    at div
```

Which maps to line 39 at https://github.com/redhat-developer/vscode-openshift-tools/blob/6821c8c736d7116075d989ab6de5997e4d05ef02/src/webview/common/setNameAndFolder.tsx#L39

PR #3387, opened 2023-10-02
Component create from Git fails if a devfile exists in git repo #3386

vrubezhny:
Fixes: #3386

### DIFF
--- a/src/webview/create-component/createComponentLoader.ts
+++ b/src/webview/create-component/createComponentLoader.ts
@@ -29,6 +29,7 @@ import {
 } from '../common-ext/createComponentHelpers';
 import { loadWebviewHtml, validateGitURL } from '../common-ext/utils';
 import { Devfile, DevfileRegistry, TemplateProjectIdentifier } from '../common/devfile';
+import { DevfileV1 } from '../../util/devfileV1Type';
 
 interface CloneProcess {
     status: boolean;
@@ -281,11 +282,19 @@ export default class CreateComponentLoader {
                         action: 'cloneFailed',
                     });
                 } else {
+                    const isGirDevfileExists = await isDevfileExists(tmpFolder);
                     void CreateComponentLoader.panel.webview.postMessage({
                         action: 'devfileExists',
-                        data: await isDevfileExists(tmpFolder),
+                        data: isGirDevfileExists,
                     });
-                    void CreateComponentLoader.getRecommendedDevfile(tmpFolder);
+
+                    if (isGirDevfileExists) {
+                        // Use the Devfile existing in Gir-repo
+                        void CreateComponentLoader.getExistingDevfile(tmpFolder);
+                    } else {
+                        // Use recommended Devfile
+                        void CreateComponentLoader.getRecommendedDevfile(tmpFolder);
+                    }
                 }
                 break;
             }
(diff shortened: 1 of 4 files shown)


src/webview/create-component/pages/fromExistingGitRepo.tsx
```diff
@@ -453,7 +453,7 @@ export function FromExistingGitRepo({ setCurrentView }) {
                         setCurrentPage('fromGitRepo');
                     }}
                     createComponent={createComponentFromGitRepo}
-                    devfile={recommendedDevfile.isDevfileExistsInRepo ? undefined : selectedDevfile ? selectedDevfile : recommendedDevfile.devfile}
+                    devfile={recommendedDevfile.isDevfileExistsInRepo ? recommendedDevfile.devfile : selectedDevfile ? selectedDevfile : recommendedDevfile.devfile}
```
datho7561, line 456: It would be nice to pass a boolean indicating whether we are working with an existing Devfile, and disabling the component name and port fields in this case, since setting the component name and port will do nothing when there is an existing Devfile, which is unintuitive.

datho7561 (changes requested):

src/webview/create-component/pages/fromExistingGitRepo.tsx
vrubezhny: Acknowledged

src/webview/create-component/createComponentLoader.ts
```diff
@@ -424,6 +438,67 @@ export default class CreateComponentLoader {
...
+                rawDevfile = JSYAML.load(file.toString());
+            }
+
+            void CreateComponentLoader.panel.webview.postMessage({
+                action: 'getRecommendedDevfile'
+            });
+
+            const devfile: Devfile = {
+                description: rawDevfile.metadata.description,
+                registryName: devFileYamlPath,
```
datho7561, line 483: I think it's better to leave this blank. When importing from git, it shows this folder on the "Set Name, Port, and Folder" step. This folder is always some temp folder that we created, so this information isn't helpful to the user.

ItShowsTheTempFolder

datho7561 (changes requested):
In addition to the comment I added, it would be nice to populate the "Component Name" field with the component name from the existing Devfile instead of the one that we generate.

src/webview/create-component/createComponentLoader.ts
vrubezhny: That's my bad -  the goal was to show a Git-repo URL, but I forgot to change this line. But probably you're right - we don't need to show anything here.

vrubezhny:
> In addition to the comment I added, it would be nice to populate the "Component Name" field with the component name from the existing Devfile instead of the one that we generate.

@datho7561 Do you mean we should get the devfile's `metadata.name` in any case:  either this is a devfile from a Git-repo or a recommended devfile?

datho7561:
I think in the case of Wild-West-Frontend (i.e. when importing a git repo or local codebase with an existing Devfile), it makes sense to populate the field with the name from the existing Devfile, since if its already set, the user probably doesn't want to change it, even if it's different from the repo name.

vrubezhny:
> I think in the case of Wild-West-Frontend (i.e. when importing a git repo or local codebase with an existing Devfile), it makes sense to populate the field with the name from the existing Devfile, since if its already set, the user probably doesn't want to change it, even if it's different from the repo name.

@datho7561 Resonably. I've updated the form so the initial component name is taken from a  devfile provided

datho7561 (approved):
Looks good. Thanks, Victor!

Issue #3294, opened 2023-09-20
"Create Component from Git" doesn't sanitize the suggested component name


datho7561:
eg. create a component from `https://github.com/eclipse/lemminx.git` (make sure to include the `.git` at the end)

The suggested component name doesn't get sanitized like it does for the fix introduced in https://github.com/redhat-developer/vscode-openshift-tools/issues/3150

eg. it suggests `lemminx.git` instead of the expected `lemminx-git`

PR #3396, opened 2023-10-03
"Create Component from Git" doesn't sanitize the suggested component …

vrubezhny:
…name #3294

Fixes: #3294

### DIFF
--- a/src/webview/common/sanitize.ts
+++ b/src/webview/common/sanitize.ts
@@ -1,0 +1,25 @@
+/*-----------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE file in the project root for license information.
+ *-----------------------------------------------------------------------------------------------*/
+
+/**
+ * Builds a sanitized compponent name from a component folder full path or
+ * a Git-repo URL
+ *
+ * @param pathOrURL an absolute path ot a Git-repo URL
+ * @returns sanitized component name
+ */
+export function buildSanitizedComponentName(pathOrURL?: string): string {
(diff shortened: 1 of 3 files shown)


src/webview/common/sanitize.ts
```diff
@@ -0,0 +1,25 @@
...
+ *-----------------------------------------------------------------------------------------------*/
+
+/**
+ * Builds a sanitized compponent name from a component folder full path or
+ * a Git-repo URL
+ *
+ * @param pathOrURL an absolute path ot a Git-repo URL
+ * @returns sanitized component name
+ */
+export function buildSanitizedComponentName(pathOrURL?: string): string {
```
datho7561, line 13: I realize this mess of regex is something that I wrote and should have tested more thoroughly. Here is a case that isn't sanitized properly: https://github.com/datho7561/12345--painful--to---manage....git...repo...name

datho7561 (changes requested):
Found an edge case with the name sanitizing function that doesn't work as expected. Otherwise, looks great and works well!

src/webview/common/sanitize.ts
datho7561: It becomes `-painful-to-manage-git-repo-name`, which isn't valid.

Minimal reproducer is `1234-a`:
- `1234` gets reduced to `` in the last step, leaving `-a`. The leading `-` is invalid

vrubezhny: Yeah, Hope it's fixed now, at least for your examples now it returns  `a` and `painful-to-manage-git-repo-name`

datho7561 (approved):
Looks good. Thanks, Victor!

PR #3408, opened 2023-10-06
Improve error message when podman is present but podman machine isn't

### LINKED ISSUE #3405: podman installed but propose to install podman
I have podman installed on my machine and wanted to start a component. I got:

image

I don't know if this is related but I have podman installed but no podman machine are defined on my system (Win11Pro)

### LINKED ISSUE #3665: Investigate replacing `Odo.isPodmanPresent`
Currently, we use `odo version -o json` in order to figure out if podman is installed on the user's system.

We depend on a library called `which` which can be used to search the user's PATH for the `podman` binary. This means we don't need to rely on `odo` to check if `podman` is installed.

We will need to be careful, since on Windows (and I'm guessing macOS as well) `odo version -o json` reports that podman is not installed if the binary is present, but podman machine doesn't exist. In [this PR](https://github.com/redhat-developer/vscode-openshift-tools/pull/3408), I'm trying to improve the error message in this case. We can probably check if the podman machine has been created by using the podman CLI, and alert the user somehow if it isn't.

datho7561:
Fixes #3405, Fixes #3665

### DIFF
--- a/src/openshift/component.ts
+++ b/src/openshift/component.ts
@@ -6,6 +6,7 @@
 import * as fs from 'fs/promises';
 import * as JSYAML from 'js-yaml';
 import * as path from 'path';
+import { which } from 'shelljs';
 import { commands, debug, DebugConfiguration, DebugSession, Disposable, EventEmitter, extensions, ProgressLocation, Uri, window, workspace } from 'vscode';
 import { Oc } from '../oc/ocWrapper';
 import { Command } from '../odo/command';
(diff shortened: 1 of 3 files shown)


src/openshift/component.ts
```diff
@@ -208,12 +209,23 @@ export class Component extends OpenShiftItem {
...
-        void window.showErrorMessage('Podman is not present in the system, please install podman on your machine and try again.', 'Install podman')
-            .then(async (result) => {
-                if (result === 'Install podman') {
-                    await commands.executeCommand('vscode.open', Uri.parse('https://podman.io/'));
-                }
-            });
+        const podmanOnPath = which('podman');
+        if (podmanOnPath) {
+            const SETUP_INSTRUCTIONS = 'Open setup instructions';
+            void window.showErrorMessage('Podman is present on the system, but is not fully set up yet.', SETUP_INSTRUCTIONS)
```
datho7561, line 215: @jeffmaury what do you think about this wording?

jeffmaury: I don't think it's gonna fix the issue as `Component.odo.isPodmanPresent` would return `true` and code won't be executed

datho7561: `odo version` only seems to report podman as installed if the podman machine is present, so in the case that the binary is on the path but the machine is not set up, it will go to this case. I've tested it on my Windows machine.

datho7561: @jeffmaury if you have time, could you please tell what you think of the wording of this message? I'll double check that it shows up under the right circumstances.

datho7561:
Okay. I've updated the pull request so that it no longer uses `odo`. This also means the logic is a bit easier to follow. When you get a chance, could you please review this PR, @msivasubramaniaan ? The three cases I'm trying to handle are:
1. `podman` is not present on the system. When the user tries to run a Devfile component on podman, there is an error message telling the user to install podman
2. `podman` is present on the system, but the podman machine is stopped (you can stop the podman machine using `podman machine stop`). When the user tries to run a Devfile component on podman, there is an error message that tells the user that podman setup is not complete, with a link to the setup docs.
3. `podman` is present and the podman machine is running. When the user tries to run a Devfile component, it starts as expected.

This can only be tested on Windows or macOS, since Linux computers (usually) don't use the podman machine (since they can run containers without starting a separate Linux VM).

mohitsuman:
@msivasubramaniaan can you please review this PR soon ?

msivasubramaniaan:
@datho7561 Shall I get demo video of it? that would help me to know the step to test the same

msivasubramaniaan (approved):
LGTM

Sorry for taking much time to review this one.
I have tested all the scenarios on my local and it is working as expected. Thanks @datho7561  for this PR

PR #3426, opened 2023-10-11
Client-side URL redirect

vrubezhny:
See: https://github.com/redhat-developer/vscode-openshift-tools/security/code-scanning/15

### DIFF
--- a/src/webview/common/devfileListItem.tsx
+++ b/src/webview/common/devfileListItem.tsx
@@ -7,11 +7,19 @@ import { Box, Chip, Stack, Tooltip, Typography } from '@mui/material';
 import * as React from 'react';
 import { Devfile } from '../common/devfile';
 import DevfileLogo from '../../../images/context/devfile.png';
+import validator from 'validator'
 
 export type DevfileListItemProps = {
     devfile: Devfile;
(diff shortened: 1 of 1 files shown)


datho7561:
This seems good to me so far. I don't know if this will get rid of the security vulnerability warning. We are reading the URLs from user-supplied Devfiles, and these Devfiles could contain any URL. I think that's what the warning is about.

vrubezhny:
> We are reading the URLs from user-supplied Devfiles, and these Devfiles could contain any URL. I think that's what the warning is about.

@datho7561 yes, basically this can be any URL, while we need only those that really point to a picture that can be shown by the ``. On the other side we pass this URL from the request data to the image `src` attribute without any validation.  The first problem (URL can have any value) probably cannot be solved, but we can try at least checking if the URL value is a well-formed URL string. And this is how I tried to fix this.  

I'd like to merge this PR anyway as it doesn't break anything while providing a simple URL validation. And also to see if it makes CodeQL happy with https://github.com/redhat-developer/vscode-openshift-tools/security/code-scanning/15 

Shouldn't CodeQL react on PR to branch main? I don't see any CodeQL sings on the PR...

Untitled

PR #3427, opened 2023-10-11
3352 add platform option into the build process on serverless function

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3352

Note: ["linux/amd64"](https://github.com/knative/func/blob/main/docs/reference/func_build.md#options) the only platform currently supported


src/serverlessFunction/functions.ts
```diff
@@ -63,31 +63,38 @@ export class Functions {
...
     }
 
-    private async buildProcess(context: FunctionObject, view: FunctionView) {
+    private async buildProcess(context: FunctionObject) {
         const isOpenShiftCluster = await Oc.Instance.isOpenShiftCluster();
         const buildImage = await this.getImage(context.folderURI);
         const terminalKey = `build-${context.folderURI.fsPath}`;
 
+        const specificPlatform = await window.showQuickPick(['Yes', 'No'], {
+            placeHolder: 'Do you want to build it for \'linux/amd64\' platform?',
```
datho7561, line 80: It's not very clear what select "No" does in this case, especially when you are on a Linux/amd64 machine. I think this message should be improved to explain what happens if you pick "No"

datho7561 (changes requested):

src/serverlessFunction/functions.ts
msivasubramaniaan: For better user experience added the context menu with build config. 
image

datho7561:
After looking a little more into it, I don't think it makes sense to have an option to build for Linux x64, since it builds for Linux x64 by default, even on Windows, where it uses docker/podman to build for Linux x64. However, I think its helpful to have separate options for the "pack" and "s2i" builds.

msivasubramaniaan:
> After looking a little more into it, I don't think it makes sense to have an option to build for Linux x64, since it builds for Linux x64 by default, even on Windows, where it uses docker/podman to build for Linux x64. However, I think its helpful to have separate options for the "pack" and "s2i" builds.

I will keep the Base build and S2I build. Do you have any better idea to incorporate the `--platform` parameter?

datho7561 (approved):
Looks good. Thanks, Muthu!

PR #3435, opened 2023-10-13
Fix create service webview

### LINKED ISSUE #3081: Fix "Create Service" view
Here are some things that I think should be fixed:
- [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form) has "theming" functionality, where you can use form components from a particular component library. We should use the [material ui theme](https://github.com/rjsf-team/react-jsonschema-form/tree/main/packages/mui) to align with the test of the extension
- The current method of opening the form is hard to find. We should put it somewhere easy to find, such as in the OpenShift Explorer view
  - In order to facilitate this, we will need to add another "page" to the form where the user selects what service they want to create before specifying the details for the service
- Many of the services you can create have a lot of unrequired properties in the JSON schema. There is a bug where unrequired boolean values are set to `false` instead of `undefined`. It would be nice to make sure these cases work properly. It would also be nice to potentially hide the unrequired properties, so that the forms are shorter (the user can always go back and configure them further

datho7561:
- Move the create service button from the Kubernetes tree to the context menu for the cluster
- Add a page to the wizard to select which CustomResourceDefinition (CRD) to create an instance of, in lieu of using the Kubernetes tree to select the CRD
  - Show the documentation for the Operator that corresponds with the CRD on this page, since many Operators require additional post-installation steps before their associated CRDs work as expected
- Material UI 5 -ify the form page
- Add a Patternfly-like collapse widget for displaying nested objects
- Remove properties from the schema if they are not required and aren't configured in the example YAML
- If the user doesn't have access to list CRDs, drop them into a YAML editor with the example YAML for the CRD
  - We used to try to scrape the schema for the CRD from the swagger definitions, but I removed this, since the schema in the swagger definitions is often missing many properties and requires further processing to get it to work with React JsonSchema Form

Fixes #3081


datho7561:
https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/89f64212-b3a1-410b-af99-82fee554f4f5

Here is a demo. In this demo, I first use "Create Operator-backed Service" while logged in as "kubeadmin". As kubeadmin, I have access to the JSON schemas for CustomResourceDefinitions (CRDs). This means I can generate a form to specify the required information to create the service. In most non-OpenShift clusters that are designed to run on your development machine (such as k3s, minikube, kind), you can access the CRDs by default.

After I have created a MongoDB instance using the kubeadmin account, I log in as "developer". I go through the same steps to create a new MongoDB instance, but instead of a form, I am given a YAML document containing the example YAML for the MongoDB CRD. This is because the developer account for OpenShift Local, doesn't have access to the full JSON schema for the MongoDB CRD. I can use "Kubernetes: Create" or "OpenShift: Create" to create the instance of the service specified by the YAML, and can save the YAML for future use.

mohitsuman (approved):
LGTM

PR #3475, opened 2023-10-20
Add "OK" and "Cancel" buttons to cluster login quick picks input fields

vrubezhny:
This PR adds a possibility to finish or cancel an entering value operation (like, for new cluster URL, new username, password, token inputs) with pressing an `Enter` or `Cancel` clickable buttons:

image

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -150,7 +150,8 @@ export class Cluster extends OpenShiftItem {
                 buttons: [deleteBtn],
             }));
             quickPick.items = contextNames;
-            quickPick.buttons = [QuickInputButtons.Back];
+            const cancelBtn = new quickBtn(new ThemeIcon('close'), 'Cancel');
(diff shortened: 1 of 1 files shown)


datho7561:
If  I try to add a new user, it errors and I get "Excpected a string but got an undefined"

vrubezhny:
No need to use keyboard if you already have your credentials saved:

[Login to a cluster](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/089b7caf-3c62-4e24-831b-e35b373d6655)

You can go back from any step of login workflow:

[Go back from any step](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/d8a09822-61d8-4cc5-b621-06ef277f7c19)

vrubezhny:
> If I try to add a new user, it errors and I get "Excpected a string but got an undefined"

@datho7561 Fixed this as well as few other issues I found (returning back from login method selection)

datho7561 (approved):
Seems to be working well. Thank you, Victor!

Issue #3452, opened 2023-10-18
When creating a component  use a first workspace folder as an initial path value 


vrubezhny:
As a developer, when creating a component I want the component folder path field to be pre-set with a Workspace folder path.

Normally a component is created in a User Workspace or its child folder, so ideally if user will need to only type this part of path value having the already pre-filled prefix part of the path value (because the path prefix can be a way longer than the  relative path to a component) :

![Screenshot from 2023-10-18 15-28-20](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/07d958f3-ad2f-49be-95cc-7cde0d22723c)

As `vscode.Workspace.rootPath` is deprecated, we still can use the first path resolved from the first `WorkspaceFolder` returned by  `vscode.Workspace.workspaceFolders[]`.

The same initial path is to be used when opening the `Select Folder` folder selection dialog.

PR #3482, opened 2023-10-24
When creating a component use a first workspace folder as an initial …

vrubezhny:
…path value #3452

Fixes: #3452

### DIFF
--- a/src/webview/create-component/createComponentLoader.ts
+++ b/src/webview/create-component/createComponentLoader.ts
@@ -18,7 +18,7 @@ import { ComponentTypesView } from '../../registriesView';
 import sendTelemetry from '../../telemetry';
 import { ExtensionID } from '../../util/constants';
 import { DevfileConverter } from '../../util/devfileConverter';
-import { selectWorkspaceFolder } from '../../util/workspace';
+import { getInitialWorkspaceFolder, selectWorkspaceFolder } from '../../util/workspace';
 import {
     getDevfileCapabilities,
     getDevfileRegistries,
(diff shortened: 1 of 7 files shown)


vrubezhny:
A most common workspace folder is used as the initial folder when creating a component:

[Screencast from 2023-10-24 18-15-16.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/e6e6ecd8-9a8b-4a45-96b4-de0b262ac45b)

datho7561:
If you open the Devfile registry viewer, and try to create a component through that workflow, you will get an error popup when you get to the "Set name and folder" page, and the folder doesn't get filled.

src/webview/create-component/createComponentLoader.ts
```diff
@@ -566,6 +583,32 @@ export default class CreateComponentLoader {
             });
         }
     }
+
+    static getInitialWorkspaceFolder(): string | undefined {
+        const wsFolders: path.ParsedPath[] = [];
+        vscode.workspace.rootPath && wsFolders.push(path.parse(vscode.workspace.rootPath));
+        vscode.workspace?.workspaceFolders?.forEach((f) => wsFolders.push(path.parse(f.uri.fsPath)));
+
+        if (wsFolders.length >= 0) {
```
datho7561, line 592: This condition is always true. I think you mean `wsFolder.length > 0`, since you access the first element of the array below.

datho7561 (changes requested):
Found a small bug in the code

src/webview/create-component/createComponentLoader.ts
vrubezhny: Yes, you're right.

vrubezhny: Fixed

vrubezhny:
> If you open the Devfile registry viewer, and try to create a component through that workflow, you will get an error popup when you get to the "Set name and folder" page, and the folder doesn't get filled.

Creating a component from Devfile Registry is fixed.

datho7561 (approved):
Looks good. Thanks, Victor!

PR #3490, opened 2023-10-27
CodeLens to apply a k8s YAML file

### LINKED ISSUE #2873: Add CodeLens for create/delete resource from YAML
It would be nice to have CodeLens to create or delete resources on the cluster from a YAML file.

Currently we have the "OpenShift: Create" command that creates resources from a YAML file, but this feature would be more visible if it appears in the document. It would be nice to also integrate with vscode-yaml in order to only show this popup when the file is a Kubernetes file (i.e. it is linked to a Kubernetes schema)

datho7561:
To try out this PR, try cloning https://github.com/quarkusio/quarkus-super-heroes and open one of the YAML files under `./deploy/k8s`. Once the extension is activated, there should be a new CodeLens at the top of the file.

This CodeLens will not appear in `devfile.yaml`.

Closes #2873

### DIFF
--- a/package.json
+++ b/package.json
@@ -191,6 +191,7 @@
 		"onCommand:openshift.oc.about",
 		"onCommand:openshift.output",
 		"onCommand:openshift.create",
+		"onCommand:openshift.delete",
 		"onCommand:openshift.open.developerConsole",
 		"onCommand:openshift.open.developerConsole.palette",
 		"onCommand:openshift.explorer.addCluster",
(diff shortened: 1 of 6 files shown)


datho7561:
oops I should also provide the option to delete

mohitsuman:
@datho7561 what is the usecase for this PR ?

Is it to directly apply the deployment YAML file to the cluster using kubectl ? Can you please summarise how this will impact/improve the workflow.

datho7561:
It makes the existing feature to apply the YAML more visible

datho7561:
A bug I expect this to have currently is that it won't handle detecting multi resource Kubernetes files, and I expect it will crash with an error log if the YAML file contains misformed YAML.

PR #3491, opened 2023-10-30
remote deployment functionality added

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3351


datho7561:
https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/3662cbf4-8e40-4940-8da6-ded07fa61081

datho7561:
A few bugs I ran into, I'll try to address them:
- The version of `func` that we provide has a bug that prevents the `PipelineRun` from cloning the git repository. Updating to the latest version fixes this. I can do it manually this time, but we should set up a GitHub Action to do this; I've created an issue for this.
- There are two warning popups that appear that say the local branch hasn't been pushed to the remote, even when it has been.

src/extension.ts
```diff
@@ -70,7 +71,9 @@ export async function activate(extensionContext: ExtensionContext): Promise<unkn
     await commands.executeCommand('setContext', 'isVSCode', env.uiKind);
     // UIKind.Desktop ==1 & UIKind.Web ==2. These conditions are checked for browser based & electron based IDE.
     migrateFromOdo018();
+    void extensionContext.globalState.update('hasTekton', await isTektonAware());
```
datho7561, line 74: I think we could do this better. Currently this blocks the extension startup, and means you need to restart the extension after you install Tekton, and doesn't take into account switching clusters.

datho7561: These checks run so fast that I think it's worthwhile rerunning it every time we go to run remote deploy.

PR #3492, opened 2023-10-30
added repo URL in helm UI

msivasubramaniaan:
Added repo URL into the UI
image

**Note**: Repo URL was hardcoded as we are currently supporting one repo.

### DIFF
--- a/src/webview/helm-chart/app/helmListItem.tsx
+++ b/src/webview/helm-chart/app/helmListItem.tsx
@@ -104,6 +104,10 @@ function HelmChartListContent(props: HelmListItemProps) {
                 >
                     {props.selectedVersion.description}
                 </Typography>
+                <Typography
+                    variant='body2'
+                    sx={{ whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis' }}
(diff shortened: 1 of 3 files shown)


src/webview/helm-chart/app/helmListItem.tsx
```diff
@@ -104,6 +104,10 @@ function HelmChartListContent(props: HelmListItemProps) {
                 >
                     {props.selectedVersion.description}
                 </Typography>
+                <Typography
+                    variant='body2'
+                    sx={{ whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis' }}
+                >Repo: https://charts.openshift.io/</Typography>
```
mohitsuman, line 110: @msivasubramaniaan Let's make the URL as a hyperlink so that users can directly open it if needed.

mohitsuman (changes requested):

src/webview/helm-chart/app/helmListItem.tsx
msivasubramaniaan: image
Done

Issue #1803, opened 2020-10-15
Can't select all Watch log content


fbricon:
While investigating #1800, I tried to watch my component. After making a local change in my java project, the watcher tried to build the project (and failed). After doing Ctrl+A, everything displayed on the page is selected (I'm only interested in the log, not the page headers). So it seems! When pasting the content, somewhere else, I either get 33 or 133 lines out of 616.

<img width="1280" alt="Screen Shot 2020-10-15 at 3 32 37 PM" src="https://user-images.githubusercontent.com/148698/96134908-ae5f6680-0efb-11eb-82bc-1f11988d5bbc.png">

dgolovin:
@fbricon workaround would be disabling WebWiew based log viewer and use terminal

![disable-webview](https://user-images.githubusercontent.com/620330/96176355-16ed1a00-0ee1-11eb-92de-7f0351147dc9.gif)

But I agree that should be working in webview in first place.

fbricon:
It's disabled by default, and indeed Show Log/Follow Log and Describe do use the terminal by default.

However `Watch Log` uses the Webview, there's no way to manually select anything using the cursor. And Ctrl+A is broken so...

dgolovin:
I look at it again and see if I could get rid of webview and keep 'Watch Sessions' view at the same time. I think that is possible with pseudo terminal.

rgrunber:
@datho7561 @fbricon , is there any advantage this option gives us over the newer OpenShift Terminal View (which is itself a webview now) ?

I would be fine with fixing this, but does it make sense to keep it longer term of the existing terminal provides everything done here ?

fbricon:
I don't care which view is kept eventually, as long as Select All works as expected

PR #3500, opened 2023-10-31
Remove the use of webview for describe and logs as replaced by OS Ter…

vrubezhny:
…minal

This commit completely removes the use of LazyLog-based webviews previously used to display the contents of component describe and show/follow log functionality.

The "openshiftToolkit.useWebviewInsteadOfTerminalView" configuration preference is also removed as well as all its usages.

Fixes: #1803


datho7561 (approved):
Looks good. Thanks, Victor!

PR #3507, opened 2023-11-01
Adding create component from git UI test

lgrossma:
see: https://github.com/redhat-developer/vscode-openshift-tools/issues/3068

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        node: [ '18' ]
+        node: [ '18.15.x' ]
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        node: [ '18' ]
+        node: [ '18.15.x' ]
```
vrubezhny, line 26: 18.15.x  doesn't look like the is to install a latest version of Node.js 18 - do we have any limitations on Node.js version related to VSCode OS Tools and/or test?

vrubezhny (commented):
Please see a question inline

lgrossma:
@vrubezhny Yes, the VSCode extension tester used for UI tests does not work with higher node versions due to selenium incompatibility.

vrubezhny:
> @vrubezhny Yes, the VSCode extension tester used for UI tests does not work with higher node versions due to selenium incompatibility.

@lgrossma Thanks! This probably was a reason why I couldn't update vscode ext. tester to 5.10.0 neither to 5.6.0.

datho7561 (approved):
Looks good, and seems to be passing consistently on my computer

PR #3508, opened 2023-11-01
List operator-backed services in the sidebar

### LINKED ISSUE #3494: List running operator-backed services in Application Explorer
It would be nice to list Operator-backed services running on the cluster in the Application Explorer.

Some notes for implementing this feature:

Operator-backed services are each just instances of a custom resource definition (CRD). If you have admin access to the cluster, you can list all custom resource definitions using `oc get crd`. If you don't have admin access to the cluster, then you can use something like this: https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/webview/create-service/createServiceViewLoader.ts#L165

Once you have the list of CRDs, you will need to find instances of the CRDs. To do this, you can use `oc get $NAME_OF_CRD` (For instance `oc get MongoDB` if you have a MongoDB Operator installed). There is a helper method in `ocWrapper.ts` called `getKubernetesObjects` that wraps this command.

datho7561:
Closes #3494


datho7561:
To test:
1. Install an operator onto the cluster that supports running an instance of a service. For instance, I used the Enterprise MongoDB Operator. I did this through the OpenShift web console
2. Create an instance of the service. Right click on the cluster in the Application Explorer in the sidebar, then select "Create Service" > "Run"
3. Go through the wizard and create a service
4. The service should appear in the Application Explorer sidebar (you will need to expand the project)

vrubezhny:
> @vrubezhny let me know if you don't have time to review this (the setup might take a while). Mohit wants this as part of the release today, so I might go ahead and merge it either way.

Reviewing it

vrubezhny (approved):
`Create Service -> Create Operator-Backed Service`   action actually hasn't  created any service for me, instead it shown me a list of Service Kinds for a cluster and, once I selected a kind, it shown me a sample YAML file that (after some editing) I could use to create a service using Console Dashboard (opened in browser). 
After the service was created on the specified Cluster project and refreshing the OS Tools Application Explorer it was shown in the  Application Explorer Tree.

Maybe I should have selected some other kind that could show a form to configure and finally create a service?

datho7561:
> Maybe I should have selected some other kind that could show a form to configure and finally create a service?

This is working as expected, although I guess the workflow is not intuitive.

The idea is that you modify any details in the YAML it gives you to, then you save the file and run the command "OpenShift: Create" from the command palette to create the service (this runs the command `kubectl create -f $FILENAME` on the file). In [this PR](https://github.com/redhat-developer/vscode-openshift-tools/commit/358bfee7ad2351f0657e8b1fd403e4bd0c5c517b), I made it so that a CodeLens (similar to Code Mining from Eclipse) appears at the top of the YAML, and when you click it it runs "OpenShift: Create".

It would be nice to provide a proper form to provide the details for the service, but the issue is that in order to access the full list of information needed for this form, you need to be logged in as a cluster administrator. If you are logged in as a cluster administrator, it should show a proper form instead of just the YAML.

PR #3521, opened 2023-11-03
Add OK/Cancel buttons to the text input  fields

vrubezhny:
This PR adds OK/Cancel buttons to the text input fields used at the creation of a cluster  namespace/project as well as at the adding of a Devfile Registry (name, URL, token)

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -4,7 +4,8 @@
  *-----------------------------------------------------------------------------------------------*/
 
 import { KubernetesObject } from '@kubernetes/client-node';
-import { ExtensionContext, InputBox, QuickInputButton, QuickInputButtons, QuickPickItem, QuickPickItemButtonEvent, ThemeIcon, Uri, commands, env, window, workspace } from 'vscode';
+import { ExtensionContext, QuickInputButtons, QuickPickItem, QuickPickItemButtonEvent, ThemeIcon, Uri, commands, env, window, workspace } from 'vscode';
+import { quickBtn, inputValue } from '../util/inputValue';
 import { CommandText } from '../base/command';
 import { CliChannel } from '../cli';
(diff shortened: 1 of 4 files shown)


vrubezhny:
All the input fields have common view like the following for normal text values:

image

and for the passwords/tokens:

image

datho7561 (commented):
Works very well from my testing.

One thing I ran into is that the back button doesn't work as expected in the "Add Devfile Registry" workflow. I think it would be helpful to have the back button working. Do you want to work on that in this PR, or should I file this as an issue to work on in the future?

vrubezhny:
> One thing I ran into is that the back button doesn't work as expected in the "Add Devfile Registry" workflow. I think it would be helpful to have the back button working. Do you want to work on that in this PR, or should I file this as an issue to work on in the future?

Thanks for looking at this. I think you're right, and `Back` button should allow to navigate back through all the fields. I'll continue working on  this PR.

vrubezhny:
Added the possibility to go back when creating a Devfile Registry item:

[Screencast from 2023-11-06 18-22-33.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/6bcbc697-1284-4413-9fa2-dbbe2f33b98e)

datho7561:
If  I enter an invalid URL (eg. `registry.stage.devfile.io` (without the `https://`)) and then hit the enter key, it freezes.

vrubezhny:
> If I enter an invalid URL (eg. `registry.stage.devfile.io` (without the `https://`)) and then hit the enter key, it freezes.

I can't reproduce any kind of freeze here. But yes, there is a bug: if input is not valid it just closes the input URL dialog leaving the promise not resolved, so it might freeze later on repeated retries.

datho7561 (approved):
Looks good and seems to working well. Thanks, Victor!

PR #3530, opened 2023-11-06
Fix `--build` option when using `func run`

### LINKED ISSUE #3522: "Build and run" for Serverless Function fails
The following error is output onto the terminal:

```
Error: "true" is not a known builder. Available builders are "pack" and "s2i"
```

datho7561:
It seems like they changed `-b` abbreviation to mean "builder" instead of "build", somewhere between 1.10.0 and 1.12.0.

Fixes #3522

### DIFF
--- a/src/serverlessFunction/commands.ts
+++ b/src/serverlessFunction/commands.ts
@@ -100,7 +100,7 @@ export class ServerlessCommand {
     static runFunction(location: string, runBuild: boolean): CommandText {
         const commandText = new CommandText('func', 'run', [
             new CommandOption('-p', location),
-            new CommandOption('-b', runBuild.toString())
+            new CommandOption('--build', runBuild.toString())
         ]);
         return commandText;
     }


vrubezhny:
I don't know what was the meaning of `-b` option previously, but the `func` version installed locally on my PC gives me:
 
 ```
 $ func version
v0.38.0-99-gcd0bc6ae

 $ func run --help
 ...
 Flags:
      --build string[="true"]   Build the function. [auto|true|false]. ($FUNC_BUILD) (default "auto")
  -b, --builder string          Builder to use when creating the function's container. Currently supported builders are "pack" and "s2i". (default "pack")
      --builder-image string    Specify a custom builder image for use by the builder other than its default. ($FUNC_BUILDER_IMAGE)
...
```

And `-b` doesn't support any `boolean` values neither is an alias to  `--build`, while looking at  provided option value `--build` perfectly suites the  `(boolean) true` value.

vrubezhny (approved):
Looks good to me .
When switched locally to this PR I cannot reproduce the issue.
Thanks!

datho7561:
Thanks for the review!

Issue #3545, opened 2023-11-07
Cannot read properties of null (reading 'getUserName')


datho7561:
In telemetry, we are receiving many errors saying "Cannot read properties of null (reading 'getUserName')".

Looking at where this function is called, I found this spot in cluster.ts: https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/openshift/cluster.ts#L628 where we call `this.getUserName`. However, this function is static. Maybe this is where the bug is? (`this` in JavaScript can get tricky)

vrubezhny:
@datho7561 I'll take a look. It might be an inpolut field validation issue. Or an initial value not set.
Are there any stacktraces reported?

datho7561:
I think this is the correct stack trace:
```
	/redhat.vscode-openshift-connector-1.10.0-linux-x64/out/src/openshift/cluster.js:585:55
<anonymous>:null:null
/redhat.vscode-openshift-connector-1.10.0-linux-x64/out/src/openshift/cluster.js:15:58
```

vrubezhny:
@datho7561 Hopefully https://github.com/redhat-developer/vscode-openshift-tools/pull/3547 will fix this.

As far as I remember VSCode (or ES Lint) was complaining on using `this.` to access a static member function... Now it's not reported anymore as an error. It might be changed something in linting.

PR #3547, opened 2023-11-07
Cannot read properties of null (reading 'getUserName') #3545

vrubezhny:
Fixes: #3545


datho7561 (approved):
Looks good. I figured out how to reproduce the issue*, and this fixes it.

* Log out of the cluster by right clicking the cluster and selecting log out. The application explorer will update to show three buttons, one of which is login. Click this button, and go through the wizard as usually. When you get to the select user step, it will close and show an error popup

PR #3567, opened 2023-11-09
[build] Update 'rjsf/*' to 5.13.6

vrubezhny:
Incorporates: #3564,  #3565,  #3566


datho7561:
Tests are passing locally for me

vrubezhny:
> Tests are passing locally for me

For some reasons I cannot normally execute Public UI Tests locally (they always fail), so I'd like to make sure that this doesn't really break anything. 
In theory, the PR could bring some changes to the UI elements forcing the  UI test failures.

datho7561:
We don't test the "Create Operator-backed Service" UI yet, so this shouldn't affect the tests.

vrubezhny:
> We don't test the "Create Operator-backed Service" UI yet, so this shouldn't affect the tests.
> 
> I tested the "Create Operator-backed Service" manually, and it seemed to be working.

For the several test runs the  UI failures were always  different but all are related to Create Component Wizard, so I suppose it could be just a random test failure or a delays in view rendering to putting the right elements to their expected coordinates.  
Problem is that if we merge this now without fixing - we'll get the main branch build failing.

datho7561:
From my understanding these tests have already been failing on the main branch on and off. (see https://github.com/redhat-developer/vscode-openshift-tools/actions/runs/6792690481/job/18466375179)

vrubezhny:
> From my understanding these tests have already been failing on the main branch on and off. (see https://github.com/redhat-developer/vscode-openshift-tools/actions/runs/6792690481/job/18466375179)

No good. But well, then I'm gong to merge this PR

Issue #3113, opened 2023-08-17
Show Helm charts from all Helm repos in the Helm chart explorer


datho7561:
The only Helm repo that is available in the Helm Chart explorer is the OpenShift Helm repo, even if the user has configured additional ones outside of VS Code OpenShift. This is a problem, since many of the charts in the OpenShift Helm repo (including most of the interesting ones like Postgres, Redis, and Jenkins) only work on OpenShift, since they rely on Routes or DeploymentConfigs.

In order to get the extension to work better with other Kubernetes clusters, I think that we should:
- Show all the charts from all the Helm repos that the user has configured in the Helm chart explorer webview
- Add some sort of indicator, perhaps a [small chip](https://mui.com/material-ui/react-chip/#sizes-chip), to show which charts come from which repo
- Add a list of checkboxes in the Helm chart explorer to show/hide charts for a particular repo

There is a mechanism to add/remove Helm repositories in the Kubernetes extension, so I don't think we need to do that as part of this issue.

datho7561:
I think @msivasubramaniaan mentioned he was working on this

msivasubramaniaan:
Hello @mohitsuman and @datho7561 

When we plan to integrate the CLI approach for getting all the installed Helm repos, `helm search repo -l -o=json` there were some discrepancies between the web API call and the CLI on helm. 

- **Web API call(https://charts.openshift.io/index.yaml) - wavefront**
```
- annotations:
      charts.openshift.io/digest: sha256:27d8a54382f1c39c55f9239bc7e17195d4c416742113fcaae1013b93b1689be0
      charts.openshift.io/lastCertifiedTimestamp: '2023-04-14T19:48:53.916688+00:00'
      charts.openshift.io/provider: VMWare
      charts.openshift.io/providerType: partner
      charts.openshift.io/submissionTimestamp: '2023-04-14T19:55:53.925857+00:00'
      charts.openshift.io/supportedOpenShiftVersions: '>=4.6'
      charts.openshift.io/testedOpenShiftVersion: '4.9'
    apiVersion: v2
    appVersion: 1.16.0
    dependencies:
    - condition: kubeStateMetrics.enabled
      name: kube-state-metrics
      repository: https://prometheus-community.github.io/helm-charts
      version: ^4.16.0
    description: Wavefront Collector for Kubernetes
    digest: 16892309c6c0be3f669be955f0ca162e432de28af631ad8e781912dd6f7cead3
[...]

PR #3568, opened 2023-11-09
Supports multiple Helm repos

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3113
helm-multi-repo


msivasubramaniaan:
Currently `CUD` operations were missing on helm on our extension. Please provide your vote for adding those based on it we can create a ticket on our backlog

vrubezhny (dismissed):
Looks good to me.

Filtering works very similar to what we have in Devfile Registry.

datho7561:
I've changed my mind on having filters for keywords, I think we should remove them. There are far too many unique keywords to display them nicely in the UI:

Screenshot from 2023-11-15 10-04-10

(for reference here are my installed helm repos:)

```
❯ helm repo list
NAME            URL                                                
bitnami-repo    https://charts.bitnami.com/bitnami                 
openshift       https://charts.openshift.io/                       
istio           https://istio-release.storage.googleapis.com/charts
ot-helm         https://ot-container-kit.github.io/helm-charts/    
wildfly         https://docs.wildfly.org/wildfly-charts/           
mongodb         https://mongodb.github.io/helm-charts
```

vrubezhny:
> I've changed my mind on having filters for keywords, I think we should remove them. There are far too many unique keywords to display them nicely in the UI:

Yeah. This looks too big. Agreed.

msivasubramaniaan:
Is it okay to show only first two keywords on each chart if it has more than two keywords. In that way atleast we can reduce the size. Wdyt?

vrubezhny:
I'm mot sure if showing the only a few  first keywords will be useful here since we don't know the importance of each keyword...

Probably we can try calculating the keyword i"mportance" based on the position of keyword among the other keywirds in the same helm chart and the number of occurrence of each keyword and then we list the first N most "important" keywords...
 
But still, making such assumptions looks very artificial, given the variety in keywords shown on David's screenshot,  so I don't know if such a selection will be useful...

@datho7561 WDYT?

datho7561:
I agree with Victor.

I think sorting the keywords by occurrence, and then only allow filtering by the top 30 (or some other number) could be helpful. However, the implementation for this will get complex:
- Performing the calculation could get expensive, given there are so many tags
- As a result, we need to be careful to cache the results of the calculation so that it only redoes it when the charts change (i.e. a repo is added or deleted)

vrubezhny (dismissed):
Looks good.

Agree on that we can create a separate issue for adding keywords filtering if needed

datho7561 (changes requested):
A few things I noticed:

1. I tried installing a few charts from the MongoDB helm repo (https://mongodb.github.io/helm-charts/) onto a KinD cluster, but I get the following error message in the webview: 
```
chart "enterprise-database" matching 1.13.0 not found in openshift index. (try 'helm repo update'): no chart name found `my-db`
```
I'm not doing anything related to openshift, so why does the error message mention openshift? **update:** I can replicate the same issue when using `crc`, it's not specific to `kind`.

2. The checkbox for hiding entries from the OpenShift Helm repo is disabled. I think it would be nice to enable it. My use case for hiding the OpenShift Helm charts is that I can't install these charts on a non-OpenShift cluster, since the charts use DeploymentConfigs (which are specific to OpenShift clusters).

3. It would be nice to remove the horizontal line below the list of "provider types" checkboxes so that it's similar to the Devfile registry

The last two comments aren't that important, but I think we should fix the first one before merging this PR.

msivasubramaniaan:
> A few things I noticed:
> 
> 1. I tried installing a few charts from the MongoDB helm repo (https://mongodb.github.io/helm-charts/) onto a KinD cluster, but I get the following error message in the webview:
> 
> ```
> chart "enterprise-database" matching 1.13.0 not found in openshift index. (try 'helm repo update'): no chart name found `my-db`
> ```
> 
> I'm not doing anything related to openshift, so why does the error message mention openshift? **update:** I can replicate the same issue when using `crc`, it's not specific to `kind`.
> 
> 2. The checkbox for hiding entries from the OpenShift Helm repo is disabled. I think it would be nice to enable it. My use case for hiding the OpenShift Helm charts is that I can't install these charts on a non-OpenShift cluster, since the charts use DeploymentConfigs (which are specific to OpenShift clusters).
> 3. It would be nice to remove the horizontal line below the list of "provider types" checkboxes so that it's similar to the Devfile registry
> 
> The last two comments aren't that important, but I think we should fix the first one before merging this PR.

Thanks for the good catch @datho7561. The reason was the repo name was hardcoded with 'openshift'. Now I have changed and made it dynamically. Please review the latest code

datho7561 (approved):
Looks good and works well. Great work, Muthu!

PR #3596, opened 2023-11-16
[build] Add arm64 odo binaries into the build

vrubezhny:


datho7561 (approved):
Looks good. Thanks, Victor!

Issue #3609, opened 2023-11-21
CUD options on helm repo


msivasubramaniaan:
Need to provide an option to add/delete/update the helm repo in the extension

PR #3617, opened 2023-11-22
Add, delete and update support on helm repo

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3609


datho7561 (changes requested):
A few things I noticed:
- Please add integration tests for the helm commands that you're running
- It would be nice if the list of charts in the chart view updated when you add/remove a repo, similar to what we do for Devfile registries
- IMO it's confusing having the list of repos in the tree view and the webview. I think it would be better if we only showed them in the webview, since from the webview you can add, edit and delete repositories, but from the tree view you can't do these things. As an alternative, maybe we can provide add, rename, and delete context menu options and just use quickpicks to set the information, similar to what we did for the devfile registry tree view.

Other things I noticed:
- I ran into this bug, which makes working with helm repos really annoying, I think we should address it in a separate issue though: https://github.com/redhat-developer/vscode-openshift-tools/issues/3622

msivasubramaniaan:
@datho7561 The integration test I'll create the separate PR and fixed all other mentioned points. Please have a look

datho7561 (approved):
Looks good, and seems to be working reliably. Thanks, Muthu!

datho7561:
Looks like there are merge conflicts

datho7561:
Looks like there are merge conflicts

Issue #3622, opened 2023-11-23
Helm Chart search is stuck loading if you close then reopen it


datho7561:
To reproduce:
- Open the Helm Chart view, it should work as expected
- Close the Helm Chart view
- Reopen the Helm Chart view. It's stuck loading.

PR #3623, opened 2023-11-24
Helm chart web view stuck when opening the second time

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3622


datho7561 (approved):
This fixes the bug for me. Thanks, Muthu!

PR #3624, opened 2023-11-24
Fix integration test debug config

datho7561:
It used to run the instrument command, however the coverage instrumentation doesn't play well with VS Code debugging. Instead, I set it to run the compile command like the other debugging configurations.


msivasubramaniaan (approved):
LGTM

Issue #3563, opened 2023-11-08
Cover "Create Serveless Function" in UI test


lgrossma:

datho7561:
Closed by #3626

PR #3626, opened 2023-11-26
Adding Create Serverless function UI test

lgrossma:
see: https://github.com/redhat-developer/vscode-openshift-tools/issues/3563


datho7561:
In order to get the tests working, I needed to make this change: https://github.com/redhat-developer/vscode-openshift-tools/pull/3772 . I think we should merge that change first. If I apply that change to this branch, then the tests pass.

datho7561 (approved):
Works great for me!

PR #3628, opened 2023-11-27
knative icon will be shown on the terminal for serverless call

msivasubramaniaan:
icon-change

As per @mohitsuman suggestion the terminal icon would be `knative` if it is a serverless function call and the `title` should be in camel case.


mohitsuman:
Fixes #3637

datho7561 (approved):
Seems to be working well. Thanks, Muthu!

PR #3629, opened 2023-11-27
UI improvements on dev file registry web view

msivasubramaniaan:
image

### DIFF
--- a/src/webview/common/devfileListItem.tsx
+++ b/src/webview/common/devfileListItem.tsx
@@ -70,7 +70,7 @@ function DevfileListContent(props: DevfileListItemProps) {
             <Stack
                 direction="column"
                 spacing={1}
-                sx={{ flexShrink: '5', minWidth: '0', maxWidth: '35rem' }}
+                sx={{ flexShrink: '5', minWidth: '0', maxWidth: '45rem' }}
             >
                 <Stack direction="row" spacing={2} alignItems="center">
                     <Typography
@@ -93,23 +93,27 @@ function DevfileListContent(props: DevfileListItemProps) {
                 </Stack>
                 <Typography
                     variant="body2"
-                    sx={{ whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis' }}
+                    sx={{ whiteSpace: 'pre-wrap', textAlign:'justify' }}
                 >
                     {props.devfile.description}
                 </Typography>
                 <Stack direction="row" spacing={1}>
-                    <Chip
-                        size="small"
-                        label="Debug Support"
-                        icon={props.devfile.supportsDebug ? <Check /> : <Close />}
-                        color={props.devfile.supportsDebug ? 'success' : 'error'}
-                    />
-                    <Chip
-                        size="small"
-                        label="Deploy Support"
-                        icon={props.devfile.supportsDeploy ? <Check /> : <Close />}
-                        color={props.devfile.supportsDeploy ? 'success' : 'error'}
-                    />
+                    {
+                        props.devfile.supportsDebug && <Chip
+                            size="small"
+                            label="Debug Support"
+                            icon={<Check />}
+                            color={'success'}
+                        />
+                    }
+                    {
+                        props.devfile.supportsDeploy && <Chip
+                            size="small"
+                            label="Deploy Support"
+                            icon={<Check />}
+                            color={'success'}
+                        />
+                    }
                     {(props.devfile.tags && props.devfile.tags.map((tag, i) => {
                         if (i >= 4) {
                             return;
--- a/src/webview/common/devfileSearch.tsx
+++ b/src/webview/common/devfileSearch.tsx
@@ -654,43 +654,43 @@ export function DevfileSearch(props: DevfileSearchProps) {
     return (
         <>
             <Stack direction="column" height="100%" spacing={3}>
-                <Typography variant="h5">{props.titleText}</Typography>
+                <Stack direction="row" useFlexGap={true}>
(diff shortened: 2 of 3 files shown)


datho7561:
I like the direction of this PR. Removing the title to better use the space is a great idea. There are a few changes that I think would be nice, though:
- In the Devfile registry view, there is no gap at the top of the page, but in the Components view, there is a gap. It would be nice if this was consistent (either both have gaps or neither has a gap)
- The "Back" button and the information box explaining what a Devfile is are hidden when opening the Devfile search from "Create Component". Removing the gap at the top of the page as mentioned in the first point should fix this.
- When you click "Show More", the leftmost column expands a little. I think it would be a good idea to add some padding to the unexpanded version of the list of tags so that it doesn't shift when you expand it. The width of the scroll bar is `var(--vscode-text-width)` (i.e. the value of the CSS variable `--vscode-text-width`)

mohitsuman:
> I like the direction of this PR. Removing the title to better use the space is a great idea. There are a few changes that I think would be nice, though:
> 
> * In the Devfile registry view, there is no gap at the top of the page, but in the Components view, there is a gap. It would be nice if this was consistent (either both have gaps or neither has a gap)

All the views should e consistent, so we should be removing the gap. The idea is to utilise the real estate in the layout.

> * The "Back" button and the information box explaining what a Devfile is are hidden when opening the Devfile search from "Create Component". Removing the gap at the top of the page as mentioned in the first point should fix this.

@msivasubramaniaan is the screenshot in the description latest of the UX changes ?

> * When you click "Show More", the leftmost column expands a little. I think it would be a good idea to add some padding to the unexpanded version of the list of tags so that it doesn't shift when you expand it. The width of the scroll bar is `var(--vscode-text-width)` (i.e. the value of the CSS variable `--vscode-text-width`)

+1 to this.

mohitsuman (changes requested):
Few Changes in the layout:

- Make search bar height smaller and placeholder inside it at center
- Pagination items at the center of search bar alignment
- Move Tags to the right and reduce logo size, to adjust more entries in the layout.
- Show the Tags Filter in same order as we show The registries
- Adjust the devfile yaml theme based on VSCode Light and Dark Theme

msivasubramaniaan:
> Few Changes in the layout:
> 
> * Make search bar height smaller and placeholder inside it at center
> * Pagination items at the center of search bar alignment
> * Move Tags to the right and reduce logo size, to adjust more entries in the layout.
> * Show the Tags Filter in same order as we show The registries
> * Adjust the devfile yaml theme based on VSCode Light and Dark Theme

Addressed all the mentioned items. Here is the latest UI:
image

src/webview/common-ext/createComponentHelpers.ts
```diff
@@ -251,25 +253,42 @@ export function getDevfileRegistries(): DevfileRegistry[] {
...
+
+function devfileSort(a: Devfile, b: Devfile): number {
+    const QUARKUS_REGEX = /[Qq]uarkus/;
+    const aQuarkus = QUARKUS_REGEX.test(a.name);
+        const bQuarkus = QUARKUS_REGEX.test(b.name);
+        if (aQuarkus && !bQuarkus) {
+            return -1;
+        } else if (bQuarkus && !aQuarkus) {
+            return 1;
+        }
```
vrubezhny, line 285: I don't understand why are the `Quarkus`  Devfile(-s) are explicitly processed here... 
If we want `Quarkus` Devfiles to be always "promoted" to be on top - it's OK, but doesn't work, because selection of a tag and following deselection makes any Tag (including the `Quarkus` one) to disappear from the visible part of the list.
[Screencast from 2023-12-04 13-45-07.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/78adc447-fea9-40cc-8d85-66bfb750ad8d)

msivasubramaniaan:
@datho7561 and @vrubezhny 

Thanks for the review. I just fixed the issue identified by @vrubezhny please take a look.

@mohitsuman FYI I moved the dev file registry selection on top.

test/ui/suite/devfileRegistries.ts
```diff
@@ -126,7 +126,7 @@ export function testDevfileRegistries() {
             // initialize web view editor
             const webView = new RegistryWebViewEditor('Devfile Registry - DefaultDevfileRegistry');
             await webView.initializeEditor();
-            expect(await webView.getRegistryStackNames()).to.include.members(['Quarkus Java', 'Django', 'Go Runtime', 'Maven Java', 'Node.js Runtime', 'Open Liberty Gradle']);
+            expect(await webView.getRegistryStackNames()).to.include.members(['Quarkus Java', 'Django', 'Maven Java', 'Node.js Runtime', 'Open Liberty Gradle']);
```
datho7561, line 129: Please add the remaining members that you expect to show up (there should be 12 entries)

datho7561 (approved):
I like that we can show more Devfiles in one page and I like the change to support light themes for the YAML viewer. I'm not enthusiastic about most of the other changes. It seems to work properly.

vrubezhny (approved):
Looks good to me. Thanks!

test/ui/suite/devfileRegistries.ts
datho7561: (not that important, I'll resolve this so we can merge)

datho7561 (approved):

datho7561:
still cannot merge?

msivasubramaniaan:
@mohitsuman FYR,

image
image
image

mohitsuman (approved):
lgtm

Issue #3543, opened 2023-11-07
Ctrl+clicking a link in the OpenShift Terminal should open the URL


datho7561:
xtermjs supports opening links by Ctrl+Clicking them. However, a bit of code needs to be added in order to support this.

This is not that bit of an issue, since you can copy any URLs you see and paste them into the browser.

PR #3641, opened 2023-11-28
Ctrl+clicking a link in the OpenShift Terminal should open the URL #3543

vrubezhny:
Fixes: #3543


datho7561 (approved):
Seems to be working well for me. Thanks, Victor!

Issue #3520, opened 2023-11-03
Error popup when expanding project in Application Explorer


datho7561:
I get the following error:

```
Command failed: "/home/davthomp/bin/oc" get openshiftpipelinesascodes.operator.tekton.dev -o=json Error from server (Forbidden): openshiftpipelinesascodes.operator.tekton.dev is forbidden: User "developer" cannot list resource "openshiftpipelinesascodes" in API group "operator.tekton.dev" at the cluster scope
```

The root cause is that the user that the developer is logged in as on the cluster does not have permissions to list instances of the given Custom Resource Definition.

PR #3644, opened 2023-11-28
Error popup when expanding project in Application Explorer #3520

vrubezhny:
Fixes: #3520


datho7561:
This seems to fix the issue now.

datho7561 (approved):

PR #3674, opened 2023-12-04
Make 'LoginUtils.requireLogin()' to not block the execution.

vrubezhny:
Checking whether the user is logged in should not block 'oc' binary execution in case login is required

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -70,4 +70,33 @@ export class CliChannel {
         return cp.spawn(toolLocation, [cmd.parameter, ...cmd.options.map((o)=>o.toString())], optWithTelemetryEnv);
     }
 
+    /**
+     * Executes a specified command in non-blocking sychronous way and returns the command output
+     *
+     * @param cmd A comand to execute
+     * @param opts Execution options that may be used to specify operation timeout, maximun buffer size, etc.
+     * @returns Output of from the command execution
+     */
+    async executeSyncTool(cmd: CommandText, opts: cp.ExecFileOptions): Promise<string> {
+        const options = {
+            ...opts,
+        } as cp.ExecFileOptionsWithStringEncoding
+
+        if (options.maxBuffer === undefined) {
+            options.maxBuffer = 2 * 1024 * 1024;
+        }
+        if (options.timeout === undefined) {
+            options.timeout = 10 * 1000; // 10 seconds
+        }
+        if (options.encoding === undefined) {
+            options.encoding = 'utf-8';
+        }
+        if (options.killSignal === undefined) {
+            options. killSignal = 'SIGILL'
+        }
+
+        const toolLocation = await ToolsConfig.detect(cmd.command);
+        const optWithTelemetryEnv = CliChannel.applyEnv(options, CliChannel.createTelemetryEnv()) as cp.ExecFileSyncOptionsWithStringEncoding;
+        return cp.execFileSync(toolLocation, [cmd.parameter, ...cmd.options.map((o)=>o.toString())], optWithTelemetryEnv);
+    }
 }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -92,7 +92,7 @@ export class CliChannel {
             options.encoding = 'utf-8';
         }
         if (options.killSignal === undefined) {
-            options. killSignal = 'SIGILL'
+            options.killSignal = 'SIGKILL';
         }
 
         const toolLocation = await ToolsConfig.detect(cmd.command);
--- a/src/oc/ocWrapper.ts
+++ b/src/oc/ocWrapper.ts
@@ -10,7 +10,6 @@ import { CommandOption, CommandText } from '../base/command';
 import { CliChannel } from '../cli';
 import { Platform } from '../util/platform';
 import { ClusterType, KubernetesConsole } from './types';
-import {isOpenShift } from '../k8s/clusterExplorer';
 
 /**
  * A wrapper around the `oc` CLI tool.
@@ -324,22 +323,6 @@ export class Oc {
         if (result.stderr) {
             throw new Error(result.stderr);
         }
-    }
-
-    /**
-     * Log out of the current OpenShift cluster.
-     *
-     * @throws if you are not currently logged into an OpenShift cluster
-     */
-    public async logout(): Promise<void> {
-        if (await isOpenShift()) {
-            await CliChannel.getInstance().executeSyncTool(new CommandText('oc', 'logout'), { timeout: 5000 });
-        }
-        // For non-OpenShift cluster, dropping the `current-context` in Kube confg may be the only
-        // way to logout.
-        // However we do it also in case of OpenShift cluster in order to make logout behavior
-        // to be consistent for all kinds of clusters
-        await Oc.Instance.unsetContext();
     }
 
     /**
--- a/src/oc/ocWrapper.ts
+++ b/src/oc/ocWrapper.ts
@@ -150,16 +150,6 @@ export class Oc {
     }
 
     /**
-     * Returns the username of the current user.
-     *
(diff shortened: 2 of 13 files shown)


src/util/loginUtil.ts
```diff
@@ -24,8 +24,11 @@ export class LoginUtil {
      * @returns true if the user needs to log in to access the cluster, and false otherwise
      */
     public async requireLogin(): Promise<boolean> {
-        return await Promise.any([Oc.Instance.getCurrentUser(), Odo.Instance.getProjects()])
-            .then(() => false)
-            .catch(() => true);
+        if (await Oc.Instance.isLoggedIn()) {
+            return await Promise.any([Oc.Instance.getCurrentUser(), Odo.Instance.getProjects()])
+                .then(() => false)
+                .catch(() => true);
```
vrubezhny, line 30: `Oc.Instance.isLoggedIn()` actually does the same as `Oc.Instance.getCurrentUser()`  but in not-blocking way, so probably we don't don't need this extra method, but we have to make `Oc.Instance.getCurrentUser()`  to be non-blocking instead. 

The thing I'd like to ask here... Once `oc whoami` returns the logged in user, should we continue checking whether no error happens on `odo list project` execution?

datho7561: I think we should be good to remove this function and replace it with `Oc.Instance.isLoggedIn()` everywhere it's used.

src/oc/ocWrapper.ts
```diff
@@ -455,4 +456,20 @@ export class Oc {
...
+    /**
+     * Checks if we're logged in. If not, any other async call to `oc`/`odo` binary might be blocked on
+     * waiting for user input
+     *
+     * @return 'true' if a user is logged in to a culster, 'false' otherwise
+     */
+    async isLoggedIn(): Promise<boolean> {
+        try {
+            const result: string = await CliChannel.getInstance().executeSyncTool(
+                new CommandText('oc', 'whoami'), { tomeout: 500 }  as CommonOptions);
```
datho7561, line 469: Please fix or remove the timeout

datho7561 (changes requested):
I did some testing on Windows and non-OpenShift clusters, and it seems to be working there as well (with the caveat that you can't log in/out of non-OpenShift clusters).

vrubezhny:
> I think the reason this patch work is that it uses `execFile` instead of `exec`. `exec` creates a shell to run the program in, whereas `execFile` directly executes the executable.  The `oc` binary is likely able to detect whether or not it's being run in a shell, so it won't produce the prompt for username and password when not being invoked through a shell. This might cause issues on Windows; I'll have to double check.

May be, but `oc` is actually prompts  for a username to be entered, which you can check  in debugger. The only thing changed here is a timeout added, which kills the `oc` process when timeout expires - thus resolving the hang-up problem 

> 
> I don't think this timeout does anything, since it's mispelt: https://github.com/redhat-developer/vscode-openshift-tools/pull/3674/files#diff-fb06bee2638b8012f20db654d2191ea1dd8448a73ebe4da36750b489a188ecbbR469 .

There is 10 seconds default value for the timeout used... So, misspelling actually doesn't break the things (But yes, it is misspelled)

datho7561:
> I thought we can make oc.Instance.getCurrentUser() to use execFile+timeout so it can be used to check if a user is logged in in LoginUtil.Instance.requireLogin() without any additional calls to Odo.Instance.getProjects() and adding any new isLoggedIn() method...
Good call. I like that refactoring better.

Maybe this will end up fixing the issue Mohit has been running into (and me occasionally) where the view takes a while to load if the kubecontext is set to an inaccessible cluster.

vrubezhny:
> Maybe this will end up fixing the issue Mohit has been running into (and me occasionally) where the view takes a while to load if the kubecontext is set to an inaccessible cluster.

I  don't think so... I haven't dig into this yet, maybe it'll speed up the things a little bit, but not too much.

vrubezhny:
> where the view takes a while to load if the kubecontext is set to an inaccessible cluster.

If a cluster "is inaccessible" we're actually aren't able to speed up anything. We still need to log in and this will take as much time as it takes (if possible at all).  The only thing we can do is limiting the time to connect/login with a timeout (and make such timeout to be configurable through a user preference - because the normal  time needed to login to different clusters  may vary a lot, so it's difficult to imagine/calculate what default value could it be).

vrubezhny:
> I did some testing ...(with the caveat that **_you can't log in/out of non-OpenShift clusters_**).

Do we have any special binary/API to be invoked for non-OpenShift clusters? I might be missing something but I thought we always use `oc login` / `oc logout`, don't we?

datho7561:
Most non-OpenShift clusters don't have a username/password system. Instead, you configure an access token in your kubeconfig, and any Kubernetes clients (eg. `kubectl`, `oc`, any libraries) should use those.

vrubezhny:
> Most non-OpenShift clusters don't have a username/password system. Instead, you configure an access token in your kubeconfig, and any Kubernetes clients (eg. `kubectl`, `oc`, any libraries) should use those.

So, the solution might be the following:
- call  `oc logout`, then:
   - if this succeeds: leave as is (if it was an openshift-like cluster,  the  :"current-context" property will not be modified on `~/.kube/config`)
   - if this fails: this is highly-likely a non-openshift cluster, so execute `oc config  unset current-context` (this will clear the  "current-context" property, making OS Tools to show "Login" and other buttons.

vrubezhny:
The demo of login/logout to Kind and OS Clusters, and switching to the these clusters in case we're already logged in:
[Screencast from 2023-12-15 20-04-45.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/76e0e5c3-7f0e-4723-af31-58a799b17869)

vrubezhny:
@datho7561 Could you, please, take a look again. Now 'Login' works (during my tests) for any type of cluster including Kind and Sandbox, the only problem is a ~3 seconds delay that may occur while checking whether we're logged in in case we're logged in to an OS cluster (potentially any remote/slow/etc. cluster) where `oc status` takes about 3 seconds to successfully complete in my case.

src/cli.ts
```diff
@@ -70,4 +70,33 @@ export class CliChannel {
...
+            options.maxBuffer = 2 * 1024 * 1024;
+        }
+        if (options.timeout === undefined) {
+            options.timeout = 10 * 1000; // 10 seconds
+        }
+        if (options.encoding === undefined) {
+            options.encoding = 'utf-8';
+        }
+        if (options.killSignal === undefined) {
+            options. killSignal = 'SIGILL'
```
datho7561, line 95: `SIGILL` -> `SIGKILL`

test/integration/ocWrapper.test.ts
```diff
@@ -221,27 +231,35 @@ suite('./oc/ocWrapper.ts', function () {
...
+            if (isOpenShift) {
+                await Oc.Instance.loginWithUsernamePassword(
+                    clusterUrl,
+                    username,
+                    password,
+                );
+                const currentUser = await getCurrentUser();
+                expect(currentUser.trim()).to.equal(username);
+            } else {
+                this.skip
```
datho7561, line 251: I think you need to invoke `skip` as a method in order for it to work

datho7561 (changes requested):
Sorry about the delay in the review. This is working very well! I found two small things in the code.

src/cli.ts
vrubezhny: Interesting.... However this really does work... 

ACK

test/integration/ocWrapper.test.ts
vrubezhny: Yes, you're right

datho7561 (approved):
Looks good and works great. Thanks Victor!

PR #3679, opened 2023-12-05
Load Application Explorer faster when cluster is inaccessible

datho7561:
Adjust the timeout to avoid waiting a really long time for the extension to load the Application Explorer.


datho7561:
I think the bug is resolved in `main`, and this PR is no longer needed.

PR #3708, opened 2023-12-08
sync repo before installation and close the model once installation success

msivasubramaniaan:
This PR sync the repository before installing the chart. Also close the model if installation was success


datho7561 (approved):
Works great! Sorry about the delay, took me a while to reproduce the issue where the charts aren't loaded.

Issue #3493, opened 2023-10-30
Error "Server doesn't have DeploymentConfig" when expanding the application explorer with KinD cluster


datho7561:
Expand the application explorer fully when attached to a KinD cluster, and you will get the following error popup:

```
Command failed: "/home/davthomp/bin/oc" get DeploymentConfig -o=json error: the server doesn't have a resource type "DeploymentConfig"
```

datho7561:
I don't see the failure any more. I guess this got fixed in a different issue.

PR #3715, opened 2023-12-10
added 'FINDSTR' when windows machine instead of grep

msivasubramaniaan:
**Problem:** 
https://github.com/redhat-developer/vscode-openshift-tools/blob/2e489f5f9ff26087c7623693782c57af91f9133b/src/oc/ocWrapper.ts#L380 always fails in windows machine wrt `'grep' `command.


vrubezhny:
This probably is related to #3493 - I can't reproduce this on Linux, but it definitely could be a reason of the  issue on Windows side. So this PR could probably fix #3493

datho7561 (approved):
Looks good. Without this change, I can't see the link to open the OpenShift Dashboard, but with it I can.

PR #3719, opened 2023-12-11
Update application explorer when component state changes

datho7561:
i.e. update the tree when the component starts or the component stops


vrubezhny (approved):
Looks good and works like a charm!

Thanks!

PR #3772, opened 2024-01-02
Display notification when logging into the cluster succeeds

datho7561:
The message string wasn't being returned properly. This is a regression that the cluster tests caught.


vrubezhny (approved):
Looks good to me, 
Please merge

PR #3782, opened 2024-01-08
Fix error processing in 'LoginUtils.requireLogin'

vrubezhny:
Error mesage: Cannot read properties of null (reading 'toLowerCase')


datho7561:
I don't know how to reproduce this issue, but "Login" seems to still work after the change

vrubezhny:
> I don't know how to reproduce this issue, but "Login" seems to still work after the change

I was doing some context switching stuff on Friday, then left the app working through the weekend (putting my note to asleep and waking it up for a few times, thus loosing the connection to the cluster) so today I had to re-login to a cluster, but  it looks like `oc status` wasn't available for some reason, so it's execution finished with an error  not returning any result in `stderr` (its value was null). The PR fixes such a case.

PR #3798, opened 2024-01-11
Deploy container image to cluster

### LINKED ISSUE #2418: Allow to deploy docker images to connected OpenShift cluster
Allow the user the option to provide the `docker image` and the extension will deploy the image to the connected OpenShift cluster.

- [ ] Detect Kubernetes environments: Use the current kubeconfig. You can also easily switch from one environment to another.
- [ ] Login into clusters: Use the connect OpenShift cluster
- [ ] List projects (namespace): Browse and select the project in which you want to deploy your application.
- [ ] Select container images: Pick and choose any container image you already have built and deployed on a container registry.
- [ ] Deploy container images: Deploy the image to the connected OpenShift cluster
- [ ] Show Logs: During deployment and build, show the logs in the container view
- [ ] Open in Browser - Open the deployed image in a new tab in the browser

datho7561:
- Add "Create Deployment from Container Image URL" wizard to project context menu
  - It uses the VS Code text input
  - It has a back button and remembers what you've entered when you go back, similar to what Victor did for the login workflow
  - Once the deployment's been created, the logs for the container are opened in the OpenShift Terminal
- Add demo gif and walkthrough entry for "Create Deployment from Container Image URL"
- Add "Delete" context menu item for Kubernetes objects (eg. Deployments)
- Add "Watch logs" context menu item for Kubernetes objects (eg. Deployments)

deploy-a-container-image

Closes #2418

### DIFF
--- a/package.json
+++ b/package.json
@@ -364,7 +364,7 @@
 			},
 			{
 				"command": "openshift.helm.openView",
-				"title": "Open Helm View",
+				"title": "Browse and Install Helm Charts",
 				"category": "OpenShift"
 			},
 			{
@@ -756,8 +756,23 @@
 				"category": "OpenShift"
 			},
 			{
+				"command": "openshift.deployment.create.fromImageUrl",
+				"title": "Create Deployment from Container Image URL",
+				"category": "OpenShift"
+			},
+			{
 				"command": "openshift.resource.load",
 				"title": "Load",
+				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.resource.delete",
+				"title": "Delete",
+				"category": "OpenShift"
+			},
+			{
+				"command": "openshift.resource.watchLogs",
+				"title": "Watch Logs",
 				"category": "OpenShift"
 			},
 			{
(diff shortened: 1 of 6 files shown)


datho7561:
@mohitsuman when you have a change, could you please tell me what do you think about this workflow?

vrubezhny (approved):
Looks and works  good,

Please take into account the suggestions listed inline

vrubezhny:
Some more things to take a look at:

After I deleted the container image I've deployed earlier to my Kind cluster and [updating/rebasing/rebuilding] restarting the OS Tools I see the following error message:

image

datho7561:
I looked into how Kubernetes handles specifying which ports to expose to the network internal to the cluster. Roland seemed to think this is done as a part of the Docker/OCI image creation (i.e. adding `EXPOSE 8080` to the Dockerfile/Containerfile to expose 8080). After some Googling, this seems to be the case. As a result, I don't think that we need to add another step in the wizard to specify the port. If the user wants to expose the container outside of the cluster, they can use the feature Muthu added for creating a Route.

vrubezhny (approved):
Looks good to me. Thanks!

PR #3804, opened 2024-01-12
Improve switch context workflow

vrubezhny:


datho7561 (approved):
Works very well! Thanks, Victor!

PR #3810, opened 2024-01-16
Correctly package binaries for linux-arm64 .

rgrunber:
- The release workflow should ensure we package the linux-arm64 binaries for a tool into the corresponding vsix, as opposed to the amd64 versions
- Update filename metadata for helm on linux-arm64

### DIFF
--- a/src/tools.json
+++ b/src/tools.json
@@ -164,7 +164,7 @@
                 "url": "https://developers.redhat.com/content-gateway/rest/mirror2/pub/openshift-v4/clients/helm/3.12.1/helm-linux-arm64.tar.gz",
                 "sha256sum": "2eb64b3d1772ae4ca0f2b34c50826860222c01f8f7306965c9aa84c11b5a83b4",
                 "dlFileName": "helm-linux-arm64.tar.gz",
-                "cmdFileName": "helm-linux-amd64"
+                "cmdFileName": "helm-linux-arm64"
             }
         }
     },
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,6 +56,7 @@ jobs:
           targets["win32-x64"]=win32
           targets["win32-arm64"]=win32
           targets["linux-x64"]=linux
+          targets["linux-arm64"]=linux-arm64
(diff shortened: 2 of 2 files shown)


src/tools.json
```diff
@@ -65,7 +65,7 @@
             },
             "darwin-arm64": {
                 "url": "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.14.7/openshift-client-mac-arm64-4.14.7.tar.gz",
-                "sha256sum": "5f045d5aa7db7e3dd2b125c5f940112a541b77726ad99e9f6b00827f9dee1621",
+                "sha256sum": "d97b84603382e24c9927150b8bfe83be89b0e63f74a593d59022483a8d039600",
```
datho7561, line 68: The sha256sum of the downloaded `.tar.gz` does not match the expected sha256sum given on the OpenShift mirror website from which the tarball is downloaded.

Here is the sum reported by the OpenShift mirror website: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.7/sha256sum.txt . It's the old one that Victor added (`5f045...`).

This might signal that the openshift executable for Apple Silicon is malicious.

vrubezhny: The checksum is taken from `https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.14.7/sha256sum.txt` file, from the following row:
```
5f045d5aa7db7e3dd2b125c5f940112a541b77726ad99e9f6b00827f9dee1621  openshift-client-mac-arm64-4.14.7.tar.gz
```
So, even if its value is  wrong, it's not our fault. 

Maybe it'll be better if we'll be really downloading and calculating the checksums when we check for the version updates?

PS: Or the value is really a correct one, but @rgrunber has some other data cached for the binary and as such performs the wrong calculation. Quite possible.

vrubezhny (approved):
Looks good to me.

PS: Probably we have to  file an issue for https://mirror.openshift.com/pub/openshift-v4/clients maintainers regarding the wrong sha256 checksums.

Issue #3591, opened 2023-11-14
Shouldn't change my kube conf if I delete the current project


adietish:
**Steps:**
1. ASSERT: have your own project (not the default one) on your cluster
2. ASSERT: your project is the current project in the kube config (`oc project` would tell you)
4. EXEC: in vscode: pick "Delete" in it's context menu
5. EXEC: check what current project it is using after it was successfully deleted
6. EXEC: verify the current project in kube config using `oc project`

**Result:**
VSCode is now using "default" as current project. The current project in the kube config was changed to "default", too.

**Expected result:**
vscode shouldn't change the current project in the kube config. It should error instead and maybe allow me to change to a current different project in the error.

This was agreed as the behavior for intellij-openshift-connector in https://github.com/redhat-developer/intellij-openshift-connector/issues/618#issuecomment-1810310127
We ideally should be consistent in intellij and vscode.

vrubezhny:
@adietish Is  [his](https://github.com/redhat-developer/intellij-openshift-connector/issues/618#issuecomment-1810310127) fixed in IntelliJ-OpenShift?  
Could you please provide a short video with or a screenshot with the result of project deletion, including possible error messages that may appear on IltelliJ as result?

vrubezhny:
@adietish After the PR #3804 is merged, the Kube config isn't touched anymore after the project is deleted on a cluster.

[Screencast from 2024-01-16 16-07-00.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/e55c05f9-9c89-4700-8858-14d6918f93ed)

The deleted project is still shown in App Explorer,  but as a "Missing project" with the only action available - `Create new or select active project`. Contexts isn't get switched after the project deletion, so you can create another project or select an existing one for this context (and Kube config is to be updated in case of different project is selected/created)   

The only problem I see (see the screencast) is that when we try to create a new project after we've just deleted the active project within the same  context - everything goes well, but the App Explorer project tree doesn't get updated, so we need to update it manually. But this is another issue.

vrubezhny:
@adietish Do have any objections to close this as fixed with #3804?

vrubezhny:
Closing as fixed with https://github.com/redhat-developer/vscode-openshift-tools/pull/3804

PR #3813, opened 2024-01-17
Force refresh App explorer on create project...

vrubezhny:
...when new project name is the same as the one set in the current-context.

The steps to reproduce:
- On Openshift cluster: create a project giving it some name
- assert that the project is successfully created, the App Explorer is refreshed automatically and the new project is lister in the App Explorer tree
- Delete the project
- Assert that the project is successfully deleted, the App Explorer is refreshed automatically and the deleted projectbow is marked as "Missing ptoject: in the App Explorer
- Click on the "Missing project" tree item to "Create new or set Active project"
- Chose "Create new project" and type exactly the same name as it was used to create the project tne the beginning
- Assert that the project is successfully created
- FAIL: Assert the App Explorer is refreshed automatically and tthe new project is listed in the App Explorer

### DIFF
--- a/src/openshift/project.ts
+++ b/src/openshift/project.ts
@@ -11,6 +11,7 @@ import { Odo } from '../odo/odoWrapper';
 import { Progress } from '../util/progress';
 import { VsCommandError, vsCommand } from '../vscommand';
 import OpenShiftItem from './openshiftItem';
+import { KubeConfigUtils } from '../util/kubeUtils';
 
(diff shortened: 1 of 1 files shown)


vrubezhny:
The screen cast demonstrating the issue: https://github.com/redhat-developer/vscode-openshift-tools/issues/3591#issuecomment-1893957743

datho7561 (approved):
Looks good and works great. Thanks, Victor!

PR #3821, opened 2024-01-18
Don't show "Open in Developer Console" on non-OpenShift clusters

### LINKED ISSUE #3814: "Open in Console" context menu item appears for non-OpenShift clusters
Victor mentioned that "Open in Console" wasn't working for Deployments created by "Create Deployment from Image URL". It opens to page with JSON and an error code instead of the OpenShift web UI.

datho7561:
- Fix condition for when the context menu item is shown

Fixes #3814

### DIFF
--- a/package.json
+++ b/package.json
@@ -1794,7 +1794,7 @@
 				},
 				{
 					"command": "openshift.resource.openInDeveloperConsole",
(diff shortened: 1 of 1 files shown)


vrubezhny (approved):
Looks good and works like a charm! Thanks!

PR #3824, opened 2024-01-18
Allow pasting in the terminal

### LINKED ISSUE #3512: Add paste to the OpenShift terminal
Some commands require entering a URL into the terminal. We should support pasting through the context menu and a keyboard shortcut.

datho7561:
To test, add an environment variable to a serverless function. This will open an interactive prompt in the terminal, which prompts for the name of the variable in the second step. There is both a keybinding (Ctrl+Shift+V) and a context menu item.

Closes #3512


msivasubramaniaan (approved):
LGTM

PR #3829, opened 2024-01-18
Pin minor version of nodejs to use in cyclic dependency check

### LINKED ISSUE #3828: CI failure due to new `npm` version
Seems like the GitHub Actions runner is now [using a newer version of npm](https://github.com/redhat-developer/vscode-openshift-tools/actions/runs/7576269259/job/20634684685?pr=3827) (10.2.3).

This is causing the CI to fail. We should look into fixing the CI.

datho7561:
Fixes #3828

### DIFF
--- a/.github/workflows/circular-dependency-check.yml
+++ b/.github/workflows/circular-dependency-check.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Use Node.js 18
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 18.17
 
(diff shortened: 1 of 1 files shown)


vrubezhny (approved):
Looks good to me.

Just a question: Is there any real problem related to Node minor version?

datho7561:
No, the problem is that with the latest version of node 18, the `npm` version will be `>=10`, and we state that `npm` must be `<10` in the `package.json`, so that check fails, causing the CI to fail.

Issue #3640, opened 2023-11-28
Show context menu to Open in Terminal for the function created


mohitsuman:
For all the Functions created, there has to be a context menu to Open that fn in the customised terminal.

Once the terminal is closed, the user should be able to navigate quickly to that terminal for the specific function called.

PR #3843, opened 2024-01-25
add terminal view context on serverless function session

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3640

### DIFF
--- a/src/webview/openshift-terminal/openShiftTerminal.ts
+++ b/src/webview/openshift-terminal/openShiftTerminal.ts
@@ -202,6 +202,10 @@ class OpenShiftTerminal {
         this._ptyExited = true;
     }
 
+    public get file() {
+        return this._file;
+    }
+
     /**
      * Returns the name of this terminal.
      *
@@ -236,6 +240,10 @@ class OpenShiftTerminal {
      */
     public get isPtyLive() {
         return this._pty && !this._ptyExited;
+    }
+
+    public get isPtyExit() {
+        return this._ptyExited;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


msivasubramaniaan:
Here is the demo for this PR: 
serverless-session

src/webview/openshift-terminal/openShiftTerminal.ts
```diff
@@ -461,6 +485,9 @@ export class OpenShiftTerminalManager implements WebviewViewProvider {
                     } else if (message.kind === 'resize') {
                         terminal.resize(message.data.cols, message.data.rows);
                     } else if (message.kind === 'closeTerminal') {
+                        if (terminal.file.endsWith('func.exe')) {
```
datho7561, line 488: 1. Please use the `onExit` callback, eg. pass these three lines of code as a lambda function here: https://github.com/redhat-developer/vscode-openshift-tools/pull/3843/files#diff-f9b7ce8b2e1d73c6d56dbcb8003e652267411f5c89e851753523d568b575fc48R121
2. This only works on Windows, since on Linux and macOS the file name is `func` (without `.exe`). However, if you use the callback method described in 1., you should be able to avoid making this check, since the callback will only be called in the `func`-related commands.

datho7561 (changes requested):

src/webview/openshift-terminal/openShiftTerminal.ts
msivasubramaniaan: **Point 1** didn't work it out, `onExit` will be called when the process terminated. But I want to refresh the tree when the terminal was closed.

**Point 2**, I will update the check based on the platform.

datho7561 (approved):
Looks good. Thanks, Muthu!

PR #3844, opened 2024-01-25
Switch to service account after logging into OpenShift Sandbox

### LINKED ISSUE #3838: Developer Sandbox Login experience enhancement
Improve the developer sandbox login experience using Builder service account.

datho7561:
The service account will remain authenticated for longer than 15 minutes.

To try this out:
1. Log into an OpenShift Sandbox cluster using the Login workflow
2. Run `oc whoami`. You should see a reference to `pipeline`, which is the serviceaccount that's being used
3. The Application Explorer should display you as logged in and work as expected

Closes #3838

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -4,6 +4,9 @@
  *-----------------------------------------------------------------------------------------------*/
 
 import { KubernetesObject } from '@kubernetes/client-node';
+import { Cluster as KcuCluster, Context as KcuContext } from '@kubernetes/client-node/dist/config_types';
+import * as fs from 'fs/promises';
+import * as YAML from 'js-yaml';
 import { ExtensionContext, QuickInputButtons, QuickPickItem, QuickPickItemButtonEvent, ThemeIcon, Uri, commands, env, window, workspace } from 'vscode';
 import { CommandText } from '../base/command';
 import { CliChannel } from '../cli';
@@ -14,7 +17,7 @@ import * as NameValidator from '../openshift/nameValidator';
 import { TokenStore } from '../util/credentialManager';
 import { Filters } from '../util/filters';
 import { inputValue, quickBtn } from '../util/inputValue';
-import { KubeConfigUtils } from '../util/kubeUtils';
+import { KubeConfigUtils, getKubeConfigFiles } from '../util/kubeUtils';
 import { LoginUtil } from '../util/loginUtil';
 import { Platform } from '../util/platform';
 import { Progress } from '../util/progress';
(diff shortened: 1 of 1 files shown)


vrubezhny:
> @vrubezhny I made the changes you mentioned, specifically:
> 
> * There is a notification box that asks if it's okay to use the service account

:heavy_check_mark:  

> * The existing context is overwritten instead of adding a new one

:grey_question: The current user is changed to `pipeline`:  

image

So now we have two users: the original one + new `pipeline` user - both with their tokens set:
```
- name: vrubezhny/api-sandbox-m3-1530-p1-openshiftapps-com:6443
  user:
    token: sha256~xxxx-xxxx
- name: pipeline
  user:
    token: 
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
Once we know for sure that the token for my original user (`vrubezhny/api-sandbox-m3-1530-p1-openshiftapps-com:6443`) will be timed out soon - there is no need to keep it here...
So why not just replace this user's token with the pipeline one instead of creating a new `pipeline` user and switching into it?

> * I changed the line wrap settings of the YAML serializer to 120 to avoid `>-` appearing in the `~/.kube/config`

:white_check_mark:  (There are some tokens that definitely contain more than 120 chars - so it's needed to watch a bit on how the YAML is serialized for those ones, but at the moment I don't see any problems

datho7561:
Okay, I've updated the PR to rewrite the token of the current user instead of creating a new one.

vrubezhny:
> > * I changed the line wrap settings of the YAML serializer to 120 to avoid `>-` appearing in the `~/.kube/config`
> 
> ✅ (There are some tokens that definitely contain more than 120 chars - so it's needed to watch a bit on how the YAML is serialized for those ones, but at the moment I don't see any problems)

Nope, 120 chars looks like is not enough:

image

src/openshift/cluster.ts
```diff
@@ -878,6 +888,35 @@ export class Cluster extends OpenShiftItem {
         return Cluster.tokenLogin(apiEndpointUrl, true, clipboard);
     }
 
+    static async installPipelineUserContext(): Promise<void> {
+        const kcu = new KubeConfigUtils();
+        const kcPath = path.join(homedir(), '.kube', 'config');
```
vrubezhny, line 893: Shouldn't we take into account the `env.KUBECONFIG` value here?

vrubezhny (changes requested):
Looks good overall and works like a charm.

The only question is that we probably must use `env.KUBECONFIG` when compiling a path to the current Kube config

vrubezhny (approved):
Looks good to me and works like a charm! Thanks!

PR #3848, opened 2024-01-29
Stop trying to connect to a cluster when it doesn't appear to be avai…

vrubezhny:
…lable

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -530,10 +530,10 @@ export class Cluster extends OpenShiftItem {
                                 );
                                 if (promptResponse === startCrc){
                                     await commands.executeCommand('openshift.explorer.addCluster', 'crc');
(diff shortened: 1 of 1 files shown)


datho7561:
We should do something similar for crc here:

https://github.com/redhat-developer/vscode-openshift-tools/pull/3848/files#diff-4c0f5d259edd0b2da811850fbc4f92c43ba5ee7d2f162594588bdaeb363050ebR536

Otherwise you will face the same problem of infinite loop (eg. when you dismiss the notification, it appears again)

datho7561 (changes requested):
This works for clusters that aren't `crc` or sandbox, but the same fix needs to be applied for `crc` and sandbox.

datho7561 (approved):
Looks good and works for me. Thanks, Victor!

PR #3865, opened 2024-02-01
Revert "Update func to 1.12.1"

vrubezhny:
Reverts redhat-developer/vscode-openshift-tools#3858


vrubezhny:
Reverting this as we already have updated `func` to 1.13.0

datho7561 (approved):
Seems good. I've tested some of the features that use `func` locally.

PR #3874, opened 2024-02-02
Delete `Odo.getKubeconfigEnv()` (it is not needed)

### LINKED ISSUE #3664: Delete `Odo.getKubeconfigEnv`
The method is only used here: https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/odo/odoWrapper.ts#L123

However, you can run `odo registry -o json` even if you don't have a `~/.kube/config`, so we shouldn't need to call it there.

datho7561:
Closes #3664


vrubezhny (approved):
Looks good to me, thanks!

PR #3875, opened 2024-02-03
Fix require login check, use 'oc whoami' to check current cluster and user

vrubezhny:


datho7561 (approved):
Looks good and works well for me. This addresses the issues I was running into with logging out. Thanks, Victor!

Issue #3878, opened 2024-02-05
When turned to using service account for a Sandbox a wrong warning message appears 


vrubezhny:
Normally, in case when a user switched into a context that uses some namespace but a different namespace/project is set after the context is actually switched to (for example, to some default project provided by the cluster for the account), we notify a user with  a warning like:

![Screenshot from 2024-02-05 15-47-33](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/518cbc38-2a70-4707-bd1c-d1037d177724)

This is OK, but the same warning appears in case of replacing the user token with the 'service account` one (see: #3844) even if the actual project is exactly the same as the one specified in the context we used to switch to, which looks wrong.

PR #3879, opened 2024-02-05
When turned to using service account for a Sandbox a wrong warning me…

vrubezhny:
…ssage appears #3878

Fixes: #3878


vrubezhny:
In order to reproduce you need to have at least two contexts pointing to a Sandbox cluster, one of them is to be a valid account (valid user and namespace), and the second one some "outdated" (pointing to a different project or an uotdated username), f.i. I have:
- a working context:
```
  - context: 
      cluster: api-sandbox-m3-1530-p1-openshiftapps-com:6443
      namespace: vrubezhny-dev
      user: vrubezhny/api-sandbox-m3-1530-p1-openshiftapps-com:6443
    name: vrubezhny-dev/api-sandbox-m3-1530-p1-openshiftapps-com:6443/vrubezhny
```
- an "outdated" context:
```
  - context:
      cluster: api-sandbox-m3-1530-p1-openshiftapps-com:6443
      namespace: vrubezhny-2-dev
      user: vrubezhny-2/api-sandbox-m3-1530-p1-openshiftapps-com:6443
    name: vrubezhny-2-dev/api-sandbox-m3-1530-p1-openshiftapps-com:6443/vrubezhny-2
```

Steps to reproduce:
- Log out from your Sandbox account
- Switch to some other OS or Kind context (so you'll be able to switch to any of your Sandbox accounts later)
- Copy your Sandbox's active user token to clipboard 
- Switch to the real working Sandbox context using your token and when asked if you want to use a `service account` token - answer `Yes`.
- Observe the switching is complete with no Warnings

- Log out from your Sandbox account and switch to any other  OS or Kind context
- Again copy your Sandbox's active user token to clipboard  (if needed)
- Switch to the "outdated" Sandbox context using your token and when asked if you want to use a `service account` token - answer `Yes` (actually if you choose `No` the Warning message is still to appear)
- Observe the switching is complete with  but the mentioned Warning appeared this time (because the real project and user you're switched to are different from the ones specified in the selected context)
- Observe (in your Kube config) that finally the context you're switched to is your real working Sandbox context, not the "outdated" one

vrubezhny:
> I think I probably misunderstand something. If you're trying switching to the outdated context, but then using the token causes you to switch to the valid context, then it makes sense to present the warning that the namespace is different.

This is exactly what it should do. 
But bug is that when you're switching to a VALID sandbox context and select "switch to using service account" you still be informed on that project/namespace is different while the cluster/user/namespace are actually the same. 
(And the issue is that I get context info too early (before it's really logged in and switched, so the new context is compared not against the selected context, but the previously used one)

vrubezhny:
@datho7561 here is the demo of the wrong warning message appearance:

[Screencast from 2024-02-12 15-02-20.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/084180a3-b147-4521-8a14-8f62cd5c8db4)

We're going to switch to project `vrubezhny-dev` on a sandbox cluster, once logged in we got proposed to switch to the service account and it's done after we agreed. So, finally we're logged in and switched to the required `vrubezhny-dev` project on the required sandbox cluster - so no any warning is to be displayed.  But it IS  still displayed.

datho7561 (approved):
Looks good. Thank you, Victor!

Issue #3850, opened 2024-01-29
"Components" sidebar section take a while to load


datho7561:
When starting up, the "Components" sidebar takes a while to load (a couple of seconds).

A further investigation will need to be performed as to why (eg. which function calls take the longest).

datho7561:
Here's how to reproduce the slowness:
1. set up your kubeconfig to point to an invalid cluster. In my case, I point it to my local `crc` instance, but don't start `crc`
2. launch VS Code, then open the OpenShift sidebar. All the views in the sidebar should take over a minute to load.

Here is the different layes
1. We call `getAllComponents()` in `activate`
2. This calls `OdoWrapper.Instance.getRegistries()` to get the list of registries
3. `getRegistries()` uses `odo preference view -o json` in order to dump the config as JSON, which contains the list of registries.

In the version of `odo` that we bundle with the extension (3.15.0), this call takes ~40 seconds if the cluster is inaccessible. i.e. you can replicate this slow response on command line by calling `odo preference view -o json` using odo 3.15.0 . Therefore, the root cause is an upstream issue.

However, we can mitigate the implications of this issue by not calling `getAllComponents` in `activate`. I don't know why we were doing this anyways, fetching and updating the list of devfile registries should be encapsulated in the registry tree view.

PR #3880, opened 2024-02-05
Do not call getAllComponents() in activate

datho7561:
Part of #3850


datho7561:
Good catch! I fixed this issue, and refactored the code related to collecting the list of Devfiles.

datho7561:
Serverless function tests are failing for me. The root causes could be that something is preventing the refresh button from being pressed, or that VS Code reports it has run out of filesystem watchers during the tests.

vrubezhny (approved):
Looks good to me - Devfile Registry editing works fine and GH Actions tests are passed.
Thanks!

Issue #3812, opened 2024-01-17
As a user I initially couldn't find "Change Active Project"


adietish:
I suggest to 
* add "Change Active Project" to the context menu of the cluster- and project-node in the tree. 
* use a more intuitive icon than the 2 arrows. 

This would improve user friendlyness I think. Here's how I was behaving/thinking as a noob user:

When I tried to switch the current project/namespace, I looked into the context menu but didnt find it. So I thought that the extension doesn't have this functionality. I then noticed the 2 arrows to the right of the cluster- and project node and thought that it was about refreshing or syncing. When I then clicked on it I found out that this was the functionality I was looking for.

datho7561:
- I think removing the "change project" on the cluster is a good idea
- we should check if vscode-k8s already has something in the toolbar
- we should check if there is enough room on the toolbar to add a new item for the project

vrubezhny:
I've just added a draft PR that in Application Explorer this:

- removes 'Change Active Project' inline action from the Cluster tree item (but it's still shown for the Project tree item);
- adds 'Change Active Project' context menu item to the Cluster and Project tree items context menu.

![Screenshot from 2024-02-06 19-15-00](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/3017da9e-de90-484b-8af9-5b40c19038bd)  ![Screenshot from 2024-02-06 19-15-32](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/7518ea82-9832-4909-82fd-a02f2469efca)


I searched thrught the 'standard' vscode [icons](https://code.visualstudio.com/api/references/icons-in-labels) but I couldn't find any acceptable icon to be a replacement for ![image](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/901c63df-759a-400c-a99d-4b94c9128338)...
The only icons I can find for "changing/replacing/sync something" are: 
[...]

PR #3887, opened 2024-02-06
As a user I initially couldn't find "Change Active Project" #3812

vrubezhny:
In Application Explorer this:
- removes 'Change Active Project' inline action from the Cluster tree item (but it's still shown for the Project tree item);
- Adds 'Change Active Project' context menu item to the Cluster and Project tree items context menu.
- Disables the k8s Context and Namespace status bar items and adds the similar OpenShift Tools own items that use OpenShift Tools own actions to switch context/change current project

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,6 +204,7 @@
 		"onCommand:openshift.explorer.refresh",
 		"onCommand:openshift.componentTypesView.refresh",
 		"onCommand:openshift.project.create",
+		"onCommand:openshift.project.set",
 		"onCommand:openshift.project.delete",
 		"onCommand:openshift.project.delete.palette",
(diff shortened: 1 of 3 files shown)


vrubezhny:
The changes made are:

- [Context menu items changed for Set active project action ](https://github.com/redhat-developer/vscode-openshift-tools/issues/3812#issuecomment-1930702496)
- [Status Bar Items from k8s extension are replaced by the OS Tools own ones for Current Context and Current Namespace/Project informers/Pickers](https://github.com/redhat-developer/vscode-openshift-tools/issues/3812#issuecomment-1945111404)

datho7561 (approved):
Looks good and works great! Thanks, Victor!

PR #3908, opened 2024-02-15
Don't prompt before switching over to service account on Sandbox

datho7561:
Per our discussion, this makes the experience smoother

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -849,11 +849,7 @@ export class Cluster extends OpenShiftItem {
             try {
                 await Oc.Instance.loginWithToken(clusterURL, ocToken);
                 if (Cluster.isOpenShiftSandbox(clusterURL)) {
-                    const YES = 'Yes';
-                    const result = await window.showInformationMessage('OpenShift Sandbox logs you out after 15 minutes. Would you like to switch to a service account to prevent this?', YES, 'No');
-                    if (result === YES) {
-                        await Cluster.installPipelineUserContext();
-                    }
+                    await Cluster.installPipelineUserContext();
                 }
                 return Cluster.loginMessage(clusterURL);
             } catch (error) {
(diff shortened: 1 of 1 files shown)


vrubezhny (commented):
It happens that if I'm logged out and trying to login to sandbox using a provided token, I get the following error:
Screenshot from 2024-02-16 12-02-48

in line 914 of `installPipelineUserContext()`:
```
        const pipelineToken = Buffer.from(pipelineTokenSecret.data.token, 'base64').toString();
```
... because `pipelineTokenSecret` is undefined. (I've taken a look on gathered secrets - there is really no a secret which `matadata.name` is starting with `pipeline-token`).  

As result  I'm logged in, but still using the provided token, not the service account one

vrubezhny:
> (I've taken a look on gathered secrets - there is really no a secret which `matadata.name` is starting with `pipeline-token`).

It doesn't matter if I user this PR or `main` branch, currently the situation is quite the when I try to login to `api-sandbox-m3-1530-p1-openshiftapps-com:6443` (created just today after my previous Developer Sandbox for Red Hat OpenShift account was expired).

So in any case we have to check if the service account token is really available on a cluster

datho7561:
Which ServiceAccounts are available for you on the cluster? You should be able to find this in the OpenShift Console web site switching to "Administrator", then clicking on "User Management" > "ServiceAccounts" in the sidebar.

vrubezhny:
> Which ServiceAccounts are available for you on the cluster? You should be able to find this in the OpenShift Console web site switching to "Administrator", then clicking on "User Management" > "ServiceAccounts" in the sidebar.

image

datho7561:
hmm, so they removed the `pipeline` account, which was the one with the `edit` permissions. I'm now starting to wonder if this is intentionally fixing a security vulnerability. The 15 minute long token is a security mechanism that prevents someone from using a computer where someone forgot to log out to make changes to the cluster. Switching to the long-lived token breaks this mechanism, so I was surprised the cluster provided such an account by default.

vrubezhny:
Yes, looks reasonable.  
But still I believe there should be a way for developers to setup an isolated non-production cluster and get an  account with no any time limitation (for development purposes, and not for production). The other question is if the same cluster is used for the development and for the production - then it's probably not possible.

datho7561:
@vrubezhny I'm going to modify this PR so that it doesn't install the token if there is no `pipeline` ServiceAccount. We should get this in for the release.

vrubezhny:
> @vrubezhny I'm going to modify this PR so that it doesn't install the token if there is no `pipeline` ServiceAccount. We should get this in for the release.

Any news on possibility to have Service Account available in future?

vrubezhny (approved):
Looks good to me. Thanks!

Issue #3822, opened 2024-01-18
Context menu item to scale Deployments


datho7561:
It might be nice to provide a context menu item to scale deployments.

eg. see this issue for intellij-kubernetes: https://github.com/redhat-developer/intellij-kubernetes/issues/615

datho7561:
Here's what the workflow looks like in intellij-kubernetes:

https://github.com/redhat-developer/vscode-openshift-tools/assets/22376627/33a413bf-b196-4803-a964-0c8c235264d9

Here's what I think the workflow should be:
1. Right click a deployment in the context menu, and select "Scale"
2. Enter a number (positive or zero) using a quick pick to specify the new number of replicas
3. The deployment is scaled to the new number of replicas

PR #3910, opened 2024-02-16
scaling deployment replicas

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3822

### DIFF
--- a/src/deployment.ts
+++ b/src/deployment.ts
@@ -108,7 +108,7 @@ export class Deployment {
                     return 'Scale value should be a number';
                 } else  if (parseInt(trimmedValue, 10) < 0) {
                     return 'Scale value should be equal or greater than zero';
-                } else if (Deployment.checkFloat(trimmedValue)) {
+                } else if (Deployment.checkFloat(trimmedValue) || trimmedValue.indexOf('.') !== -1) {
                     return 'Scale value should be integer';
                 }
             },
--- a/src/deployment.ts
+++ b/src/deployment.ts
@@ -11,7 +11,7 @@ import { CommandText } from './base/command';
 import { OpenShiftExplorer } from './explorer';
 import { Oc } from './oc/ocWrapper';
 import { validateRFC1123DNSLabel } from './openshift/nameValidator';
-import { quickBtn } from './util/inputValue';
+import { inputValue, quickBtn } from './util/inputValue';
 import { vsCommand } from './vscommand';
 import { OpenShiftTerminalManager } from './webview/openshift-terminal/openShiftTerminal';
 
@@ -85,6 +85,32 @@ export class Deployment {
     static shellIntoDeployment(component: KubernetesObject): Promise<void> {
         void OpenShiftTerminalManager.getInstance().createTerminal(new CommandText('oc', `exec -it ${component.kind}/${component.metadata.name} -- /bin/sh`), `Shell to '${component.metadata.name}'`);
         return Promise.resolve();
+    }
+
+    @vsCommand('openshift.deployment.scale')
+    static async scalingPod(component: KubernetesObject): Promise<void> {
+        if (!component) {
+            return;
+        }
+        let count = '1';
+        count = await inputValue(`How many replicas would you like to scale ${component.kind}/${component.metadata.name}?`,
+            count,
+            false,
+            (value: string) => {
+                const trimmedValue = value.trim();
+                if (trimmedValue.length === 0) {
+                    return 'Scale value cannot be empty';
+                }
+                if (parseInt(trimmedValue, 10) < 0) {
+                    return 'Scale value should greater than zero';
+                }
+            },
+            'Scale Count'
+        );
+        const response = await Oc.Instance.scalePod(component.metadata.name, count);
+        if (response.indexOf('scaled') !== -1) {
+            OpenShiftExplorer.getInstance().refresh(component);
+        }
     }
 
     /**
--- a/src/deployment.ts
+++ b/src/deployment.ts
@@ -8,7 +8,7 @@ import * as path from 'path';
 import validator from 'validator';
 import { Disposable, QuickInputButtons, ThemeIcon, TreeItem, window } from 'vscode';
 import { CommandText } from './base/command';
-import { OpenShiftExplorer } from './explorer';
+import { DeploymentPodObject, OpenShiftExplorer } from './explorer';
 import { Oc } from './oc/ocWrapper';
 import { validateRFC1123DNSLabel } from './openshift/nameValidator';
 import { inputValue, quickBtn } from './util/inputValue';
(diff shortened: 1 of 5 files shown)


src/explorer.ts
```diff
@@ -300,6 +323,10 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
                     result = [...helmRepos.sort(Helm.ascRepoName)];
                 }
             }
+        } else if ('kind' in element && element.kind === 'Deployment') {
+            const pods = await Oc.Instance.getKubernetesObjects('pods');
+            const filteredPods: DeploymentPodObject[] = pods.filter((pod) => pod.metadata.name.indexOf(element.metadata.name) !== -1 && pod);
```
datho7561, line 328: I don't think `&& pod` is needed here. `&& pod` will check if pod is not `null` or `undefined`, but since pod has a field `metadata` that you access, we know it's not null.

src/deployment.ts
```diff
@@ -87,6 +87,32 @@ export class Deployment {
...
+        let count = '1';
+        count = await inputValue(`How many replicas would you like to scale ${component.kind}/${component.metadata.name}?`,
+            count,
+            false,
+            (value: string) => {
+                const trimmedValue = value.trim();
+                if (trimmedValue.length === 0) {
+                    return 'Scale value cannot be empty';
+                }
+                if (parseInt(trimmedValue, 10) < 0) {
```
datho7561, line 104: I am allowed to enter 0.001 into the wizard. When this is passed to `oc scale deployment`, it fails.

I think we should do one of:
- run `parseInt` on count before passing it to `oc scale deployment`
- Only allow integer values in the wizard

src/deployment.ts
```diff
@@ -87,6 +87,32 @@ export class Deployment {
...
+                    return 'Scale value should greater than zero';
+                }
+            },
+            'Scale Count'
+        );
+        const response = await Oc.Instance.scalePod(component.metadata.name, count);
+        if (response.indexOf('scaled') !== -1) {
+            OpenShiftExplorer.getInstance().refresh(component);
```
datho7561, line 112: If you are increasing the number of replicas (eg. 1 --> 3), then the new pods don't show up when you refresh the tree here, since the pods take a while to reach the "Running" state. I think this can be a bit confusing for the user.

However, I can't think of a way to address this other than refreshing the Application Explorer on a timer, which we are trying to avoid, so I think this is good for now.

src/explorer.ts
```diff
@@ -196,14 +205,28 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
                 }
+            } else if (element.kind === 'Pod') {
+                const contextElement: DeploymentPodObject = element;
+                return {
+                    contextValue: 'openshift.k8sobject.Deployment.pod',
+                    label: contextElement.metadata.name,
+                    description: `${contextElement.kind.substring(0, 1).toLocaleUpperCase()}${contextElement.kind.substring(1)}`,
+                    collapsibleState: TreeItemCollapsibleState.None,
+                    iconPath: new ThemeIcon('layers-active'),
+                    tooltip: `${contextElement.status.phase} (1/1)\n${contextElement.status.podIP}`,
```
datho7561, line 216: I don't think it's worth listing (1/1) if this value is hard coded. For example, what if pod contains a sidecar container, so there should be two containers listed?

Also, you filter out pods that are in the "Running" phase, so the first part will always be "Running"

src/explorer.ts
```diff
@@ -300,6 +323,10 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
                     result = [...helmRepos.sort(Helm.ascRepoName)];
                 }
             }
+        } else if ('kind' in element && element.kind === 'Deployment') {
+            const pods = await Oc.Instance.getKubernetesObjects('pods');
+            const filteredPods: DeploymentPodObject[] = pods.filter((pod) => pod.metadata.name.indexOf(element.metadata.name) !== -1 && pod);
+            return filteredPods.filter((pod) => pod.status.phase === 'Running');
```
datho7561, line 329: May I ask what's the reason for only listing pods that are running? I think it could be helpful to show pods that are starting or stopping.

datho7561 (changes requested):
Works pretty well! Here are some things I noticed

src/explorer.ts
vrubezhny: I also don't get the meaning of that hard coded `(1/1)` string in pod's tooltip.

Maybe this is a place that later is to be changed to something like `(pod #x of y)` or something like this, but if not, I don't see any point in hard coding it here.

src/deployment.ts
```diff
@@ -87,6 +87,32 @@ export class Deployment {
         return Promise.resolve();
     }
 
+    @vsCommand('openshift.deployment.scale')
+    static async scalingPod(component: KubernetesObject): Promise<void> {
+        if (!component) {
+            return;
+        }
+        let count = '1';
```
vrubezhny, line 95: Do we store somewhere or can we find out somehow and use the current scale number as an initial value?

src/explorer.ts
vrubezhny: > May I ask what's the reason for only listing pods that are running? I think it could be helpful to show pods that are starting or stopping.

+1

vrubezhny (changes requested):
Agree with David's notes.  +see  some questions inline.

src/explorer.ts
msivasubramaniaan: Removed `pod`

src/deployment.ts
msivasubramaniaan: Scale value now accepts only positive integer

src/deployment.ts
msivasubramaniaan: Refreshing the tree view based on `'Pod is Running'` text on terminal view. Now we can get the updated pods

src/explorer.ts
msivasubramaniaan: There were no possibilities of getting the number of `pod` on running state. So removed the `(1/1)` hardcoded value

src/explorer.ts
msivasubramaniaan: Now listing all the `pods` from cluster

msivasubramaniaan:
@datho7561 and @vrubezhny addressed all the comments which you mentioned. Please have a look on the latest code

vrubezhny:
When I'm scaling to 0, I see  all the running pods are shutting down (and disappearing from the App explorer tree)  with the following lines output to the OS Terminal:
```
Cleaning resources, please wait
 ✗  Finished executing the application (command: debug) [43s]
 ✗  watch command was unable to push component: fail starting the port forwarding: multiple Pods exist for the selector: component=nodejs-starter-1. Only one must be present
 ```  
 Is this normal? I mean the second error: `.watch command was unable to push component:...`

**Comment by vrubezhny** [comment]
After I scaled to 2, when clicking on a Deploypent or on one of the  running Pods the Deployment YAML is to be opened... So after a while it stops opening Deployment YAML when  clicking on a Pod and suddenly  stops `Dev` for a Component...

Not sure if this is related to Scaling or is it any different issue:

image

[Screencast from 2024-02-26 18-37-24.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/a9989fb5-ce0c-442f-b50e-3d7c980fb4b8)

**Comment by msivasubramaniaan** [comment]
> After I scaled to 2, when clicking on a Deploypent or on one of the running Pods the Deployment YAML is to be opened... So after a while it stops opening Deployment YAML when clicking on a Pod and suddenly stops `Dev` for a Component...
> 
> Not sure if this is related to Scaling or is it any different issue:
> 
> image
> 
> [Screencast from 2024-02-26 18-37-24.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/a9989fb5-ce0c-442f-b50e-3d7c980fb4b8) This time the error is `Shutting down the API server...`:
> 
> ```
>  ✗  Finished executing the application (command: debug) [43s]
>  ✗  watch command was unable to push component: fail starting the port forwarding: multiple Pods exist for the selector: component=nodejs-starter-1. Only one must be present
> I0226 18:38:39.986832  259197 starterserver.go:147] Shutting down the API server: context canceled
> ```

We don't have control on getting the `terminating` state of the pod, that's why I have shown only running pods. If you check it on the cluster the mentioned pod was terminated and there were no config of it. so it fails.

File: src/deployment.ts [hunk 7]
```diff
@@ -87,6 +87,46 @@ export class Deployment {
...
+                const trimmedValue = value.trim();
+                const number = Number(trimmedValue);
+                if (trimmedValue.length === 0) {
+                    return 'Scale value cannot be empty';
+                }
+                if (isNaN(number)) {
+                    return 'Scale value should be a number';
+                } else  if (parseInt(trimmedValue, 10) < 0) {
+                    return 'Scale value should be equal or greater than zero';
+                } else if (Deployment.checkFloat(trimmedValue)) {
```
datho7561, line 111: I am allowed to scale to `2.0`, but when I do this, it fails, since the string `2.0` is passed to the `oc` command that scales the deployment.

src/openshift/component.ts
```diff
@@ -362,9 +362,11 @@ export class Component extends OpenShiftItem {
                         Component.stateChanged.fire(component.contextPath);
                     },
                     onText(text: string) {
-                        if (cs.devStatus === ComponentContextState.DEV_STARTING && text.includes('[p]')) {
+                        if (cs.devStatus === ComponentContextState.DEV_STARTING) {
```
datho7561, line 365: This is bad. The odo component isn't running until the '[p]' shows up. With this current change, it lists the component as running before the source code is synced into the running pod. After odo syncs the source code, it still needs to perform whatever build steps the Devfile specifies, and set up port forwarding so that the user can actually access the component.

Also, we should change `devStatus` and fire the event indicating that `devStatus` changed at the same time. With this change, `devStatus` is changed to running as soon as odo outputs any text, and the event that indicates it changed is fired when "Pods is Running" is output.

datho7561 (changes requested):

datho7561 (approved):
Works like a charm! Thanks, Muthu!

vrubezhny (approved):
Looks good to me.
The scaling actually happens - the pods are created and deleted according the Scale value requested, the only  the positive integers are accepted.

PR #3927, opened 2024-02-22
Set `dockerode` as a runtime dependency

### LINKED ISSUE #3926: Extension is unusable after installing from vsix file
When installing the extension from vsix file and launching it, all the welcome content buttons remain greyed out and trying to click on one of the action buttons results in an error like this: `Error running command openshift.explorer.login: command 'openshift.explorer.login' not found. This is likely caused by the extension that contributes openshift.explorer.login.` .This happens both on my mac and my linux vm.

datho7561:
Fixes #3926


mohitsuman (approved):
lgtm

Issue #3918, opened 2024-02-20
Fix Action labels for kubernetes clusters


vrubezhny:
This is a follow up issue for VSCode Openshift Tools for  https://github.com/redhat-developer/intellij-openshift-connector/issues/701

When using a kubernetes-based cluster, i would like to see "Create Namespace/Change namespace" and not Openshift "project" labels in context menus.

PR #3936, opened 2024-02-26
Fix Action labels for kubernetes clusters #3918

vrubezhny:
Fixes: #3918

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,6 +207,10 @@
 		"onCommand:openshift.project.set",
 		"onCommand:openshift.project.delete",
 		"onCommand:openshift.project.delete.palette",
+		"onCommand:openshift.namespace.create",
+		"onCommand:openshift.namespace.set",
+		"onCommand:openshift.namespace.delete",
+		"onCommand:openshift.namespace.delete.palette",
 		"onCommand:openshift.component.openCreateComponent",
 		"onCommand:openshift.component.createFromRootWorkspaceFolder",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -207,6 +207,10 @@
 		"onCommand:openshift.project.set",
 		"onCommand:openshift.project.delete",
 		"onCommand:openshift.project.delete.palette",
+		"onCommand:openshift.namespace.create",
+		"onCommand:openshift.namespace.set",
+		"onCommand:openshift.namespace.delete",
```
datho7561, line 212: Make sure to hide this entry from the command palette (this one expects a namespace name to be passed to the command, whereas the `.palette` one opens a quick pick)

datho7561 (changes requested):
One quick change, but otherwise seems great!

package.json
vrubezhny: We actually didn't have any `openshift.*.delete.palette` - like command handlers registered.

So, I've hidden the mentioned `.delete` commands from Command Palette and added the according `.delete.palette` commands:

[Screencast from 2024-03-04 23-39-39.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/547d3220-3917-4b49-92e6-386d3ccdb35c)

datho7561 (approved):
Looks good and works like a charm. Thanks, Victor!

Issue #3931, opened 2024-02-23
New Serverless Function is not shown after creation


lgrossma:
After creating a new Serverless Function, the Serverless Function view does not show the created function even though it exists in the filesystem.

msivasubramaniaan:
@lgrossma can you please tell us on which build you get this error? On latest build I couldn't able to reproduce

lgrossma:
@msivasubramaniaan I am testing on this commit ef0f2de5a0ad891ff764918953080d29fa504fa4.

https://github.com/redhat-developer/vscode-openshift-tools/assets/86789336/411e63e8-a173-4c29-a22a-a2167f403412

PR #3944, opened 2024-02-27
recursively get the function folder

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/3931


datho7561 (approved):
Looks good and fixes the problem!

Issue #3057, opened 2023-07-25
Stabilize UI tests in cluster-dependent suite


lgrossma:
UI tests requiring running cluster are failing on our jenkins instance.

lgrossma:
@mohitsuman If you are talking about the new release job that was migrated from jenkins, then I can't see any segment that would run the UI tests.

mohitsuman:
@lgrossma Is it still failing? We are no longer running Jenkins instances for release, so I would not worry much about this. And the PR changes with the UI tests should also run on GH action.

lgrossma:
Right now nothing fails in the cluster-dependent test suite, as the problematic tests were turned off due to the create component UI changes (I am currently working on new tests for creating component). These tests have nothing to do with the release. The cluster-dependent test suite is set of tests that need an openshift cluster running and cannot be run on GH actions.

lgrossma:
@mohitsuman Currently the tests are not even running on our jenkins due to node version incompatibility with selenium. I have prepared a fix, but haven't tried it yet. I also expect some instabilities in the test suite but it is hard to say what the actual scope of work will be required for fixing the instabilities..

PR #3960, opened 2024-03-05
Cluster dependant suite fix

lgrossma:
This PR aims to stabilize the cluster dependant UI test suite https://github.com/redhat-developer/vscode-openshift-tools/issues/3057


datho7561 (approved):
I ran locally on crc. It seems pretty stable from my testing.

Issue #3947, opened 2024-02-28
The App Explorer tree isn't refreshed after logging in to a Kind cluster


vrubezhny:
Steps to reproduce:

- Login to a Kind cluster, see the App Explorer tree is updated and showing the cluster contents
- Logout, see the Login button appeared 
- Login once again to  King cluster, FAIL: Login button still appears instead of the App Explorer tree
- Refresh, see the App Explorer tree

[Screencast from 2024-02-28 15-30-18.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/1a9af29c-459e-4448-be84-4b7a1585b8d9)

PR #3966, opened 2024-03-06
The App Explorer tree isn't refreshed after logging in to a Kind cluster #3947

vrubezhny:
Fixes: #3947


datho7561 (approved):
This fixes the problem for me

PR #3972, opened 2024-03-07
use node 20 in CI build

mohitsuman:


datho7561:
Tests are working, although extester is only targetting Node 18 and VS Code should be using NodeJS 20.

Could you please update the node version in two other spots:
- `.nvmrc`
- `package.json`: the `engines` field on line 18

PR #3977, opened 2024-03-11
Adding login/logout test

lgrossma:


vrubezhny (dismissed):
Looks good to me

kubectl
datho7561, line 1: I understand that we need the kubectl to get the tests to run, but I think we should avoid adding the binary directly to the repository.

At the very least, we should add this file to the `.vscodeignore`, so that it doesn't get packaged when we bundle the extension. This binary will only work on Linux, and we support macOS and Windows.

datho7561 (changes requested):
Looks good to me!

kubectl
lgrossma: You are right. This file should not be included and got into my PR by mistake when I was trying something.

datho7561: The main issue was that the binary was being packaged into the distributed .vsix, even those built for macOS and Windows. I see it's resolved now, thanks, Lukas!

datho7561 (approved):
Looks good. Thanks, Lukas!

lgrossma:
I ran the tests again with the changes and they still pass.

Issue #3991, opened 2024-03-18
Re-arrange application explorer similar to vsKubernetes cluster view


msivasubramaniaan:
Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/3981

PR #4003, opened 2024-03-20
Showing tree nodes similar to k8s in application explorer

msivasubramaniaan:
**Fixes:** https://github.com/redhat-developer/vscode-openshift-tools/issues/3991

image


datho7561:
Please hide `DeploymentConfigs`, `ImageStreams`, and `BuildConfigs` from the tree if the cluster doesn't have them (i.e. any non-OpenShift cluster doesn't have these).

datho7561 (approved):
Seems alright to me. One thing that I find confusing is that if I create an Operator-backed service, then it appears under "Deployments" despite the fact that it's not a deployment. For instance, here I have two ServiceBindings and a CouchDB Cluster created on the cluster, and they are listed under Deployments, despite the fact that their types are `ServiceBinding` and `CouchbaseCluster`:

Screenshot from 2024-03-22 15-02-54

I don't think there is a nice way to handle this if we use the tree view with the different types. The reason is, the best way to handle this is to provide different tree items for each potential type. You can view the full list of potential types using `oc api-resources`. However, if we do this, then we need to add many more types to the tree. If you are using a local cluster mostly for practicing with Kubernetes, then most of these categories are empty.

As it stands I think this implementation of the view from vscode-kubernetes is good, I just don't understand why we are trying to do this in the first place.

Issue #4013, opened 2024-03-25
Provide an option to delete the Pod like k8s


msivasubramaniaan:
Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/3981
![image](https://github.com/redhat-developer/vscode-openshift-tools/assets/93245779/3d49d3b3-6f09-4840-93e8-49d6a4a2b33f)

PR #4014, opened 2024-03-25
added delete, load context menu for pod

msivasubramaniaan:
Fixes: https://github.com/redhat-developer/vscode-openshift-tools/issues/4013

### DIFF
--- a/package.json
+++ b/package.json
@@ -1923,7 +1923,7 @@
 				},
 				{
 					"command": "openshift.resource.watchLogs",
-					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject\\.!(helm&&pod)/"
+					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject\\.(?!helm|pod)/"
 				},
 				{
 					"command": "openshift.deployment.shell",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1923,7 +1923,7 @@
 				},
 				{
 					"command": "openshift.resource.watchLogs",
-					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject\\.(?!helm)/"
+					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject\\.!(helm&&pod)/"
```
datho7561, line 1926: Unfortunately after this change, the watch logs and open in developer console context menus are always hidden.

I think the correct regex is `/^openshift\\.k8sObject\\.(?!helm|pod)/`

Please check out https://www.rexegg.com/ and https://regex101.com/ for working with regex

package.json
```diff
@@ -1939,7 +1939,7 @@
 				},
 				{
 					"command": "openshift.resource.openInDeveloperConsole",
-					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject/ && isOpenshiftCluster"
+					"when": "view == openshiftProjectExplorer && viewItem =~ /^openshift\\.k8sObject\\.!(helm&&pod)/ && isOpenshiftCluster"
```
datho7561, line 1942: The same change needs to be done here so that "Open in Developer Console" works

datho7561 (changes requested):
Other than the comments I left, it seems to be working well.

datho7561 (approved):
Looks good. Thanks, Muthu!

PR #4028, opened 2024-03-28
rearranged the explorer view

msivasubramaniaan:
image

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -181,6 +181,12 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
 
         // It's a Helm installation
         if ('chart' in element) {
+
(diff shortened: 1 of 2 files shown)


datho7561:
I don't understand how different resource types are being grouped together here. Deployments are a type of workload (see https://kubernetes.io/docs/concepts/workloads/controllers/).

mohitsuman:
> I don't understand how different resource types are being grouped together here. Deployments are a type of workload (see https://kubernetes.io/docs/concepts/workloads/controllers/).

We can separate deployment as an individual tree as we want OpenShift devs to focus more on that. All other controllers can go inside the Workload tree, so that the view is clean.

mohitsuman:
@datho7561 can you please review this and merge once approved ?

datho7561 (approved):
Works well.

PR #4029, opened 2024-03-28
Fix test case fails on installing helm chart

msivasubramaniaan:


datho7561:
Still fails. Maybe try Jenkins? That worked locally for me:
chart name: `redhat-jenkins`
char version: `0.0.3`

msivasubramaniaan:
Follow up with https://github.com/redhat-developer/vscode-openshift-tools/pull/4030

Issue #3925, opened 2024-02-21
stdout maxBuffer length exceeded


TTAbwill:
Type: <b>Bug</b>

## Describe the bug
Error message with OpenShift Cluster 4.12.49 by opening a namespace with a lot of deployments: stdout maxBuffer length exceeded

## Expected Behavior
List namespace objects

## Current Behavior
Error: stdout maxBuffer length exceeded

## Steps to Reproduce
1. Connect to OpenShift cluster
2. Open a namespace with a lot of objects

## Environment

Extension version: 1.11.0
VS Code version: Code 1.86.2 (903b1e9d8990623e3d7da1df3d33db3e42d80eda, 2024-02-13T19:40:56.878Z)
OS version: Windows_NT x64 10.0.22631
Modes:
Remote OS version: Linux x64 5.14.0-378.el9.x86_64

lgrossma:
@TTAbwill can you please specify how many is "a lot of objects". I tried this with 15 deployments and got no error.

TTAbwill:
@lgrossma There are 216 deployments and some statefulsets ending up in about 400 pods inside the project.

lgrossma:
@vrubezhny I managed to replicate this. Created a cluster with over 200 deployments and when trying to list all the Pods under the Workloads tree item, I got the same error. The deployments itself loaded fine though. Also the Image Streams under the workloads loaded without a problem. Seems like the pods were the culprit in this case..

vrubezhny:
@lgrossma Any stacktraces available?

datho7561:
I tried reproducing on kind, but I was unable to run the extension due to my files being open. I tried reproducing on OpenShift Sandbox, but got resource limited to 30 deployments. I'll try `minikube` or `crc` next, maybe that will get around the "opened files" limitation, since it should be running in a VM.

PR #4061, opened 2024-04-12
Make stdout maxBuffer length to be configurable

vrubezhny:
The default value is set to 4Gb.

Fixes: #3925


datho7561:
With this change, I am able to open the "Pods" and "Deployments" sections if I create 100 deployments. I am going to stress test 400 to see what happens.

vrubezhny:
> With this change, I am able to open the "Pods" and "Deployments" sections if I create 100 deployments. I am going to stress test 400 to see what happens.

The number of pods really matters: the error (`ERR_CHILD_PROCESS_STDIO_MAXBUFFER`) happens when executing `oc get pods -o=json` , and according to the reporter and @lgrossma 's experience it takes place when the number of pods is 200+.

datho7561 (approved):
4 MB allows for viewing over 400 deployments, this should (hopefully) be plenty

datho7561:
I used
```lua
#!/usr/bin/lua

for i = 1, 400, 1 do
  os.execute(string.format('oc create deployment --image quay.io/davthomp/pet-alpine my-pet-deploy%d', i));
end
```
To create 400 deployments, each with one pod. I opened both the deployments and the pods sections. No crashes occured, it loaded properly.

PR #4064, opened 2024-04-16
Add proposal message to change stdout max buffer length preference...

vrubezhny:
...in case of 'ERR_CHILD_PROCESS_STDIO_MAXBUFFER' error appearance


vrubezhny:
A short demo of the proposal message for changing the preference value for the stdout max buffer length:

[Screencast from 2024-04-17 00-57-04.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/197a594e-781d-43af-8b7c-1e8e38532bc5)

datho7561 (approved):
Looks good, and is really helpful. Thanks, Victor!

PR #4084, opened 2024-04-29
Reduce memory usage of build process

datho7561:
This takes the esbuild memory usage from 5 Gb to 1.8 Gb, for both the command line run and the background build process for VS Code. It accomplishes this by doing everything in one esbuild call instead of many parallel esbuild calls. The tradeoff is that the directory structure of the `out` folder has changed, so I had to change some webview code to accomidate this change.


datho7561:
Please let me know if this prevents the out of memory crash, Muthu. The other thing that could be crashing is eslint, since the new eslint plugin I added for linting JSDoc is kind of slow and takes a lot of memory, and vscode-eslint runs eslint in the background if you have it installed.

datho7561:
Muthu told me offline this was working, so I'll go ahead and merge this.

datho7561:
Just going to through all the webviews to make sure I didn't mess anything up...

Some images don't load, so I'll need to address that.

PR #4109, opened 2024-05-08
Fix login to sandbox workflow

datho7561:
- Install portfinder as a runtime dependency (needed to start the extension when running as a vsix)
- Remove the hack to create a service account to avoid being logged out (it seems the user is no longer allowed to list service accounts).
- Trim the whitespace surrounding the copied token (the token gets padded with spaces for whatever reason).
- Find and switch to the user's project when logging into a sandbox cluster
- Fix the missing logo in the add cluster page
- Resolve a NPE in the Application Explorer view when logging in, which would result in an error popup. Instead, the cluster name will be blank for a few milliseconds after logging in.


vrubezhny (approved):
Looks good to me. Thanks!

Issue #4101, opened 2024-05-07
Project not listed if it has a display name annotation


savissimo:
I have a project that has a `display-name` annotation:

```yaml
metadata:
  annotations:
    openshift.io/display-name: Internal applications
```

This project is not listed in the quick pick when I try to change the active project. I tracked the bug down to [this line](https://github.com/redhat-developer/vscode-openshift-tools/blob/9dffe63503ec5ebf7a13c0ba0796eb2567ee4421/src/oc/ocWrapper.ts#L553), that excludes the project because `oc projects` outputs it like this:

    int-apps - Internal applications

As soon as I removed the annotation, the project appeared in the list. 

Also, as long as the annotation is present, the project is not found and I can't even list its deployments or resources. 

Can this be fixed to take into account display names please?

vrubezhny:
@savissimo Could you please try restoring the annotation, execute `oc projects -q` and provide its full output here?

savissimo:
The `-q` option does not show the display name. Here is the output after restoring the annotation:

```
default
int-apps
kube-node-lease
kube-public
kube-system
ldap-sync
openshift
openshift-apiserver
openshift-apiserver-operator
openshift-authentication
openshift-authentication-operator
openshift-cloud-controller-manager
openshift-cloud-controller-manager-operator
openshift-cloud-credential-operator
openshift-cloud-network-config-controller
openshift-cluster-csi-drivers
openshift-cluster-machine-approver
openshift-cluster-node-tuning-operator
openshift-cluster-samples-operator
openshift-cluster-storage-operator
openshift-cluster-version
openshift-config
openshift-config-managed
openshift-config-operator
openshift-console
openshift-console-operator
openshift-console-user-settings
openshift-controller-manager
openshift-controller-manager-operator
openshift-debug-nbvdf
openshift-debug-qsxqt
openshift-debug-t5hw5
openshift-dns
openshift-dns-operator
openshift-etcd
openshift-etcd-operator
openshift-host-network
openshift-image-registry
openshift-infra
openshift-ingress
openshift-ingress-canary
openshift-ingress-operator
openshift-insights
openshift-kni-infra
openshift-kube-apiserver
openshift-kube-apiserver-operator
openshift-kube-controller-manager
openshift-kube-controller-manager-operator
openshift-kube-scheduler
openshift-kube-scheduler-operator
openshift-kube-storage-version-migrator
openshift-kube-storage-version-migrator-operator
openshift-machine-api
[...]

Issue #4080, opened 2024-04-26
API token for sandbox login showing failed


Vatsal-Kathiriya:
Type: <b>Bug</b>

## Steps to Reproduce

1.
2.
3. 

## Environment

Extension version: 1.13.0
VS Code version: Code 1.88.1 (e170252f762678dec6ca2cc69aba1570769a5d39, 2024-04-10T17:41:02.734Z)
OS version: Windows_NT x64 10.0.22631
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|12th Gen Intel(R) Core(TM) i5-12500H (16 x 3110)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|15.71GB (6.70GB free)|
|Process Argv|--crash-reporter-id 8cbc38cd-84df-4743-a5f4-465b3fcceb3b|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492cf:30256860
vscod805:30301674
binariesv615:30325510
vsaa593cf:30376535
py29gd2263:31024239
c4g48928:30535728
azure-dev_surveyone:30548225
962ge761:30959799
pythongtdpath:30769146
welcomedialog:30910333
pythonidxpt:30866567
pythonnoceb:30805159
asynctok:30898717
pythontestfixt:30902429
pythonregdiag2:30936856
pyreplss1:30897532
pythonmypyd1:30879173
pythoncet0:30885854
2e7ec940:31000449
pythontbext0:30879054
accentitlementsc:30995553
dsvsc016:30899300
dsvsc017:30899301
dsvsc018:30899302
cppperfnew:31000557
d34g3935:30971562
fegfb526:30981948
bg6jg535:30979843
ccp2r3:30993541
dsvsc020:30976470
[...]

PR #4122, opened 2024-05-13
Improve set active or create new project/namespace workflow

vrubezhny:
- Now, when switching projects/namespaces, one can manually type in a project name to be set as an active one
- No more 'Missing project/namespace' item for a project/namespace that doesn't exist on a cluster. This allows
  working normally on clusters with restrictions on list for projects.

Fixes: #3999

- The project listing is fixed, so annotated projects are shown now

Fixes: #4101

- For a Sandbox cluster a project which name contains current user name is used as a default one when logging in (A follow up to #4109)

Issue: #4080 (?)

### DIFF
--- a/src/openshift/project.ts
+++ b/src/openshift/project.ts
@@ -19,25 +19,42 @@ export class Project extends OpenShiftItem {
     static async set(): Promise<string | null> {
         let message: string = null;
         const kind = await getNamespaceKind();
+        const canCreateProjects = await Oc.Instance.canCreateNamespace();
+        const canListProjects = await Oc.Instance.canListNamespaces();
+
         const createNewProject = {
             label: `Create new ${kind}`,
             description: `Create new ${kind} and make it active`
         };
+        const manuallySetProject = {
+            label: `Manually set active ${kind}`,
+            description: `Type in ${kind} name and make it active`
+        };
         const projectsAndCreateNew = Oc.Instance
             .getProjects() //
-            .then((projects) => [
-                createNewProject,
-                ...projects.map((project) => ({
-                    label: project.name,
-                    description: project.active ? 'Currently active': '',
-                })),
-            ]);
+            .then((projects) => {
+                const items = [];
+                if (canCreateProjects) {
+                    items.push(createNewProject);
+                }
+                items.push(manuallySetProject);
+                if (canListProjects) {
+                    items.push(...projects.map((project) => ({
+                        label: project.name,
+                        description: project.active ? 'Currently active': '',
+                    })));
+                }
+                return items;
+            });
         const selectedItem = await window.showQuickPick(projectsAndCreateNew, {placeHolder: `Select ${kind} to activate or create new one`});
         if (!selectedItem) return null;
         if (selectedItem === createNewProject) {
             await commands.executeCommand('openshift.project.create');
         } else {
-            const projectName = selectedItem.label;
+            const projectName = selectedItem === manuallySetProject ?
+                await Project.getProjectName(`${kind} name`, Promise.resolve([])) : selectedItem.label;
+            if (!projectName) return null;
+
             await Oc.Instance.setProject(projectName);
             OpenShiftExplorer.getInstance().refresh();
             Project.serverlessView.refresh();
@@ -50,7 +67,7 @@ export class Project extends OpenShiftItem {
     @vsCommand('openshift.namespace.create')
     static async create(): Promise<string> {
         const kind = await getNamespaceKind();
-        const projectList = Oc.Instance.getProjects();
+        const projectList = Oc.Instance.getProjects(true);
         let projectName = await Project.getProjectName(`${kind} name`, projectList);
         if (!projectName) return null;
         projectName = projectName.trim();
@@ -76,7 +93,7 @@ export class Project extends OpenShiftItem {
     static async delFromPalette(): Promise<string | null> {
         const kind = await getNamespaceKind();
         const projects = Oc.Instance
-            .getProjects() //
+            .getProjects(true) // Get only projects existing on cluster
             .then((projects) => [
                 ...projects.map((project) => ({
                     label: project.name,
--- a/package.json
+++ b/package.json
@@ -1720,12 +1720,12 @@
 				},
 				{
 					"command": "openshift.project.set",
-					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext && canCreateNamespace && isOpenshiftCluster",
+					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext && isOpenshiftCluster",
 					"group": "c1@2"
 				},
 				{
 					"command": "openshift.namespace.set",
-					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext && canCreateNamespace && !isOpenshiftCluster",
+					"when": "view == openshiftProjectExplorer && viewItem == openshift.k8sContext && !isOpenshiftCluster",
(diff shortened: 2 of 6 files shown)


datho7561:
`getActiveProject()` fails with a null dereference if you can't list projects, since the return value of `fixActiveProject` is null. One consequence of this failure is that the Application Explorer says you are logged out, even if you are allowed to perform some actions on the cluster (such as listing pods on all namespaces).

`oc projects` fails if the user cannot list namespaces, so you might need to use another method to detect the current project. Probably the best (and maybe only) approach is to read the kubeconfig.

I haven't tried this on an OpenShift instance, I've been using `kind`. I can do a short write up on creating a service account that can't list projects on a kind cluster if you are interested.

vrubezhny:
> I haven't tried this on an OpenShift instance, I've been using `kind`. I can do a short write up on creating a service account that can't list projects on a kind cluster if you are interested.

@datho7561 This would be really helpful.

vrubezhny:
> `oc projects` fails if the user cannot list namespaces, so you might need to use another method to detect the current project. Probably the best (and maybe only) approach is to read the kubeconfig.

I use Kube config to detect an active project, because the way I run `oc projects`, even if it succeeds, doesn't allow to get the active project info from its output (I execute it with `-q` argument, which makes its output more short and not requiring any special parsing, but also make it to lack an info on which project is currently set as active.  (However I expect `oc` also to use Kube config in order to get the active project)   

The problem may happen when Kube config has active project set, but that project doesn't really exist on the cluster. 
Previously a `Missing project` was shown on the tree with the selected name, while with this fix (if there are any projects exist on the cluster) the first available project will be actually selected, but not the chosen one (either thru Kube config or by entering its name manually). This probably is worth to be restored.

vrubezhny:
Changed `fixActiveProject` so:
- it never returns `null`. 
- If Kube Config has a namespace set that doesn't exists in the cluster projects, it doesn't change the `active` property of any cluster project , but still sets such "manually" set project as a active in Kube config (So it won't be changed neither to `default` nor to the first listed project, thus making it possible to correctly select active project when a cluster doesn't allow to get the list of projects)

adietish:
Things dont work for me properly when only listing pods is allowed (as instructed in #3999):
The current namespace ("default") is displayed but reported as missing. 
image
This is erroneous since I can list pods that exist in this namespace:
```bash
$ kubectl get pod -n default
NAME                           READY   STATUS             RESTARTS           AGE
apple-app                      1/1     Running            0                  67m
banana-app                     1/1     Running            0                  67m
frontend-9cr6z                 0/1     ImagePullBackOff   0                  67m
frontend-kjrpx                 0/1     ErrImagePull       0                  67m
frontend-vbhb7                 0/1     ErrImagePull       0                  67m
mehdb-0                        0/1     CrashLoopBackOff   18 (15s ago)       68m
sise-deploy-69d88467b4-vhfsd   0/1     CrashLoopBackOff   3415 (2m10s ago)   58d
 ```

**Comment by adietish** [comment]
Furthermore I cannot select my current namespace. I cannot confirm the upcoming prompt, it stays open (namespace "default" exists, I simply cannot retrieve it. I can only retrieve pods within it):


https://github.com/redhat-developer/vscode-openshift-tools/assets/25126/ac345651-49a2-47f7-a3b0-99edb007c68b

**Comment by vrubezhny** [comment]
> Furthermore I cannot select my current namespace. I cannot confirm the upcoming prompt by hitting the return key, it stays open (namespace "default" exists, I simply cannot retrieve it. I can only retrieve pods within it):
> 

It's a filter, not an input field for a project/namespace name, you should click `Create new... ` or `Manually set...` item first, then type in a project name:


[Screencast from 2024-05-15 18-17-48.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/add039df-8529-4348-9d3d-ba1c3c51f0a8)

**Comment by vrubezhny** [comment]
I've made no item like `xxx - Missing project/namespace` to appear anymore. For a project/namespace that doesn't exist on a cluster or exist but cannot be listed due to restrictions in cluster configuration  a normal project/namespace item will appear with all its children. The content of the children items `Deployments`/`Workloads`/etc is to be defined depending on the existence or applied restrictions.

**Comment by adietish** [comment]
@vrubezhny Things work now fine 😄 
I can use the namespace "default", there's no warning "Missing Namespace" any more and I can list the pods that are running and that I am allowed to list (according to my custom, restrictive RBAC).
The only thing that I am wondering about is that listing forbidden resources cause error notifications to appear. These errors are expected since I am not allowed to list these resources. 
image
I am wondering if a child node with the error would be preferrable (screenshot of intellij-kubernetes)?

**Review by adietish — APPROVED** [review]
@vrubezhny Things work fine now 😄 

I can use the namespace "default", there's no warning "Missing Namespace" any more and I can list the pods that are running and that I am allowed to list (according to my custom, restrictive RBAC).

The only thing that I am wondering about is that listing forbidden resources cause error notifications to appear while the cathegory stays empty.
image
I am wondering if a child node with the error would be preferrable (screenshot of intellij-kubernetes)?
image

**Comment by datho7561** [comment]
> @datho7561 What do you think about this proposal? If we can detect (by `oc auth can-i get ...` or somehow else) that listing a node objects is restricted, we can create some meaningful labels like `Deployments cannot be shown due to the cluster configuration restrictions` or similar and not to try to get, f.i., Deployments children which definitely leads into the mentioned error message appearance.

I like this idea.

**Comment by datho7561** [comment]
I think we can use `kubectl api-versions` to detect if a user is authenticated with the server when they are using a non-OpenShift cluster. I'll double check, but I think this API is available when a user is authenticated with a server, regardless of the roles they have been assigned.

**Comment by vrubezhny** [comment]
The `Change Active Project/Namespace` inline and context menu actions are now enabled without any restrictions on OpenShift and Kubernates cluster projects. The contents of quick pick is limited according to the restrictions applied to the cluster: in the most restricted case (when it's not allowed to create or list projects/namespaces) it will contain the only one item `Manually set active Namespace/Project`:

[Screencast from 2024-05-16 22-39-15.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/cb461449-8ec5-44b8-8a1d-408b86e0d9a9)

File: src/openshift/project.ts [hunk 1]
```diff
@@ -19,25 +19,42 @@ export class Project extends OpenShiftItem {
...
+                return items;
+            });
         const selectedItem = await window.showQuickPick(projectsAndCreateNew, {placeHolder: `Select ${kind} to activate or create new one`});
         if (!selectedItem) return null;
         if (selectedItem === createNewProject) {
             await commands.executeCommand('openshift.project.create');
         } else {
-            const projectName = selectedItem.label;
+            const projectName = selectedItem === manuallySetProject ?
+                await Project.getProjectName(`${kind} name`, new Promise((resolve) => {resolve([])})) : selectedItem.label;
```
datho7561, line 55: For the `new Promise((resolve) => { resolve([]) })` you can instead use `Promise.resolved([])`

datho7561 (changes requested):

src/openshift/project.ts
vrubezhny: Fixed

vrubezhny:
Now the App. Explorer Tree on a "restricted" cluster looks like the following:

Screenshot from 2024-05-17 16-54-57

cc @adietish

datho7561 (approved):
Seems good to me!

datho7561:
@vrubezhny feel free to merge, unless you want to wait on feedback from Andre on the appearance of the error tree nodes.

PR #4133, opened 2024-05-16
Use oc api-resources for checking login

datho7561:


datho7561:
To check that this is working, create a kind cluster where you can only list pods, using the instructions here: https://github.com/redhat-developer/idetools.dev/pull/49.

When you open VS Code OpenShift, it should load the application explorer so that the cluster is listed, but it won't detect that the default project exists, because we still need https://github.com/redhat-developer/vscode-openshift-tools/pull/4122 for that.

Before this PR, the application explorer would not show that you are logged in, and would provide the three buttons for logging in.

datho7561:
Victor has implemented this in his PR.

PR #4140, opened 2024-05-19
Create route functionality added

msivasubramaniaan:


vrubezhny:
@msivasubramaniaan Could you please rebase the PR on top of main branch?

msivasubramaniaan:
> @msivasubramaniaan Could you please rebase the PR on top of the main branch?

@vrubezhny I have done it already. Seems there were no more commits on main branch

vrubezhny:
An annoying effect occurs on the Create route view when typing in something to the text fields... When trying to type in something into the middle of the existing text after every character is added  it moves the text cursor to the end of the string, so you have to manually move the cursor back to the required position before typing in the next character:  

[Screencast from 2024-05-22 13-51-18.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/7994d73f-059e-43b3-8834-f6a0a6bed7f1)

... I haven't spotted this on any other views we have.

vrubezhny:
> > @msivasubramaniaan Could you please rebase the PR on top of the main branch?
> 
> @vrubezhny I have done it already. Seems there were no more commits on main branch

I don't think it's true, neither GH thinks it is:

image

msivasubramaniaan:
> > > @msivasubramaniaan Could you please rebase the PR on top of the main branch?
> > 
> > 
> > @vrubezhny I have done it already. Seems there were no more commits on main branch
> 
> I don't think it's true, neither GH thinks it is:
> 
> image

image

vrubezhny (changes requested):
Just some extra lines to be removed here. Plus a few questions I asked in the main thread (field editing question, error on Kind cluster and recovery after a route creation error)

msivasubramaniaan:
@vrubezhny 

I've addressed all mentioned review comments. please have a look

vrubezhny:
@msivasubramaniaan  Now for Sandbox cluster I have: 

image

IMHO, the validator for Path field of `Create Route` view is to be fixed acordingly (`The Route "mr" is invalid: spec.path: Invalid value: "mr": path must begin with /`)

vrubezhny:
image

The 'Routes' node as well as `Create... -> Create Route` are Openshift specific, so these items/menu are to be shown only for Openshift clusters.

src/explorer.ts
```diff
@@ -355,6 +355,16 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
+
+            // The 'Create Route' menu visibility
+            let services: K8sResourceKind[] = [];
+            try {
+                services = await getServices();
+            }
+            catch (_) {
+                // operator framework is not installed on cluster; do nothing
+            }
+            void commands.executeCommand('setContext', 'showCreateRoute', services.length > 0);
```
vrubezhny, line 367: The `Create Route` should be visible only for Openshift clusters

src/explorer.ts
```diff
@@ -453,7 +469,7 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
                     name: 'build configs'
                 },
             } as OpenShiftObject;
-            result.push(pods,
+            result.push(pods, routes,
```
vrubezhny, line 472: The `Routes` App. Explorer tree item is an Openshift specific, so it should be moved inder `if (isOpenshiftCluster) {...}` here after.

vrubezhny (changes requested):
@msivasubramaniaan As routes is an Openshift specific objects - we should show it as well as `Create Route` context menu item only when working with an Openshift cluster

datho7561:
If you are trying to expose a Deployment using a Route, you first need to create a [Service](https://kubernetes.io/docs/concepts/services-networking/service/) for the Deployment.

I had to do this from the command line (`oc expose deployment/my-deployment-name --port 8080`).

It might be nice to incorporate creating the service if it doesn't already exist as a step in the route creation process. For instance, the user selects the deployment instead of the service, then when they submit the form, the extension checks if there is an existing service connect to the deployment and creates it if there isn't one, then it exposes the service.

vrubezhny (approved):
Looks good to me. Thanks!

@msivasubramaniaan Don't forget to open a separate issue for `creating the service if it doesn't already exist as a step in the route creation process` - I agree in that it would be a useful addition.

datho7561:
I am okay with handling creating a service for a deployment in a separate issue.

We should squash the commits before merging, this means you will need to do a rebase on main @msivasubramaniaan .

Issue #3845, opened 2024-01-25
Show deployment logs doesn't work if you switch projects


datho7561:
The "show deployment logs" feature works by running the command to watch the logs on a timer until it succeeds.

Currently if the user switches projects after creating a new deployment from a container image but before the logs appear, then the notification that says 'opening deployment logs' will be loading forever.

The way to remedy this is to specify the namespace in https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/explorer.ts#L376, so that even if the user switches namespaces, the correct namespace is being checked for the deployment.

PR #4145, opened 2024-05-21
Show deployment logs doesn't work if you switch projects #3845

vrubezhny:
Fixes: #3845


datho7561 (approved):
Works for me! Thanks, Victor!

Issue #3823, opened 2024-01-18
Display status information for Deployments in Application Explorer tree


datho7561:
It would be nice to report status information for deployments. For example, we can include:
- The number of available replicas
- The number of configured replicas
- If the Deployment is in CrashLoopBackoff
- Some way to indicate any other errors the Deployment has run into

For the last two, maybe we could do some sort of warning/error icon, with hover text that gives the specific error message.

vrubezhny:
As an example of a "Successful" deployment:

![Screenshot from 2024-05-22 02-50-31](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/645af555-f3ef-4a67-9f7c-473b3b6ebfb5)

...and a "Failing" deployment:
![Screenshot from 2024-05-22 02-57-18](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/5dfe9c93-258d-4728-ae2e-1c00c5fe9fdd)

Not sure if I correctly get on how can I detect a "Failed" deployment, as I'm not sure if Pods statuses are really replicated to the according Deployment's `element.status.conditions` array... A Deployment `conditions` array looks more like a log of events, like:
```
  conditions: [
    {
      lastTransitionTime: "2024-05-16T18:42:14Z",
      lastUpdateTime: "2024-05-16T18:42:14Z",
      message: "Deployment has minimum availability.",
      reason: "MinimumReplicasAvailable",
      status: "True",
      type: "Available",
    },
    {
      lastTransitionTime: "2024-05-16T18:41:59Z",
      lastUpdateTime: "2024-05-16T18:42:14Z",
      message: "ReplicaSet \"my-deployment-6b7d4f4794\" has successfully progressed.",
      reason: "NewReplicaSetAvailable",
      status: "True",
      type: "Progressing",
    },
  ],
```

`CrashLoopBackOff` status is surely can be find as a Pod status - so we probably should enumerate the available Pods statuses in order to detect its presence, shouldn't we?

@datho7561 WDYT?

datho7561:
I think you might need to list the pods in order to collect this information.

PR #4148, opened 2024-05-22
Display status information for Deployments in Application Explorer tree #3823

vrubezhny:
Fixes: #3823


vrubezhny:
With the latest PR update we show:
-  Deployment name,
- Number of Available Replicas,
- Actual number of created Replicas, 
- Detected `CrashLoopBackOff` state,
- Pod Container Failures (if any)
- Deployment Failures (if any)

The Deployment element icon has:
- `Green` modifier when all the Pods are successfully started and running,
- `Yellow` modifier when no `CrashLoopBackOff` state is detected, but not all the Pods are successfully started and running
- `Red` modified when either `CrashLoopBackOff` state is detected or at least one of the pods has 'Failed` or `Unknown`  state

image

src/explorer.ts
```diff
@@ -270,24 +275,142 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
+                    messages.push(message);
+                }
+            });
+
+        return messages;
+    }
+
+    private async getDeploymentItem(element): Promise<TreeItem> {
+        const shouldHaveReplicas = element.spec.replicas > 0;
+        const availableReplicas = element.status.availableReplicas ? element.status.availableReplicas :
```
datho7561, line 358: From my understanding the two numbers here would best be labelled "desiredReplicas" and "actualReplicas". "desiredReplicas" should be take from the `spec` section, and "actualReplicas" should be taken from the `status` section. Does this make sense to you? If not I might misunderstand something.

I think the difference between `status.availableReplicas` vs `status.readyReplicas` vs `status.replicas` is one means it's in the process of being created, one means it's has been created and is running, and one means it's been created and running and has a [probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) configured that says it's healthy.

datho7561 (changes requested):
Overall working great! Saw a few things in the code. I haven't been able to get the yellow status to show up, but it might be because it would only show up very briefly as we discussed.

(sorry it took so long to get to this).

vrubezhny:
> I haven't been able to get the yellow status to show up, but it might be because it would only show up very briefly as we discussed.

That's simple: Just deploy your `pet-alpine` image to the Sandbox cluster - we've already discussed this - the `CrashLoopBackOff` doesn't appear on Sandbox because the Pod cannot be created, the `CreateContainerError`  is reported instead and the Pod is reported as `Pending`. So it makes it "yellow".

image

src/explorer.ts
```diff
@@ -270,24 +275,142 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
+        pods.forEach((pod) => {
+            pod.status?.containerStatuses &&
+            pod.status.containerStatuses.forEach((cs) => {
+                if (cs.state?.waiting) {
+                    const reason = cs.state.waiting.reason;
+                    const message = cs.state.waiting.message;
+
+                    inCrashLoopBackOff = inCrashLoopBackOff || reason === 'CrashLoopBackOff';
+
+                    const msg = `${reason}: ${message ? message.trim(): 'No any valuable message'}`;
```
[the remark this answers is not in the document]
vrubezhny: ACK

src/explorer.ts
```diff
@@ -270,24 +275,142 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
+            inCrashLoopBackOff,
+            messages
+        }
+    }
+
+    private collectDeploymentErrors(deployment): string[] {
+        const messages: string[] = [];
+        deployment.status.conditions.filter((c) => c.status === 'False')
+            .forEach((c) => {
+                const message = `${c.reason}: ${c.message ? c.message.trim(): 'No any valuable message'}`;
```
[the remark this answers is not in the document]
vrubezhny: ACK

src/explorer.ts
vrubezhny: So, OK. I take `desiredReplicas = element.spec.replicas` and don't try searching for `element.status.availableReplicas`

For the `actualReplicas` I take `element.status.replicas` value or 0 if it doesn't exist (which is possible, f.i., when scaled to 0).

But `element.status.replicas` tells nothing on real state of the replicas, as for a failing `pet-alpine` deployment (on Sandbox) we have:
```
status: {
  ...
  replicas: 1, 
  unavailableReplicas: 1, 
  updatedReplicas: 1
  ...
}
```

So, we actually have ONE replica and this the only replica is UNAVAILABLE (not working)...

... but we'll still show a tooltip like:

```
deployment: my-pet-alpine
desiredReplicas: 1
actualReplicas: 1
...
```
So user will think that everything should be OK with this replica - it's started and should be `Running`, while it's actually not (here it's `Pending`).

Probably we should show more replicas-related counters then, shouldn't we?

datho7561: I like this behaviour!

datho7561 (approved):
Looks good to me. Very informative of the deployment's state. Thanks, Victor!

PR #4160, opened 2024-05-27
Fix create serverless function view

### LINKED ISSUE #4092: Trying to create a serverless function using the Spring Boot template causes the webview to error
... and the error boundary doesn't work, so the user gets no feedback as to what is happening (they get a blank screen).

datho7561:
- Split out the "invoke serverless function" webview from the "create serverless function" webview
- For "create serverless function" and "invoke serverless function", wait on receiving a message from the webview before sending data to it
- Fixing theming in the "invoke serverless function" webview

Fixes #4092


datho7561:
The root cause for this bug is the same as for the Helm Chart browser: we need confirmation from the webview that it's ready before we can start sending data to it.

msivasubramaniaan (approved):
Lgtm

vrubezhny (approved):
Looks good to me. Thanks!

Serverless Function Create and Invoke views work fine for me on Linux

PR #4166, opened 2024-05-29
Trim token value from Clipboard

vrubezhny:
Also the sha256-token validation is fixed.

When using 'Developer Sandbox` to login to the DevSandbox and a token is already stored in the Clipboard, its value isn't trimmed before its validation and usage:
 
image

So, when the sha256-token value begins with a whitespace character the  `Login to DevSandbox` button is shown as disabled and its tooltip says: "Token in clibboard is invalid....":

image

Also, the sha256 token validation RegEx incorrectly validates provided token - it doesn't accept `-` character as a part of token value and doesn't control the ending thus allowing a whitespaces or any text to be placed after  a whitespace or '-' character.

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -1027,7 +1027,7 @@ export class Cluster extends OpenShiftItem {
     }
 
     static validateLoginToken(token: string): boolean {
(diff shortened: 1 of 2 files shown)


datho7561 (approved):
Looks good to me! Thanks, Victor!

PR #4170, opened 2024-05-30
Dedupe Deployment error messages in tooltip

vrubezhny:
The PR deduplicates the same Deployment error messages, showing only 1 message for every problem reason.
Also, it limits the number of shown error messages for Deployment and Pod Containers to 3 messages each.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -324,8 +324,8 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
                     inCrashLoopBackOff = inCrashLoopBackOff || reason === 'CrashLoopBackOff';
 
                     const msg = `${reason}: ${message ? message.trim(): 'No valuable message'}`;
-                    // Skip duplicates and show not more than 10 errors
-                    if (messages.length <= 10 && !(messages.find((m) => m === msg))) {
+                    // Skip duplicated messages
+                    if (messages.length < 3 && !(messages.find((m) => m.startsWith(`${reason}:`)))) {
                         messages.push(msg);
                     }
                 }
@@ -344,8 +344,8 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
             .forEach((c) => {
                 const message = `${c.reason}: ${c.message ? c.message.trim(): 'No valuable message'}`;
 
-                // Skip duplicates and show not more than 10 errors
-                if (messages.length <= 10 && !(messages.find((m) => m === message))) {
+                // Skip duplicated messages
+                if (messages.length < 3 && !(messages.find((m) => m.startsWith(`${c.reason}:`)))) {
                     messages.push(message);
                 }
             });


vrubezhny:
For example, the same errors from all 5 Pod Containers were deduped into the single `CreateContainerError` message in tooltip:
- Before: 
image

-After:
image

src/explorer.ts
```diff
@@ -319,13 +323,13 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
             pod.status.containerStatuses.forEach((cs) => {
                 if (cs.state?.waiting) {
                     const reason = cs.state.waiting.reason;
-                    const message = cs.state.waiting.message;
+                    const message = this.stripTimestamp(cs.state.waiting.message);
 
                     inCrashLoopBackOff = inCrashLoopBackOff || reason === 'CrashLoopBackOff';
 
                     const msg = `${reason}: ${message ? message.trim(): 'No valuable message'}`;
                     // Skip duplicates and show not more than 10 errors
```
datho7561, line 331: Update or remove this comment if it's stale

src/explorer.ts
```diff
@@ -342,10 +346,10 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
         const messages: string[] = [];
         deployment.status.conditions.filter((c) => c.status === 'False')
             .forEach((c) => {
-                const message = `${c.reason}: ${c.message ? c.message.trim(): 'No valuable message'}`;
+                const message = `${c.reason}: ${c.message ? this.stripTimestamp(c.message.trim()): 'No valuable message'}`;
 
                 // Skip duplicates and show not more than 10 errors
```
datho7561, line 351: Same here

src/explorer.ts
```diff
@@ -319,13 +323,13 @@ export class OpenShiftExplorer implements TreeDataProvider<ExplorerItem>, Dispos
...
                     const reason = cs.state.waiting.reason;
-                    const message = cs.state.waiting.message;
+                    const message = this.stripTimestamp(cs.state.waiting.message);
 
                     inCrashLoopBackOff = inCrashLoopBackOff || reason === 'CrashLoopBackOff';
 
                     const msg = `${reason}: ${message ? message.trim(): 'No valuable message'}`;
                     // Skip duplicates and show not more than 10 errors
-                    if (messages.length <= 10 && !(messages.find((m) => m === msg))) {
+                    if (messages.length < 3 && !(messages.find((m) => m === msg))) {
```
datho7561, line 332: Another issue with checking if the messages are the same is that the pod-related messages often contain the pod's name, which has a bunch of random numbers and letters specific to the pod at the end (eg. `-64569c9d5-4mjxq`)

datho7561 (changes requested):

src/explorer.ts
vrubezhny: ops.

src/explorer.ts
vrubezhny: fixed

src/explorer.ts
vrubezhny: I'm not sure if we need to strip and dedupe there added id at the end,  but yeah, in case of `CrashLoopBackOff`:
```
CrashLoopBackOff: back-off 20s restarting failed container=crash-hello-server pod=crash-hello-server-955cdcdb7-fnkhr_vrubezhny-dev(aba0681f-b0ed-4342-9c17-bfdd1b42f23b)
 ```
- back-off time is reported (so in theory the messages with the same Pod will be repeated with different back-off times)
- different pods has different IDs which multiply the displayed data. 
But I'm not sure if the Pod IDs are to be stripped: Since we do not show any for reach Pod (by icon/tooltip etc) - the Deployment tooltip currently is the only way to see which exact Pod is failing.

WDYT?

datho7561: I am okay with not stripping the container the ids, but it might be worthwhile removing the `==` check, since we know it will probably not ever be the case that the messages are identical.

vrubezhny: Actualy, we can probably filter by 'reason' (`CrashLoopBackOff`, `MinimumReplicasUnavailable`, `CreateContainerError`, etc), showing the only first appearing message for each reason kind...
We're limiting errors to only 3 messages, so at least in this case we'll show for sure the failing Pod/Container ID, back-off delay, time etc. so users will have some idea on where to start digging the reason of failure.

vrubezhny: > ... but it might be worthwhile removing the == check, since we know it will probably not ever be the case that the messages are identical.

Why? if we continue stripping some parts of the error messages we'll highly likely meet some messages with equal text after stripping

datho7561: I guess so

datho7561 (dismissed):
Seems good to me; this makes the error log significantly more readable

Issue #4118, opened 2024-05-11
Migrate Kubernetes Resource Link Provider 


vrubezhny:
The Kubernetes Resource Link provider adds possibility to browse Deployment/Pod configurations YAML documents and open some related configs by control-clicking (currently it works in K8s extension only when a configuration document is opened using YAML output format):

[Screencast from 2024-05-11 17-10-20.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/db7e852d-9f6e-4db7-b87e-1bc9449030c0)

PR #4171, opened 2024-05-31
Migrate Kubernetes Resource Link Provider #4118

vrubezhny:
Fixes: #4118


vrubezhny:
Kubernetes Resource Link Provider is added:
- Output Format is set to `yaml`:

[Screencast from 2024-06-03 22-01-23.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/a663b852-ed32-4c15-bce3-85eb04cb1881)

- Output Format is set to 'json`:

[Screencast from 2024-06-03 22-04-03.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/947ad80e-2a7f-4539-bd45-7cecdd7b09a6)

src/k8s/vfs/json-locator.ts
```diff
@@ -0,0 +1,248 @@
+/*-----------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE file in the project root for license information.
+ *-----------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { MappingItem, Node, NodeProvider } from './locator-util';
+
+const parse = require('json-to-ast');
```
datho7561, line 9: please use `import` instead of require, it should work just as well

package.json
```diff
@@ -157,6 +159,7 @@
 		"leasot": "^13.3.0",
 		"mocha": "^10.4.0",
 		"mocha-jenkins-reporter": "^0.4.7",
+		"node-yaml-parser": "^0.0.9",
```
datho7561, line 162: The document links don't work when I install from .vsix . I think you will need to list this as a `dependency` instead of a `devDependency` in order to get it to work.

datho7561 (changes requested):
Seems to be working well at first glance, but I'll try to take a more thorough glance at it tomorrow.

src/k8s/vfs/json-locator.ts
vrubezhny: When using import it couldn't be built for some reason. I'll check it again.

package.json
vrubezhny: Sure. You're right!

vrubezhny:
@datho7561 I've addressed the first comments - moved `node-yaml-parser` to `dependencies` and used `import` for `json-to-ast` parser - so now the VSIX should work as expected.

datho7561 (approved):
Works well. Thanks, Victor!

Issue #4165, opened 2024-05-29
Use the value saved in the clipboard when logging in using a token


vrubezhny:
When logging into a cluster using a token, use the clipboard value if it represents a valid `sha256` token.

PR #4179, opened 2024-06-06
Use the value saved in the clipboard when logging in using a token #4165

vrubezhny:
Fixes: #4165

### DIFF
--- a/src/openshift/cluster.ts
+++ b/src/openshift/cluster.ts
@@ -935,13 +935,11 @@ export class Cluster extends OpenShiftItem {
     }
 
     static async readFromClipboard(): Promise<string> {
-        let r = '';
         try {
-            r = await env.clipboard.readText();
+            return await env.clipboard.readText();
         } catch (ignore) {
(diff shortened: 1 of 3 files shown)


vrubezhny:
When entering a token, its value is initialized from the Clipboard if it contains a valid sha256 token.
The following buttons are added to the Input Field:
- "Paste from Clibpoard" - allows pasting a value from the Clipboard (the value gets validated before paste, so this allows to paste the only the values that successfully pass the validation)
- "Show value"/"Hide value" - allows to make visible/hidden a "password"-like value (is shown only if entering "password"-like values (passwords, tokens, etc))

[Screencast from 2024-06-06 01-52-33.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/7a74c8aa-7463-4a2e-b95d-a7fa6a5b00b3)

datho7561 (approved):
Looks good and it works like a charm. Thanks, Victor!

Issue #4183, opened 2024-06-06
'ODO' stopped working after Devfile schema update to v.2.2.2


vrubezhny:
'ODO' binary stopped working when Devfile schema v.2.2.2 is used:

```
$ mkdir starter
$ cd starter/
$ ~/projects/vscode-openshift-tools/source/vscode-openshift-tools/out/tools/linux/odo init --name=component1 --devfile=nodejs --starter=nodejs-starter
  __
 /  \__     Initializing a new component
 \__/  \    
 /  \__/    odo version: v3.15.0 (10b5e8a8f)
 \__/

 ✓  Downloading devfile "nodejs" [1s]
 ✗  unable to parse devfile: failed to populateAndParseDevfile: unable to find schema for version "2.2.2". The parser supports devfile schema for version 2.0.0, 2.1.0, 2.2.0, v1alpha2
the command failed, the devfile has been removed from current directory
```

I believe this requires 'ODO' to be updated to support this Devfile schema version.

datho7561:
Please consider filing upstream in https://github.com/redhat-developer/odo/issues

vrubezhny:
In order to fix the integration test failures we can limit the Devfile version to <=2.2.0 for the node.js component. Here is the workaround: https://github.com/redhat-developer/vscode-openshift-tools/pull/4185

At the moment, in order to fix the issue,  we can use nightly builds of 'odo' where the problem looks to be fixed is fixed instead of the officially released version. See: https://odo.dev/docs/overview/installation#nightly-builds

PR #4185, opened 2024-06-06
A workaround for 'ODO' stopped working after Devfile schema update to v.2.2.2 #4183

vrubezhny:
This is the only workaround limiting the devfile version for node.js component to v.2.2.0 when running the integration tests

Issue: #4183

### DIFF
--- a/src/odo/odoWrapper.ts
+++ b/src/odo/odoWrapper.ts
@@ -109,6 +109,7 @@ export class Odo {
         starter: string = undefined,
         useExistingDevfile = false,
         customDevfilePath = '',
+        devfileVersion?: string
     ): Promise<void> {
         await this.execute(
             Command.createLocalComponent(
@@ -119,6 +120,7 @@ export class Odo {
                 starter,
                 useExistingDevfile,
                 customDevfilePath,
+                devfileVersion
             ),
             location.fsPath,
         );
--- a/test/integration/odoWrapper.test.ts
+++ b/test/integration/odoWrapper.test.ts
@@ -66,6 +66,9 @@ suite('./odo/odoWrapper.ts', function () {
                 'component1',
                 tmpFolder1,
                 'nodejs-starter',
+                false,
+                '',
+                '2.2.0' // A workaround for odo v.3.15.0 not supporting schema v.>2.2.0
             );
             await Odo.Instance.createComponentFromFolder(
                 'go',
@@ -258,6 +261,9 @@ suite('./odo/odoWrapper.ts', function () {
                 'component1',
                 Uri.parse(componentFolder),
                 'nodejs-starter',
+                false,
+                '',
+                '2.2.0' // A workaround for odo v.3.15.0 not supporting schema v.>2.2.0
             );
         });
 


src/odo/odoWrapper.ts
```diff
@@ -109,6 +109,7 @@ export class Odo {
         starter: string = undefined,
         useExistingDevfile = false,
         customDevfilePath = '',
+        devfileVersion?: string
```
mohitsuman, line 112: how does it know which version of devfile it should not update to? we need to define it somewhere.

mohitsuman (commented):
devfile version info ?

src/odo/odoWrapper.ts
vrubezhny: This is only a workaround to make the test working with current officially released 'odo' version  and the Default Devfile Registry.

The officially released 'odo' supports devfile schema versions `<= 2.2.0`. So this version is provided as the value of this argument hereafter in the test when creating a component from 'node.js' devfile (because this devfile has multiple versions and [if not specifying this argument] odo returns the default one, which currently is set to be `> 2.2.0.`

We do not provide this argument value when creating a component with 'Go` devfile - its devfile returned by default has  schema version `< 2.2.0` so it perfectly works with odo.

vrubezhny:
> devfile version info ?

Default Devfile registry contains multiple versions for `Node.js` devfile, by default `odo` retrieves the devfile version marked as 'default` which schema version is > `2.2.0` - this make a problem for 'odo' that is officially released.

In order to fix tests we force 'odo' to retrieve a devfile with schema version `=== 2.2.0` instead of the default one, which makes 'odo' happy when it parses this schema.

There is a proposal to use nightly builds for 'odo' binary instead of official release (see: https://github.com/redhat-developer/vscode-openshift-tools/issues/4183#issuecomment-2153624376) - the devfile schema version issue is fixed on   the main 'odo' branch and the fixed version is available at https://odo.dev/docs/overview/installation#nightly-builds

datho7561:
I still don't see the Nodejs devfile listed in the devfile registry viewer, so I don't think this will do much. (eg. you can't choose to create a project based on the nodejs devfile, so it doesn't make sense to reduce the version of that devfile that's used.)

vrubezhny:
> I still don't see the Nodejs devfile listed in the devfile registry viewer, so I don't think this will do much. (eg. you can't choose to create a project based on the nodejs devfile, so it doesn't make sense to reduce the version of that devfile that's used.)

This is a quite strange thing... as for the Default Devfile Registry:

- Web site https://registry.devfile.io/viewer shows 32 entries
- `$ odo registry --devfile-registry DefaultDevfileRegistry | wc` encounters 26 entries

... while the Devfile Registry editor shows only 11 entries.

Also, the result of `odo registry --devfle-registry DefaultDevfileRegistry` is a bit different from what we have on the web site, for instance the Web site contains devfile with `name: wildfly` while the command returns the same (or similar) defvile with `name: java-wildfly`...  Looks like they actually reference the different registries or the output is modified somehow.

vrubezhny:
We probably won't need this PR if we move to using ODO nightly builds.

See:  PR #4191 - [[build] Use ODO binary from nightly builds #4191](https://github.com/redhat-developer/vscode-openshift-tools/pull/4191)

mohitsuman:
@vrubezhny if there is a new release of odo binary, that should fix the issue?

vrubezhny:
> @vrubezhny if there is a new release of odo binary, that should fix the issue?

That's not a release (no more releases are produced unless they have a blocker issue), but nightly builds made from the ODO repository `main` branch that currently contains quite a lot of dependency updates. And the support for Devfile schema v.2.2.2 is included as well. 
I've tested the current ODO nightly build:  3.15.0 (6d83e4043-nightly)  - it creates a component from 'nodejs' devfile (schema v.2.2.2) with no problem:

```
$ odo version
odo v3.15.0 (6d83e4043-nightly)

Server: https://127.0.0.1:42801
Kubernetes: v1.27.3
Podman Client: 4.9.4

$ odo init --name=component1 --devfile=nodejs --starter=nodejs-starter
  __
 /  \__     Initializing a new component
 \__/  \    
 /  \__/    odo version: v3.15.0 (6d83e4043-nightly)
 \__/

 ✓  Downloading devfile "nodejs" [4s]
 ✓  Downloading starter project "nodejs-starter" [515ms]

Your new component 'component1' is ready in the current directory.
To start editing your component, use 'odo dev' and open this folder in your favorite IDE.
Changes will be directly reflected on the cluster.

$ cat devfile.yaml | grep schemaVersion:
schemaVersion: 2.2.2
```

vrubezhny:
> @vrubezhny if there is a new release of odo binary, that should fix the issue?

I probably got wrong the question...

If ODO team makes a new release - then the problem will definitely be solved with no any changes from our side.

Using nightly builds looks now as a bed option, as the download URL is static and the track of nightly build binaries isn't kept - this really prevents us from using them.

datho7561 (approved):
This fixes the integration test suite for the time being, which I think is very helpful.

@vrubezhny is this ready for merge?

PR #4191, opened 2024-06-08
[build] Use ODO binary from nightly builds

vrubezhny:
Since the [ODO project](https://github.com/redhat-developer/odo) has stoped producing new releases for ODO binary, the [nightly builds](https://odo.dev/docs/overview/installation/#nightly-builds) became the source of most recent and up-to-date ODO binary version.

Fixes: #4183

### DIFF
--- a/.github/workflows/check-odo.yml
+++ b/.github/workflows/check-odo.yml
@@ -4,42 +4,111 @@ on:
   schedule:
     - cron: "0 8 * * *"
   workflow_dispatch:
+env:
+  TOOL_REPO: redhat-developer/odo
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 jobs:
-  check-odo-repo:
+  check-odo:
     runs-on: ubuntu-latest
-    env:
-        TOOL_REPO: redhat-developer/odo
-        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    outputs:
+      last-commit-id: ${{ steps.repo-last-commit-info.outputs.repo-last-commit-id }}
+      last-release-tag: ${{ steps.repo-last-commit-info.outputs.repo-last-release-tag }}
+      last-release-version: ${{ steps.repo-last-commit-info.outputs.repo-last-release-version }}
+      nightly-build-version: ${{ steps.repo-last-commit-info.outputs.repo-nightly-build-version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: '${{ env.TOOL_REPO }}'
+          fetch-depth: 2
+          fetch-tags: false
+          path: redhat-developer-odo-repository
+      - name: Get Last Commit Info
+        id: repo-last-commit-info
+        run: |
+          pushd redhat-developer-odo-repository
+
+          # Last Commit ID (short or abbreviated to 9 characters version)
+          lastCommitId="$(git describe --no-match --always --abbrev=9 --dirty --broken 2>/dev/null || git rev-parse --short HEAD 2>/dev/null)"
+          echo "repo-last-commit-id=$lastCommitId" >> $GITHUB_OUTPUT
+
+          # Last Release Tag and Version
+          lastReleaseTag=$(gh release --repo ${{ env.TOOL_REPO }} view --json tagName | jq -r .tagName)
+          lastReleaseVersion="$(echo $lastReleaseTag | sed 's|v||')"
+          echo "repo-last-release-tag=$lastReleaseTag" >> $GITHUB_OUTPUT
+          echo "repo-last-release-version=$lastReleaseVersion" >> $GITHUB_OUTPUT
+
+          # Nightly Build Version
+          echo "repo-nightly-build-version=$lastReleaseVersion (${lastCommitId}-nightly)" >> $GITHUB_OUTPUT
+
+          popd
+
+  check-pr:
+    runs-on: ubuntu-latest
+    outputs:
+      pr-exists: ${{ steps.check-pr-exists.outputs.pr-exists }}
+    needs: check-odo
+    if: ${{ needs.check-odo.outputs.last-commit-id != '' }}
+    steps:
+      - name: Check if the update PR already exists
+        id: check-pr-exists
+        run: echo pr-exists=$(gh pr --repo ${{ github.repository }} list --state all --search "update odo cli to ${{ needs.check-odo.outputs.last-release-version }} (${{ needs.check-odo.outputs.last-commit-id }}-nightly) in:title" --json url | jq length) >> $GITHUB_OUTPUT
+
+  check-nightly-build:
+    runs-on: ubuntu-latest
+    outputs:
+      nightly-build-exists: ${{ steps.check-nightly-build-exists.outputs.nightly-build-exists }}
+    needs: [check-odo, check-pr]
+    if: ${{ needs.check-odo.outputs.last-commit-id != '' && needs.check-pr.outputs.pr-exists == 0 }}
+    steps:
+      - name: Check if a Nightly Build exists for the last commit id
+        id: check-nightly-build-exists
+        run: echo nightly-build-exists=`curl -s "https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-nightly-builds/odo-linux-arm64-${{ needs.check-odo.outputs.last-commit-id }}.tar.gz.sha256" | grep -E "^[A-Za-z0-9]+$" | wc -w` >> $GITHUB_OUTPUT
+
+  update-odo:
+    runs-on: ubuntu-latest
+    needs: [check-odo, check-pr, check-nightly-build]
+    if: ${{ needs.check-odo.outputs.last-commit-id != '' && needs.check-pr.outputs.pr-exists == 0 && needs.check-nightly-build.outputs.nightly-build-exists != 0 }}
     steps:
       - name: Check Out Code
         uses: actions/checkout@v4
-      - name: Get latest ODO version
-        run: |
-          echo "REPO_ODO_VERSION=$(cat src/tools.json | jq -r .odo.version)" >> $GITHUB_ENV
-          LATEST_TOOL_RELEASE_RESP=$(gh release --repo ${{ env.TOOL_REPO }} view --json tagName,url)
-          echo "LATEST_TOOL_RELEASE=$(echo ${LATEST_TOOL_RELEASE_RESP} | jq -r .tagName | sed 's|v||')" >> $GITHUB_ENV
-          echo "LATEST_TOOL_URL=$(echo ${LATEST_TOOL_RELEASE_RESP} | jq -r .url)" >> $GITHUB_ENV
-      - name: Find existing PR for ODO version
-        run: |
-          echo PR_EXISTS=$(gh pr --repo ${{ github.repository }} list --state all --search "update odo ${{env.LATEST_TOOL_RELEASE}} in:title" --json url | jq length) >> $GITHUB_ENV
       - name: Update src/tools.json with latest odo version
-        if: ${{ (env.LATEST_TOOL_RELEASE != '') && (env.LATEST_TOOL_RELEASE != env.REPO_ODO_VERSION) && (env.PR_EXISTS == 0) }}
         run: |
-          jq --indent 4 '.odo.version = "${{ env.LATEST_TOOL_RELEASE }}"' src/tools.json | jq --indent 4 '.odo.versionRange = "^${{ env.LATEST_TOOL_RELEASE }}"' | jq --indent 4 '.odo.versionRangeLabel = "version >= ${{ env.LATEST_TOOL_RELEASE }}"' > src/tools.json.new
+          jq --indent 4 '.odo.description = "ODO CLI tool"' src/tools.json \
+            | jq --indent 4 '.odo.vendor = "Red Hat Developer"' \
+            | jq --indent 4 '.odo.name = "odo"' \
+            | jq --indent 4 '.odo.version = "${{ needs.check-odo.outputs.last-release-version }}"' \
+            | jq --indent 4 '.odo.versionRange = "^${{ needs.check-odo.outputs.last-release-version }}"' \
+            | jq --indent 4 '.odo.versionRangeLabel = "version >= ${{ needs.check-odo.outputs.last-release-version }}"' > src/tools.json.new
           mv src/tools.json.new src/tools.json
           for platform in win32 darwin darwin-arm64 linux linux-arm64; do
-            old_url=`jq -r .odo.platform[\"${platform}\"].url src/tools.json`
-            new_url=`echo ${old_url} | sed "s|${{ env.REPO_ODO_VERSION }}|${{ env.LATEST_TOOL_RELEASE }}|"`
+            pltfrm="$platform"
+            ext=".tar.gz"
+            exeExt=""
+            if [[ "$platform" == "win"* ]]; then
+              pltfrm="windows"
+              ext=".zip"
+              exeExt=".exe"
+            fi
+            arch="-amd64"
+            if [[ $platform == *"-a"* ]]; then
+              arch="" # already in platform string
+            fi
+            new_url="https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-nightly-builds/odo-${pltfrm}${arch}-${{ needs.check-odo.outputs.last-commit-id }}${ext}"
             checksum=`curl -s ${new_url}.sha256`
-            jq --indent 4 ".odo.platform[\"${platform}\"].url = \"${new_url}\"" src/tools.json | jq --indent 4 ".odo.platform[\"${platform}\"].sha256sum = \"${checksum}\"" > src/tools.json.new
+            dlFileName="odo-${pltfrm}${arch}${ext}"
+            cmdFileName="odo-${pltfrm}${arch}-${{ needs.check-odo.outputs.last-commit-id }}${exeExt}"
+            jq --indent 4 ".odo.platform[\"${platform}\"].url = \"${new_url}\"" src/tools.json \
+              | jq --indent 4 ".odo.platform[\"${platform}\"].sha256sum = \"${checksum}\"" \
+              | jq --indent 4 ".odo.platform[\"${platform}\"].dlFileName = \"${dlFileName}\"" \
+              | jq --indent 4 ".odo.platform[\"${platform}\"].cmdFileName = \"${cmdFileName}\"" > src/tools.json.new > src/tools.json.new
             mv src/tools.json.new src/tools.json
           done
       - name: Create pull request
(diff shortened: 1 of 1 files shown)


vrubezhny:
In case we move to using ODO nightly builds, we do not need limiting the version of 'nodejs' devfile (or potentially of any other component devfiles), so we should close [A workaround for 'ODO' stopped working after Devfile schema update to v.2.2.2 #4183 #4185](https://github.com/redhat-developer/vscode-openshift-tools/pull/4185) as not needed anymore

datho7561 (approved):
Seems good to me. If something breaks when it runs, we can always fix it or revert it. However, I'm also interested in Roland's review, since he's much more experienced in shell scripting than I am.

.github/workflows/check-odo.yml
```diff
@@ -13,33 +13,59 @@ jobs:
...
+          echo "LATEST_TOOL_RELEASE_TAG=${latestReleaseTag}" >> $GITHUB_ENV
+          echo "LATEST_TOOL_RELEASE_VERSION=$(echo ${latestReleaseTag} | sed 's|v||')" >> $GITHUB_ENV
+          commitId="$(git ls-remote --refs --heads https://github.com/${{ env.TOOL_REPO }} refs/heads/main 2>/dev/null | cut -c1-9 2>/dev/null)"
+          echo "LATEST_TOOL_COMMIT_ID=${commitId}" >> $GITHUB_ENV
+          echo "LATEST_TOOL_NIGHTLY_BUILD=${commitId}-nightly" >> $GITHUB_ENV
       - name: Find existing PR for ODO version
         run: |
-          echo PR_EXISTS=$(gh pr --repo ${{ github.repository }} list --state all --search "update odo ${{env.LATEST_TOOL_RELEASE}} in:title" --json url | jq length) >> $GITHUB_ENV
+          latestReleaseVersionString="${{ env.LATEST_TOOL_RELEASE_VERSION }} (${{ env.LATEST_TOOL_NIGHTLY_BUILD }})"
+          echo "LATEST_TOOL_RELEASE_VERSION_STRING=${latestReleaseVersionString}" >> $GITHUB_ENV
```
rgrunber, line 29: Same thing here. We can just set `LATEST_TOOL_RELEASE_VERSION_STRING` directly.

I though it was weird to include the previous release version in the string that includes that this is a nightly release, but apparently this is what the `odo` nightly binary actually does :

```
$ ./odo-linux-amd64 version
===============================================
⚠  unable to fetch the cluster server version
===============================================
odo v3.15.0 (6d83e4043-nightly)

Podman Client: 4.9.4
```

.github/workflows/check-odo.yml
```diff
@@ -13,33 +13,59 @@ jobs:
...
+          echo PR_EXISTS=$(gh pr --repo ${{ github.repository }} list --state open --search "update odo cli to ${latestReleaseVersionString} in:title" --json url | jq length) >> $GITHUB_ENV
```
rgrunber, line 30: What if the PR gets closed, wouldn't this create another PR ? I think that's why we check `all`.

.github/workflows/check-odo.yml
```diff
@@ -13,33 +13,59 @@ jobs:
...
       - name: Update src/tools.json with latest odo version
-        if: ${{ (env.LATEST_TOOL_RELEASE != '') && (env.LATEST_TOOL_RELEASE != env.REPO_ODO_VERSION) && (env.PR_EXISTS == 0) }}
+        if: ${{ (env.LATEST_TOOL_NIGHTLY_BUILD != '') && (env.LATEST_TOOL_RELEASE_VERSION_STRING != env.REPO_ODO_VERSION) && (env.PR_EXISTS == 0) }}
         run: |
-          jq --indent 4 '.odo.version = "${{ env.LATEST_TOOL_RELEASE }}"' src/tools.json | jq --indent 4 '.odo.versionRange = "^${{ env.LATEST_TOOL_RELEASE }}"' | jq --indent 4 '.odo.versionRangeLabel = "version >= ${{ env.LATEST_TOOL_RELEASE }}"' > src/tools.json.new
+          jq --indent 4 '.odo.version = "${{env.LATEST_TOOL_RELEASE_VERSION_STRING}}"' src/tools.json \
+           | jq --indent 4 '.odo.versionRange = "${{env.LATEST_TOOL_RELEASE_VERSION_STRING}}"' \
+           | jq --indent 4 '.odo.versionRangeLabel = "version == ${{env.LATEST_TOOL_RELEASE_VERSION_STRING}}"' > src/tools.json.new
```
rgrunber, line 36: I don't think `semver` can handle strings of the form : `3.15.0 (6d83e4043-nightly)`. I think we need to keep this as 3.15.0.

rgrunber (commented):
Looks ok, though I haven't tested if it works in a fork yet.

.github/workflows/check-odo.yml
```diff
@@ -13,33 +13,59 @@ jobs:
...
         uses: actions/checkout@v4
-      - name: Get latest ODO version
+      - name: Get latest ODO nightly build version
         run: |
           echo "REPO_ODO_VERSION=$(cat src/tools.json | jq -r .odo.version)" >> $GITHUB_ENV
-          LATEST_TOOL_RELEASE_RESP=$(gh release --repo ${{ env.TOOL_REPO }} view --json tagName,url)
-          echo "LATEST_TOOL_RELEASE=$(echo ${LATEST_TOOL_RELEASE_RESP} | jq -r .tagName | sed 's|v||')" >> $GITHUB_ENV
-          echo "LATEST_TOOL_URL=$(echo ${LATEST_TOOL_RELEASE_RESP} | jq -r .url)" >> $GITHUB_ENV
+          LATEST_TOOL_RELEASE_RESP=$(gh release --repo ${{ env.TOOL_REPO }} view --json tagName)
+          latestReleaseTag="$(echo ${LATEST_TOOL_RELEASE_RESP} | jq -r .tagName)"
```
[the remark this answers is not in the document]
vrubezhny: That's possible, but will require to invoke `jq -r .tagName` repeatedly on the same output  - first to get the tag value, then to make a version value from it.  I just dislike adding extra binary executions.

.github/workflows/check-odo.yml
vrubezhny: Actually it's not the "previous" it's the current release version.  
You may think of nightly build as a Service Release, so the version stays the same, just with a few bug fixes added.

Also if I don't include it - the version will look like "6d83e4043-nightly" which is hard to associate with the real  ODO version.

.github/workflows/check-odo.yml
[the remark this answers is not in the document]
rgrunber: Ok, then let's just leave it.

.github/workflows/check-odo.yml
vrubezhny: Yes. this will create a new PR in eight hours (if nothing changes in the repository). 

OK. I'll change it back to `all`

.github/workflows/check-odo.yml
```diff
@@ -13,33 +13,59 @@ jobs:
...
+          echo "LATEST_TOOL_RELEASE_TAG=${latestReleaseTag}" >> $GITHUB_ENV
+          echo "LATEST_TOOL_RELEASE_VERSION=$(echo ${latestReleaseTag} | sed 's|v||')" >> $GITHUB_ENV
+          commitId="$(git ls-remote --refs --heads https://github.com/${{ env.TOOL_REPO }} refs/heads/main 2>/dev/null | cut -c1-9 2>/dev/null)"
```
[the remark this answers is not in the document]
vrubezhny: We do not checkout any repo here... And we can't use any "local" `git` commands. 

I'm using only `git ls-remote <options> <repo_url> <params>` or `gh release --repo <repo-name> view <params>` - or any commands like  these (where you can specify a repo or URL to repo to get some info from), but not the local ones.

.github/workflows/check-odo.yml
vrubezhny: > It's just odd because most projects would adopt the convention of naming it (${lastReelase}+1) since that's what is actually being targetted.

So, having `3.15.0` released, do you mean I have to enumerate the following nightly builds as `3.15.1` (xxxxxxxx-nightly)?...

I don't thing it'll properly describe the version of nightly build, because when user executes such a nightly build he'll get: `odo v3.15.0 (6d83e4043-nightly)`, not `odo v3.15.1 (6d83e4043-nightly)`

.github/workflows/check-odo.yml
vrubezhny: I don't remember for sure, but I think we do not use semver with these `versionRange` and `versionRangeLabel` properties on with any others when installing the binaries

rgrunber: It looks like we do at https://github.com/redhat-developer/vscode-openshift-tools/blob/b80a77ac7a4dfbf11a55980e2eed463c7c11a022/src/tools.ts#L99 if you look up the chain.

vrubezhny: And I can't leave it just as `3.15.0` because otherwise it will not make any difference between builds. Neither construct it as `3.15.0. 6d83e4043` because `6d83e4043` is not an ordinal build number - it's a part of commit ID (a quite unique number, but semver won't able to work property with it)

Anyway, using `3.15.0 (6d83e4043-nightly)` in `version`, `versionRange` and `versionRangeLabel` do not result into any problems, as far as I know.

vrubezhny: > semver.satisfies(await ToolsConfig.getVersion(location), versionRange))

:disappointed:  You're right... But then I really cannot construct a valuable ordinal build number as version build number never changes for nightly builds, so  for this release it always stays `3.15.0` and even if I construct it as `previousReleaseVersion +1` (=== 3.15.1?)  it will fail updating it on the second nightly build created because  its 'constructed version number will be `3.15.1` too.

Need to think what can we do with the version/versionRange here

vrubezhny: In order to create a valuable and comparable label, I believe, it's enough to get a date/time of commit by its ID.
The problem is that I cannot get this info from a remote git repository (using `gh repo ... `  and `git ls-remote ...` commands.
Any thoughts?

.github/workflows/check-odo.yml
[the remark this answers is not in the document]
vrubezhny: This code is outdated. The new version uses `git rev-parse HEAD` as the odo repository is now cloned and can be acquired locally.

.github/workflows/check-odo.yml
vrubezhny: The nightly version is changed to be in `<Last Release Version + 0.0.1>-<Build Number>-nightly`, so, f.i., for the release `3.15.0` and the commit with ID `6d83e4043` the nightly version will be generated as `3.15.1-116-nightly`, which is fully `semver` compatible and when comparing:

`3.15.0` < `3.15.1-116-nightly` < `3.15.1`

.github/workflows/check-odo.yml
vrubezhny: The question with `semver.satisfies()` is resolved: now the nightly version is changed to be in <Last Release Version + 0.0.1>-<Build Number>-nightly, so, f.i., for the release 3.15.0 and the commit with ID 6d83e4043 the nightly version will be generated as 3.15.1-116-nightly, which is fully semver compatible and when comparing:

3.15.0 < 3.15.1-116-nightly < 3.15.1

.github/workflows/check-odo.yml
```diff
@@ -5,41 +5,131 @@ on:
...
+
+          # Last Commit Committer Date/Time
+          echo "repo-last-commit-date=$(git log -1 --oneline --pretty=format:%cd --date=format:%Y%m%d%H%M%S ${lastCommitId} 2>/dev/null)" >> "$GITHUB_OUTPUT"
+
+          # Last Release Tag
+          lastReleaseTag="$(git describe --tags --abbrev=0 2>/dev/null)"
+          echo "repo-last-release-tag=$lastReleaseTag" >> "$GITHUB_OUTPUT"
+
+          # Number of commits between the latest release and the Last Commit ID
+          echo "repo-last-commit-build-number=$(git log --oneline ${lastReleaseTag}..${lastCommitId} 2>/dev/null | wc -l)" >> "$GITHUB_OUTPUT"
```
rgrunber, line 40: Is this really necessary ? I would get rid of it.

.github/workflows/check-odo.yml
```diff
@@ -5,41 +5,131 @@ on:
...
+        id: repo-last-commit-info
+        run: |
+          pushd redhat-developer-odo-repository
+
+          # Last Commit ID
+          lastCommitId="$(git rev-parse HEAD 2>/dev/null)"
+          echo "repo-last-commit-id=$lastCommitId" >> "$GITHUB_OUTPUT"
+
+          # Last Commit Committer Date/Time
+          echo "repo-last-commit-date=$(git log -1 --oneline --pretty=format:%cd --date=format:%Y%m%d%H%M%S ${lastCommitId} 2>/dev/null)" >> "$GITHUB_OUTPUT"
```
rgrunber, line 33: I don't see this used anywhere so I'd get rid of it.

.github/workflows/check-odo.yml
```diff
@@ -5,41 +5,131 @@ on:
...
+          fi
+
+  update-odo:
+    runs-on: ubuntu-latest
+    needs: should-update-odo
+    if: ${{ needs.should-update-odo.outputs.odo-version-changed == 'true' }}
+    steps:
+      - name: Check Out Code
+        uses: actions/checkout@v4
+      - name: Check if the update PR already exists
```
rgrunber, line 91: I got this error when running the workflow after already creating a PR :

`a pull request for branch "bump-odo-3.15.1-116-nightly" into branch "main" already exists:`

So I think there's something wrong

rgrunber (commented):
Seems to work. I just think some parts are a bit more complicated than they need to be.

.github/workflows/check-odo.yml
vrubezhny: This is used to generate nightly build version: "3.15.**1**-**_116_**-nightly"  - here `116` is the value of `repo-last-commit-build-number` 

For each nightly build we need to create an unique version  which will `semver` compatible and, when sorting, is to be placed between the last release version and the [possible] next release version, so I've used this build number as a qualifier. 

As an alternative we can use the nightly build last commit creation date, but it's worse as some commits merged on top of the main branch may be created earlier that the previously build nightly - then in theory this might make a problem. So, I'd prefer keep this `repo-last-commit-build-number` value "aka build number qualifier" here

.github/workflows/check-odo.yml
vrubezhny: This is really unused at the moment... 
The purpose is is the same as for `repo-last-commit-build-number` - provide a valid, `semver` compatible version qualifier when constructing the version of a nightly build, so it's an alternative to using  `repo-last-commit-build-number` as a version qualifier for a nightly build.

If we use this value instead of `build number` a nightly build version will look like: "3.15.1-20240612215250-nightly` and also will be `semver` compatible...
...But it's worse than a version made with `build number` as some commits merged on top of the main branch may be created earlier that the previously build nightly - so sometimes this might make a problem.

So, I would use `repo-last-commit-build-number` for a nightly build version qualifier (as it is made now) and removed this variable.

.github/workflows/check-odo.yml
vrubezhny: Check please, you probably already have  this PR created (open or closed), no? (if this is not the first time you start the GH action)

vrubezhny: This is outdated now.

.github/workflows/check-odo.yml
vrubezhny: Outdated. NO more "nightly build number" nor "commit date" is used

.github/workflows/check-odo.yml
vrubezhny: Outdated. NO more "nightly build number" nor "commit date" is used

vrubezhny:
At the moment, the ODO Team looks like preparing a release for ODO v.3.16.0, so with or without this PR merged, we'll get that update. 

But for the future, we'll, probably, still need this PR - ODO Team has added commit numbers to the ODO download URLs - so since then we able to use nightly builds with no problem (as far as I understood, the archive and  sha checksum files history is kept, so we can get sha checksum for an archive, download the archive later and check it against the saved sha checksum)

.github/workflows/check-odo.yml
```diff
@@ -4,42 +4,122 @@ on:
...
+            arch="-amd64"
+            if [[ $platform == *"-a"* ]]; then
+              arch="" # already in platform string
+            fi
+            old_url=`jq -r ".odo.platform[\"${platform}\"].url" src/tools.json`
+            new_url="https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-nightly-builds/odo-${pltfrm}${arch}-${{ needs.check-odo.outputs.last-commit-id }}${ext}"
             checksum=`curl -s ${new_url}.sha256`
-            jq --indent 4 ".odo.platform.${platform}.url = \"${new_url}\"" src/tools.json | jq --indent 4 ".odo.platform.${platform}.sha256sum = \"${checksum}\"" > src/tools.json.new
+            dlFileName="odo-${pltfrm}${arch}${ext}"
+            cmdFileName="odo${exeExt}"
```
rgrunber, line 110: https://github.com/redhat-developer/vscode-openshift-tools/blob/eb674658b2c69e8977b454a30393a9fb9efe9cbb/src/downloadUtil/downloadBinaries.ts#L40

https://github.com/redhat-developer/vscode-openshift-tools/blob/eb674658b2c69e8977b454a30393a9fb9efe9cbb/src/downloadUtil/archive.ts#L14-L24

https://github.com/redhat-developer/vscode-openshift-tools/blob/eb674658b2c69e8977b454a30393a9fb9efe9cbb/src/downloadUtil/archive.ts#L75-L81

If you look at how `cmdFileName` is used, you'll see it is the name of the file within the archive to be extracted. However, if you check the URLs for each platform you'll see that the command file name embeds the abbreviated commit sha into the file name, so this needs to be adjusted for all platforms.

rgrunber (commented):
This seems to be working as far as generating the PR is conerned. Just found an issue regarding `cmdFileName`.

.github/workflows/check-odo.yml
vrubezhny: `cmdFilename` will always be either `odo` (for Linux and Mac) or `odo.exe` for windows. 
So, I believe it's adjusted

rgrunber (dismissed):
Seems to be working well for me so I would feel free to merge. Only thing I noticed is that because of `github.event_name != 'schedule'`, this will go as far as `update-odo` even though `pr_exists` could have it terminate sooner. Is there a use-case for this ?

Otherwise, it seems fine to merge.

vrubezhny:
> Seems to be working well for me so I would feel free to merge. Only thing I noticed is that because of `github.event_name != 'schedule'`, this will go as far as `update-odo` even though `pr_exists` could have it terminate sooner. Is there a use-case for this ?
> 

The goal is to make it run some steps even if we have no any new commits available in odo repository. But this looks like a check we don't really need as we don't rely on any cache hit/not hit-like event.  
I'm going to remove it before I merge.

PR #4203, opened 2024-06-13
Fixes the enablement for 'Login to DevSandbox' using token from Clipboard

vrubezhny:
This fixes the validation/enablement for 'Login to DevSandbox' button in case of very first Clipboard Poll:
when a token is copied to the Clipboard before the polling starts such a token wasn't taken into account so, even if the token is valid the 'Login to DevSandbox' stayed 'disabled'.

### DIFF
--- a/src/webview/cluster/clusterViewLoader.ts
+++ b/src/webview/cluster/clusterViewLoader.ts
@@ -219,11 +219,13 @@ async function clusterEditorMessageListener (event: any ): Promise<any> {
 
 async function pollClipboard(signupStatus) {
     const oauthInfo = await sandboxAPI.getOauthServerInfo(signupStatus.apiEndpoint);
+    let coldStart = true; // To enforce clipboard token propagation on start
     while (panel) {
         const previousContent = (await vscode.env.clipboard.readText()).trim();
         await new Promise(r => setTimeout(r, 500));
         const currentContent = (await vscode.env.clipboard.readText()).trim();
-        if (previousContent && previousContent !== currentContent) {
+        if (coldStart || (previousContent && previousContent !== currentContent)) {
+            coldStart = false;
             let errCode = '';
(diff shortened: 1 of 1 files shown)


datho7561:
I tried to reproduce the problem this Pr is meant to fix and I can't. Just to double check, is this for the webview or the quickpick?

vrubezhny:
> I tried to reproduce the problem this Pr is meant to fix and I can't. Just to double check, is this for the webview or the quickpick?

That's the WebView. 

[on cold start] If you're not logged it to RedHat  DevSandbox - I just got through the Login steps, then, when sandbox is provisioned I press `Get Token` and then, after the token is copied into the Clipboard`, I see the `Login to DevSandbox` button disabled.

[on cold start] If you have sandbox provisioned:  

1) Go to the `display token` page and copy the token to the Clipboard.
2) On `Welcome` page press `DEVELOPER SANDBOX` button, then see `Login to DevSandbox` button disabled.
...3) Like a workaround, in order to make it to be enabled you need to copy some non-token text to clipboard, then again copy a valid token to the Clipboard.

datho7561:
I found that the "Login" button was enabled if I had the token copied into my clipboard before opening the webview. However, I didn't bother trying to press the "Get Token" button, since the login button was enabled. I'll see if that makes a difference.

vrubezhny:
> I found that the "Login" button was enabled if I had the token copied into my clipboard before opening the webview. However, I didn't bother trying to press the "Get Token" button, since the login button was enabled. I'll see if that makes a difference.

When the  `pollClipboard(signupStatus)` is invoked late (as in my case) and clipboard contains a valid token value, then inside the `pollClipboard` the `previousContent` value is exactly same as `currentContent` , so, once the values are equal,  it doesn't validate the token and doesn't send any message to the page to update the buttons.

Issue #4172, opened 2024-06-03
There is no any error/warning notification when Helm Chart cannot be installed 


vrubezhny:
Some Helm Charts available for installation cannot be installed on a cluster. The installation process finishes without any indication of a failure but no any Helm Release appears on the App. Explorer. tree:

[Screencast from 2024-06-03 22-14-50.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/1b6b0b6e-f880-47f6-8b03-36bb9e482d63)

vrubezhny:
Actually for all the Helm Charts that cannot be installed (silently closing the installation view and not adding any objects to `<project>/Helm/Releases/` App. Explorer tree node) an error actually happens that is not reported anywhere.

Such errors can be easily reported on the Helm Chart installation view: 

[Screencast from 2024-06-14 03-21-15.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/a1517873-ae56-45a3-8d9d-24aacfac8403)

PR #4206, opened 2024-06-14
There is no any error/warning notification when Helm Chart cannot be installed #4172

vrubezhny:
Fixes: #4172

### DIFF
--- a/src/helm/helm.ts
+++ b/src/helm/helm.ts
@@ -78,7 +78,7 @@ export async function installHelmChart(
 ): Promise<CliExitData> {
     await syncHelmRepo(repoName);
     return await CliChannel.getInstance().executeTool(
-        HelmCommands.installHelmChart(name, repoName, chartName, version), undefined, false
+        HelmCommands.installHelmChart(name, repoName, chartName, version)
     );
 }
 


vrubezhny:
The PR enables the error reporting on Helm Chart installation failure:

[Screencast from 2024-06-14 03-21-15.webm](https://github.com/redhat-developer/vscode-openshift-tools/assets/620781/419effe6-a509-4a5b-a599-5de20f55c554)

msivasubramaniaan (dismissed):
LGTM

mohitsuman:
@vrubezhny, after the Helm Chart is installed, a "VS Code notification bar indicating the installation is successful" should appear. This notification aids in determining whether the installation process succeeded or failed.

vrubezhny:
> @vrubezhny, after the Helm Chart is installed, a "VS Code notification bar indicating the installation is successful" should appear. This notification aids in determining whether the installation process succeeded or failed.

@mohitsuman As you can see on the screencasts the issue  https://github.com/redhat-developer/vscode-openshift-tools/issues/4172#issue-2331983154 and here after the fix  https://github.com/redhat-developer/vscode-openshift-tools/pull/4206#issuecomment-2167066192  there is "Installing ....` information message when the installation starts, but there is no any "VSCode notification bars" appearing after the successful installation. 

Do you want me to add such a notification message? (I suppose it should be an informational message then).

In case of a failure, I believe, the error message appearing on the Helm Chart installation pane is enough to notify the user, isn't it?

src/helm/helm.ts
```diff
@@ -78,7 +78,7 @@ export async function installHelmChart(
 ): Promise<CliExitData> {
     await syncHelmRepo(repoName);
     return await CliChannel.getInstance().executeTool(
-        HelmCommands.installHelmChart(name, repoName, chartName, version), undefined, false
+        HelmCommands.installHelmChart(name, repoName, chartName, version), undefined, true
```
datho7561, line 81: The default value for 2nd and 3rd parameters for `executeTool()` are `undefined` and `true`, so you can remove these two parameters and it will still work.

datho7561 (approved):
Works well, but the code can be cleaned up a little. Feel free to fix it or merge as-is.

I think it's okay for the time being to not include a notification that the chart was installed successfully, since currently the user is notified to this fact by the popup window within the webview closing.

src/helm/helm.ts
vrubezhny: Yes, You're right. Fixed.

vrubezhny:
> I think it's okay for the time being to not include a notification that the chart was installed successfully, since currently the user is notified to this fact by the popup window within the webview closing.

So OK. I'm merging it in. Feel free to open an issue if we need any other notifications to be added

PR #4225, opened 2024-06-26
Fix check-odo GH actions property array item addressing issue

vrubezhny:


vrubezhny:
Created for any case, we can fix [it](https://github.com/redhat-developer/vscode-openshift-tools/actions/runs/9675833361) this way, but we must refuse to do it in favor of https://github.com/redhat-developer/vscode-openshift-tools/pull/4191

PR #4247, opened 2024-07-01
Remove 'make-fetch-happen' dependency

vrubezhny:
As Fetch API is compatible with Node.js >= v18, the engine and dependencies are accordingly updated:

- engines:
```
"npm": ">=8.6.0",
"node": ">=18.0.0"
```
- devDependencies:
```
"@types/node": "^18.19.39",
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,9 @@
 	},
 	"bugs": "https://github.com/redhat-developer/vscode-openshift-tools/issues",
 	"engines": {
-		"vscode": "^1.64.2",
-		"npm": ">=8.0.0",
-		"node": ">=16.0.0"
+		"vscode": "^1.82.3",
+		"npm": ">=8.6.0",
+		"node": ">=18.0.0"
 	},
 	"badges": [
 		{
@@ -83,7 +83,6 @@
 		"json-schema": "^0.4.0",
 		"json-to-ast": "^2.1.0",
 		"lodash": "^4.17.21",
-		"make-fetch-happen": "^12.0.0",
 		"mkdirp": "^3.0.1",
 		"node-yaml-parser": "^0.0.9",
 		"portfinder": "^1.0.32",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -14,8 +14,8 @@
 	"bugs": "https://github.com/redhat-developer/vscode-openshift-tools/issues",
 	"engines": {
 		"vscode": "^1.64.2",
```
datho7561, line 16: The first version of VS Code that includes NodeJS 18 appears to be 1.82: https://code.visualstudio.com/updates/v1_82 . This version was released under a year ago. I don't know how old of a version of VS Code we are supposed to support.

If we are making this upgrade, we should probably list the engine version as `^1.82.3` .

datho7561 (changes requested):
Seems to work for me, but we should update the engine version if we are making this change so that versions of VS Code that don't come with Node 18 can't download this update.

package.json
vrubezhny: I think users are still able to use vscode 1.64 together with Node.js > 16.x, aren't they?

On the other hand, vscode 1.64 is out of support, so, I'm +1 for this change. 

Fixed

datho7561: My understanding is that VS Code uses whatever version of nodejs that the version of electron they depend on uses. Since fetch wasn't build into Node until Node 18, the first version of vscode that supports is whatever version added the version of Electron that uses Node 18. From what I can tell, that is `1.82.0`.

Node 16 is out of support, so I think it's safe to upgrade the engine to `1.82.0`, but we *should* change the engine version, since `fetch` isn't built in to Node 16.

test/unit/openshift/cluster.test.ts
```diff
@@ -211,7 +211,7 @@ suite('Openshift/Cluster', function() {
             });
 
             test('checks cluster url name is valid url', async () => {
-                let result: string | Thenable<string>;
+                let result: any;
```
datho7561, line 214: Not a huge deal, but feel free to leave out the type if the best description is `any`.

vrubezhny: Fixed

Issue #4255, opened 2024-07-03
Remove outdated 'migrateFromOdo018' method


vrubezhny:
The method `migrateFromOdo018` looks outdated and doesn't perform properly the migration of ODO preferences:
- `odo-config.yaml` ODO Global Preference file is not used anymore: a long time ago it was renamed to `config.yaml` and then later it was renamed again to `preference.yaml` (See: https://github.com/redhat-developer/odo/pull/1784)
- I'm pretty sure that the allowed preference key names also have changed since then, so we cannot just copy the old file to `~/.odo/preference.yaml` in order to migrate properly.

So,the following method looks like non-functional anymore and is to be removed from the extension activation:

```
function migrateFromOdo018(): void {
    const newCfgDir = path.join(Platform.getUserHomePath(), '.odo');
    const newCfg = path.join(newCfgDir, 'odo-config.yaml');
    const oldCfg = path.join(Platform.getUserHomePath(), '.kube', 'odo');
    if (!fsx.existsSync(newCfg) && fsx.existsSync(oldCfg)) {
        fsx.ensureDirSync(newCfgDir);
        fsx.copyFileSync(oldCfg, newCfg);
    }
}
```

vrubezhny:
Related to #4226

Issue #4226, opened 2024-06-26
Extension is slow to activate


fbricon:
OpenShift toolkit is the slowest of all the extensions I have installed. Taking between 300ms and almost 900ms to activate, on my MBP Pro M3.
<img width="599" alt="Screenshot 2024-06-26 at 17 28 35" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/148698/874fd02f-8d86-4936-a935-924c2a11b966">

mohitsuman:
We need to look at the events getting activated together when making calls `odo` to load the view.  Also relates to https://github.com/redhat-developer/vscode-openshift-tools/issues/3850

mohitsuman:
@datho7561 @vrubezhny @msivasubramaniaan maybe we need to load each view separately instead of loading all at once.

vrubezhny:
@fbricon Could you please take a try it again (using the main branch or by installing a VSIX built from the main branch)?

Feel free to reopen the issue in case it's still too slow on activate. Thanks!

fbricon:
Running from main, the bundled extension is still quite slow to activate (from 300ms to 1000ms)
<img width="976" alt="Screenshot 2024-07-08 at 15 13 05" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/148698/6bcfe51b-0a40-4a18-8185-8603612d4111">

One potential hint is this message when running `npx @vscode/vsce package`:

> This extension consists of 10353 files, out of which 4900 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore
 DONE  Packaged: /Users/fbricon/Dev/projects/vscode-openshift-tools/vscode-openshift-connector-1.14.0.vsix (10353 files, 170.21MB)
 
 Indeed, the zip contains a shitload of .js files, that we should bundle with webpack or eslint:
 
<img width="259" alt="Screenshot 2024-07-08 at 15 17 32" src="https://github.com/redhat-developer/vscode-openshift-tools/assets/148698/ad8bef40-3d56-4c5d-8f5e-e7ef4406839c">

PR #4256, opened 2024-07-03
Extension is slow to activate #4226

vrubezhny:
The PR adds some optimization to the extension activation.

Also, the PR removes function 'migrateFromOdo018' which looks over outdated and doesn't make its work properly after a numerous changes in ODO. See: #4255

Fixes: #4255
Fixes: #4226

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See LICENSE file in the project root for license information.
  *-----------------------------------------------------------------------------------------------*/
 
-import * as path from 'path';
 import {
     authentication,
     commands, env, ExtensionContext, languages, QuickPickItemKind,
@@ -11,7 +10,7 @@ import {
     StatusBarItem, window,
     workspace
 } from 'vscode';
-import * as k8s from 'vscode-kubernetes-tools-api';
+import { extension as k8sExtension } from 'vscode-kubernetes-tools-api';
 import { REDHAT_CLOUD_PROVIDER } from './cloudProvider/redhatCloudProvider';
 import { ComponentsTreeDataProvider } from './componentsView';
 import { DebugSessionsView } from './debug';
@@ -27,14 +26,12 @@ import { ToolsConfig } from './tools';
 import { TokenStore } from './util/credentialManager';
 import { getNamespaceKind, KubeConfigUtils, setKubeConfig } from './util/kubeUtils';
 import { Context as KcuContext } from '@kubernetes/client-node/dist/config_types';
-import { Platform } from './util/platform';
 import { setupWorkspaceDevfileContext } from './util/workspace';
 import { registerCommands } from './vscommand';
 import { OpenShiftTerminalManager } from './webview/openshift-terminal/openShiftTerminal';
 import { WelcomePage } from './welcomePage';
 import { registerYamlHandlers } from './yaml/yamlDocumentFeatures';
 
-import fsx = require('fs-extra');
 import { Oc } from './oc/ocWrapper';
 import { K8S_RESOURCE_SCHEME, K8S_RESOURCE_SCHEME_READONLY, KubernetesResourceVirtualFileSystemProvider } from './k8s/vfs/kuberesources.virtualfs';
 import { KubernetesResourceLinkProvider } from './k8s/vfs/kuberesources.linkprovider';
@@ -43,16 +40,6 @@ import { KubernetesResourceLinkProvider } from './k8s/vfs/kuberesources.linkprov
 // this method is called when your extension is deactivated
 export function deactivate(): void {
     // intentionally left blank
-}
-
-function migrateFromOdo018(): void {
-    const newCfgDir = path.join(Platform.getUserHomePath(), '.odo');
-    const newCfg = path.join(newCfgDir, 'odo-config.yaml');
-    const oldCfg = path.join(Platform.getUserHomePath(), '.kube', 'odo');
-    if (!fsx.existsSync(newCfg) && fsx.existsSync(oldCfg)) {
-        fsx.ensureDirSync(newCfgDir);
-        fsx.copyFileSync(oldCfg, newCfg);
-    }
 }
 
 async function verifyBundledBinaries(): Promise<{ odoPath: string, ocPath: string, helmPath: string }> {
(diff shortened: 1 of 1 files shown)


vrubezhny:
My calculations for the extension loading time before (installed from MS Marketplace) and after the fix (installed from locally built VSIX) differ in numbers, here are some examples:

- Before the fix (installed from MS Marketplace):
Screenshot from 2024-07-03 22-45-44

- After the fix (installed from locally built VSIX):
Screenshot from 2024-07-03 22-40-32

src/extension.ts
```diff
@@ -258,34 +244,38 @@ export async function activate(extensionContext: ExtensionContext): Promise<unkn
         }
     }
 
-    OpenShiftExplorer.getInstance().onDidChangeContextEmitter.event(async (context) => {
-        const kind = await getNamespaceKind();
-        void Oc.Instance.getActiveProject().then((namespace) =>
-            updateContextStatusBarItem(activeNamespaceStatusBarItem, 'project-node', namespace, `Current ${kind}: ${namespace}`,
-                !isNamespaceInfoStatusBarDisabled()));
+    const updateContextAndProjectStatusBarItems = async function(context?: string): Promise<void> {
+        return getNamespaceKind().then(async (kind) => {
```
datho7561, line 248: this function is async. You can await `getNamespaceKind` instead of using `.then(...`

src/extension.ts
```diff
@@ -72,9 +59,9 @@ async function registerKubernetesCloudProvider(): Promise<void> {
 
 export async function activate(extensionContext: ExtensionContext): Promise<unknown> {
     void WelcomePage.createOrShow();
-    await commands.executeCommand('setContext', 'isVSCode', env.uiKind);
+    void commands.executeCommand('setContext', 'isVSCode', env.uiKind);
     // UIKind.Desktop ==1 & UIKind.Web ==2. These conditions are checked for browser based & electron based IDE.
```
datho7561, line 63: Please remove this comment, it doesn't seem relevant anymore

datho7561 (changes requested):
Seems pretty good so far. Noticeably speeds up the extension start up. I think the main speed up is not blocking on registering the commands/webviews on startup. I was worried this would introduce concurrency bugs, but it doesn't seem to introduce any.

I've added two small comments.

vrubezhny:
There's still something strange in the extension activation time calculation... 


The extension's `activate` method

src/extension.ts
vrubezhny: ACK

src/extension.ts
vrubezhny: Removed

vrubezhny:
> Seems pretty good so far. Noticeably speeds up the extension start up. I think the main speed up is not blocking on registering the commands/webviews on startup. I was worried this would introduce concurrency bugs, but it doesn't seem to introduce any.

Still trying to introduce anything that breaks it :sunglasses: ... I've turned setting up the Kube Context call to be async (it takes quite a lot of time to finish and also it might require a user intervention, so...) and wait at the end for it's promise to be resolved... Still doesn't look like I broke anything (even the status bar active context and project items set up).

@datho7561 Anyway, take a look please

> I've added two small comments.

Done

datho7561:
Try out this patch to do the timing:

```diff-typescript
diff --git a/src/extension.ts b/src/extension.ts
index 62cdd465..4a39e36d 100644
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See LICENSE file in the project root for license information.
  *-----------------------------------------------------------------------------------------------*/
 
+const startTime = new Date().getTime();
+import { Context as KcuContext } from '@kubernetes/client-node/dist/config_types';
 import {
     authentication,
     commands, env, ExtensionContext, languages, QuickPickItemKind,
@@ -25,16 +27,15 @@ import { startTelemetry } from './telemetry';
 import { ToolsConfig } from './tools';
 import { TokenStore } from './util/credentialManager';
 import { getNamespaceKind, KubeConfigUtils, setKubeConfig } from './util/kubeUtils';
-import { Context as KcuContext } from '@kubernetes/client-node/dist/config_types';
 import { setupWorkspaceDevfileContext } from './util/workspace';
 import { registerCommands } from './vscommand';
 import { OpenShiftTerminalManager } from './webview/openshift-terminal/openShiftTerminal';
 import { WelcomePage } from './welcomePage';
 import { registerYamlHandlers } from './yaml/yamlDocumentFeatures';
 
-import { Oc } from './oc/ocWrapper';
-import { K8S_RESOURCE_SCHEME, K8S_RESOURCE_SCHEME_READONLY, KubernetesResourceVirtualFileSystemProvider } from './k8s/vfs/kuberesources.virtualfs';
 import { KubernetesResourceLinkProvider } from './k8s/vfs/kuberesources.linkprovider';
+import { K8S_RESOURCE_SCHEME, K8S_RESOURCE_SCHEME_READONLY, KubernetesResourceVirtualFileSystemProvider } from './k8s/vfs/kuberesources.virtualfs';
+import { Oc } from './oc/ocWrapper';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function
 // this method is called when your extension is deactivated
@@ -276,6 +277,9 @@ export async function activate(extensionContext: ExtensionContext): Promise<unkn
     // Wait for finishing Kube Config setup
     await setKubeConfigPromise;
 
+    // eslint-disable-next-line no-console
+    console.log(`vscode-openshift activation time: ${new Date().getTime() - startTime}`);
+
     return {
         verifyBundledBinaries
     };

```

vrubezhny:
> Try out this patch to do the timing:

Yep... 
image

Now the reported activation times are closer to each other... even the vscode's time is less than the Extension activation time (?)...

But I cannot do anything to speed up the time spent to import the required dependencies...

datho7561:
> But I cannot do anything to speed up the time spent to import the required dependencies... Can I?

Not really. *Maybe* we could all the code into one file, but we would need to make sure that we don't use any libraries that prevent that, and we can investigate/do that in a future PR.

I'll take another look at the changes and make sure it looks good.

datho7561 (approved):
This does the trick for me. I no longer have to wait a few seconds to interact with the Application Explorer. It loads in what feels like around a second.


I don't know if you want to wait on feedback from Fred, or go ahead and merge. Either is fine to me.

vrubezhny:
> Maybe we could all the code into one file, but we would need to make sure that we don't use any libraries that prevent that, and we can investigate/do that in a future PR.

Yeah. I got the idea.

@datho7561 Let's merge this one. And then we can ask Fred for his feedback and improve it later.

PR #4283, opened 2024-07-15
validated bad container image URL

msivasubramaniaan:


src/deployment.ts
```diff
@@ -152,12 +152,23 @@ export class Deployment {
...
                 if (validator.isURL(inputBox.value)) {
                     inputBox.validationMessage = undefined;
                 } else {
                     inputBox.validationMessage = 'Please enter a valid URL';
                 }
+
+                //check the mentioned url has image referece
+                inputBox.validationMessage = 'Checking the image info'
+                inputBox.enabled = false;
+                if (! await Oc.Instance.hasImageInfo(inputBox.value)) {
```
vrubezhny, line 165: This makes the validation SO SLOW, so that it misses the characters I'm typing in.


https://github.com/user-attachments/assets/9a15e609-f8c0-4218-8c7c-63723a224f70

The text I've typed in was: "docker.io"

vrubezhny (changes requested):
We should change the way we validate it. See the in line comment

vrubezhny (dismissed):
Looks good to me.
After moving the image info validation  to `onDidAccept` it works really well!
Also it doesn't make any excessive 'oc' binary invocations, which is also good, imho.

src/oc/ocWrapper.ts
```diff
@@ -675,6 +675,13 @@ export class Oc {
             .catch((error) => namespaces);
     }
 
+    public async hasImageInfo(url: string): Promise<boolean> {
+        const result = await CliChannel.getInstance().executeTool(
```
datho7561, line 679: Try this with `docker.io/library/mongo`. For me at least, it fails, so I can't deploy mongodb

datho7561 (changes requested):

datho7561:
It would be nice to have some sort of indication that a verification is going on; I don't know yet how this would be implemented though.

src/oc/ocWrapper.ts
vrubezhny: image

I could successfully deploy `mongodb` using URL `docker.io/library/mongo`. Some temporary/network problem?

msivasubramaniaan:
> It would be nice to have some sort of indication that a verification is going on; I don't know yet how this would be implemented though.

@datho7561 added progress to show the info
image-url-validation

src/oc/ocWrapper.ts
datho7561: What version of the mongo DB image do you have? I have `a31b196b207d` which should be the latest image, since I recently deleted all my container images to free some space. Maybe this affects it?

datho7561: I'll double check on windows

vrubezhny: Can I check it with OS Tools somehow?

Docker says the `mongo:latest` is `a31b196b207d`...

datho7561: The command result is an error for me, since it detects windows and linux versions of the docker image

datho7561:
The progress indicator on the image check looks great! I'd like to figure out what's going on with the mongodb image before merging.

datho7561:
I just hit my pull rate limit for docker. Maybe this is related?

src/oc/ocWrapper.ts
vrubezhny: > The command result is an error for me, since it detects windows and linux versions of the docker image

I didn't get this...  So, you have the same error result while deploying the same image on Windows and *nix? 
And what does it give to us? 

... And yes, I think if you hit the docker rate limit, you'll probably won't  be able to deploy any image till the counter is dropped. Not 100% sure though.

vrubezhny:
> I just hit my pull rate limit for docker. Maybe this is related?

If I execute `oc image info docker.io/library/mongo` I get quite a long list of `mongo` images with different versions, and the command output ends with:

```
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS             DIGEST
  linux/amd64    sha256:19c11a8f1064fd2bb713ef1270f79a742a184cd57d9bb922efdd2a8eca514af8
  linux/arm64/v8 sha256:f9b01e368fd95a783a68471930a15c2494d525318ff65a1b408ff3e6bc83e3d4
  windows/amd64  sha256:bbb44a208559a1fad8f164e0bd97db2b9f4b699d6adfa01ba4f20b9ec60eb832
  windows/amd64  sha256:bc25ab206f79c4618ca23b8fb5f8b2710783ed6a5058f30e8f196a85f3f5c3fb
```

None of these images has neither `a31b196b207d` (`mongo:latest`) nor `1cd395100002` sha256 and actually might not be full. 

So, I'm not actually sure if we can really check what's getting installed with `oc create deployment ... --image=docker.io/library/mongo`.

datho7561:
@vrubezhny @msivasubramaniaan I think this will work, however I have exceeded the rate limit and need to wait before I can try it again:

`oc image info docker.io/library/mongo:latest  --show-multiarch=true`

changes I made:

`:latest` --> This means info is only collected for the newest version of the container image

`--show-multiarch=true` --> from the argument description, this should prevent an error code from being returned when the container supports multiple architectures

vrubezhny:
> `oc image info docker.io/library/mongo:latest --show-multiarch=true`

Do you suggest adding `:latest` and `--show-multiarch=true` to the command arguments?

src/oc/ocWrapper.ts
datho7561: Seems to be working now for me. Does it work for you, Victor?

datho7561 (approved):
Looks good to me! Thanks, Muthu!

vrubezhny (approved):
Looks good to me. Thanks!

Issue #4289, opened 2024-07-16
Slow loading of the OpenShift explorer caused by duplicate `oc` commands


fbricon:
I have an openshift project on devsandbox with a of couple pods. 
<img width="517" alt="Screenshot 2024-07-16 at 16 02 14" src="https://github.com/user-attachments/assets/064391ae-5fc7-4ed1-9880-7f3df7ab8cd7">

Loading the explorer is fairly slow. Looking at the OpenShift output view, I captured the commands executed during a refresh request:

```
1:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/func" list -o json
4:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" api-resources | grep openshift
162:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" auth can-i create projectrequests
166:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" auth can-i list projects
170:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" projects -q
174:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" api-resources | grep openshift
332:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" projects -q
336:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" auth can-i get csv
339:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" get csv -o=json
349:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" get service -o=json
526:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" projects -q
530:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" auth can-i delete projects --namespace=marketplace-stats-dev
534:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" api-resources | grep openshift
692:"/Users/fbricon/Dev/projects/vscode-openshift-tools/out/tools/darwin/oc" api-resources | grep openshift
[...]

PR #4300, opened 2024-07-18
Slow loading of the OpenShift explorer caused by duplicate oc commands #4289

vrubezhny:
Fixes: #4289

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -4,12 +4,15 @@
  *-----------------------------------------------------------------------------------------------*/
 
 import { VSCodeSettings } from '@redhat-developer/vscode-redhat-telemetry/lib/common/vscode/settings';
-import * as vscode from 'vscode';
 import * as cp from 'child_process';
+import * as vscode from 'vscode';
 import { CommandText } from './base/command';
 import { ToolsConfig } from './tools';
 import { ChildProcessUtil, CliExitData } from './util/childProcessUtil';
 import { VsCommandError } from './vscommand';
+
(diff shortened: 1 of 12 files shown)


src/cli.ts
```diff
@@ -52,13 +55,24 @@ export class CliChannel {
...
 
-    async executeTool(command: CommandText, opts?: cp.ExecOptions, fail = true): Promise<CliExitData> {
+    async executeTool(command: CommandText, opts?: cp.ExecOptions, fail = true, executionContext?: ExecutionContext): Promise<CliExitData> {
         const commandActual = command.toString();
         const commandPrivacy = command.privacyMode(true).toString();
         const [cmd] = commandActual.split(' ');
         const toolLocation = await ToolsConfig.detect(cmd);
         const optsCopy = CliChannel.applyEnv(opts, CliChannel.createTelemetryEnv())
+
+        const key = `${command.command} ${[command.parameter, ...command.options.map((o)=>o.toString())]}`;
```
datho7561, line 65: I believe that `commandActual` is the same string as this `key`, and that has already been computed

datho7561 (changes requested):
Small fix, but otherwise is working amazingly!

datho7561 (dismissed):
Seems good. Thanks, Victor!

src/cli.ts
vrubezhny: It's not the same, but actually using `commandActual` value is better.
Fixed

datho7561 (approved):

Issue #4320, opened 2024-07-25
Migrate the odo commands


mohitsuman:
The extension should figure out a workflow to have NO usage of the odo commands.

- [ ] deploy
- [ ] delete component
- [ ] dev
- [ ] describe component
- [ ] logs
- [ ] init
- [ ] run

PR #4235, opened 2024-06-27
4151 create a deployment using a build config

msivasubramaniaan:

PR #4330, opened 2024-07-30
Replace odo analyze command

msivasubramaniaan:
Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/4320

[@davthomp](https://redhat-internal.slack.com/team/U013QPX5HL2)  
[@vrubezhny](https://redhat-internal.slack.com/team/UJSV9MH7T)

Below are the things missed and I don't know how to handle below two items. Let me know if you have any idea.
https://github.com/devfile/alizer doesn't provide any checksum on the release, So the CLI on our extension downloaded each time
alizer devfile command not having "name" key in the response. But odo analyze does
Ex: {
  devfile: "java-maven",
  devfileRegistry: "Test1",
  devfileVersion: "1.3.1",
  name: "org-eclipse-lemminx",
}


datho7561:
I still don't think I get the point of this PR. If we are removing `odo` completely, then we won't be able to work with Devfiles, so we won't need to analyze which one to use.

I understand using alizer instead of odo for https://github.com/redhat-developer/vscode-openshift-tools/pull/4235 , but that's not what this is.

msivasubramaniaan:
> I still don't think I get the point of this PR. If we are removing `odo` completely, then we won't be able to work with Devfiles, so we won't need to analyze which one to use.
> 
> I understand using alizer instead of odo for #4235 , but that's not what this is.

Hello @datho7561 
As this PR for just replace `odo analyze` command with `alizer devfile`. It is common for both component creation using git import and #4235

src/odo/odoWrapper.ts
```diff
@@ -194,13 +194,12 @@ export class Odo {
...
+    public async alizerAnalyze(currentFolderPath: Uri): Promise<AlizerAnalyzeResponse[]> {
         const cliData: CliExitData = await this.execute(
-            new CommandText('odo', 'analyze -o json'),
-            currentFolderPath,
+            new CommandText('alizer', `devfile ${currentFolderPath.fsPath}`)
         );
-        const parse = JSON.parse(cliData.stdout) as AnalyzeResponse[];
-        return parse;
+        const parse = JSON.parse(cliData.stdout) as AlizerAnalyzeResponse[];
+        return [[...parse].shift()];
```
datho7561, line 202: Shift returns the first element. Why not just return the first element (or undefined if there is no element) and change the return type to `Promise<AlizerAnalyzeResponse>` ?

src/odo/odoWrapper.ts
```diff
@@ -194,13 +194,12 @@ export class Odo {
         );
     }
 
-    public async analyze(currentFolderPath: string): Promise<AnalyzeResponse[]> {
+    public async alizerAnalyze(currentFolderPath: Uri): Promise<AlizerAnalyzeResponse[]> {
```
datho7561, line 197: It might be a good idea to move this function to it's own file, since it no longer uses odo, and this file contains wrappers around odo commands

datho7561 (changes requested):
From my manual testing, this seems good. I noticed two small things in the code.

test/integration/odoWrapper.test.ts
```diff
@@ -100,12 +101,12 @@ suite('./odo/odoWrapper.ts', function () {
         });
 
         test('analyze()', async function () {
```
vrubezhny, line 103: @msivasubramaniaan The `test('analyze()', async function () {` test case should probably also be moved out from `test/integration/odoWrapper.test.ts` into a separate test, shouldn't it?

vrubezhny (changes requested):
`odoWrapper` test also needs some attention.

vrubezhny:
@msivasubramaniaan The related test are still failing:

```
1) Create component from git URL
2) "after each" hook: context for "Create component from git URL"
3) "after all" hook: context for "Create component from template project"
 ```
 
 The first one looks like hangs or take a vary long time to complete, causing the following tests to fail too. But it could be also that some `alizer` related code failure causes it not to finish successfully.
 
 Could you please take a look?

msivasubramaniaan:
@datho7561 and @vrubezhny Moved `odo analyze` command and corresponding test cases as well. Please review and merge. Once this merged I will make the https://github.com/redhat-developer/vscode-openshift-tools/pull/4235 PR ready

vrubezhny (approved):
Looks good to me. Thanks!

datho7561 (changes requested):

datho7561 (approved):
Looks good to me. Thanks, Muthu!

PR #4357, opened 2024-08-12
Allow commands to work with .yml extension

branic:
This change allows the "Apply YAML to cluster" and "Delete YAML from clsuter" commands to accpet a file ending in `.yml`. Using a `.yml` extension name for YAML files is common and widely used so it makes sense to support both the `.yaml` and `.yml` naming.

### DIFF
--- a/src/yamlFileCommands.ts
+++ b/src/yamlFileCommands.ts
@@ -19,7 +19,8 @@ export class YamlFileCommands {
         let message: string;
 
         if (!document
-            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.json'))) {
+            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.yml')
+            || document.fileName.endsWith('.json'))) {
             message =
                 '\'OpenShift: Create\' command requires a .yaml or a .json file opened in editor.';
         }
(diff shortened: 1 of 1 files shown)


src/yamlFileCommands.ts
```diff
@@ -19,7 +19,8 @@ export class YamlFileCommands {
         let message: string;
 
         if (!document
-            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.json'))) {
+            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.yml')
+            || document.fileName.endsWith('.json'))) {
             message =
                 '\'OpenShift: Create\' command requires a .yaml or a .json file opened in editor.';
```
vrubezhny, line 25: The message text probably worth to be changed too in order to reflect the change. WDYT?

src/yamlFileCommands.ts
```diff
@@ -62,7 +63,8 @@ export class YamlFileCommands {
         let message: string;
 
         if (!document
-            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.json'))) {
+            || !(document.fileName.endsWith('.yaml') || document.fileName.endsWith('.yml')
+            || document.fileName.endsWith('.json'))) {
             message =
                 '\'OpenShift: Delete\' command requires a .yaml or a .json file opened in editor.';
```
vrubezhny, line 69: Same question here

src/yamlFileCommands.ts
branic: I agree. I've updated the message text.

src/yamlFileCommands.ts
branic: I agree. I've updated the message text.

vrubezhny (approved):
Looks good to me. Thanks!

PR #4395, opened 2024-08-28
Delete component in UI tests using context menu instead of filesystem

lgrossma:
See: https://github.com/redhat-developer/vscode-openshift-tools/issues/3693


vrubezhny (approved):
Looks good to me. Thanks!

PR #4400, opened 2024-08-29
[build] Fix vulnerability in @octokit/rest 16.0.0 - 16.43.2

vrubezhny:
```

lodash.set  *
Severity: high
Prototype Pollution in lodash - https://github.com/advisories/GHSA-p6mc-m468-83gw
fix available via `npm audit fix --force`
Will install @octokit/rest@21.0.2, which is a breaking change
node_modules/lodash.set
  @octokit/rest  16.0.0 - 16.43.2
  Depends on vulnerable versions of lodash.set
  node_modules/@octokit/rest
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
 				"@mui/lab": "^5.0.0-alpha.173",
 				"@mui/material": "^5.16.7",
 				"@mui/styles": "^5.16.7",
-				"@octokit/rest": "^16.30.1",
+				"@octokit/rest": "^21.0.2",
 				"@rjsf/core": "^5.20.1",
 				"@rjsf/mui": "^5.20.1",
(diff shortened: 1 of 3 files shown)


vrubezhny:
@msivasubramaniaan Do we really  need this code? It seems to be not used anywhere... Or do you plan to use it somehow later?

datho7561:
I was able to completely remove these files without causing any build failures:
- `src/util/getGitService.ts`
- `src/util/githubService.ts`

msivasubramaniaan (changes requested):
`githubServices.ts` and `getGitService.ts` files are no longer needed. We can remove the files from repo and its dependency

vrubezhny:
Closing as not needed anymore

PR #4406, opened 2024-08-30
added yaml editor for helm values

msivasubramaniaan:


vrubezhny:
At the first look... I cannot install a Helm Chart due to "Install" button inaccessible (it's just not visible under the YAML editor and we have no vertical scroll at the same time):

image

msivasubramaniaan:
> At the first look... I cannot install a Helm Chart due to "Install" button inaccessible (it's just not visible under the YAML editor and we have no vertical scroll at the same time):
> 
> image

Here is the screenshot on my machine
image
Anyway I am look at the issue which u mentioned. Mean time may I know your screen size?

vrubezhny:
> Here is the screenshot on my machine

Just make the window less in height:

image

The YAML editor scroll appears for a long YAML properties file, but the whole chart window still doesn't show the button if it doesn't fit.

msivasubramaniaan:
> > Here is the screenshot on my machine
> 
> Just make the window less in height:
> 
> image
> 
> The YAML editor scroll appears for a long YAML properties file, but the whole chart window still doesn't show the button if it doesn't fit.

Can you please check with the latest code?

vrubezhny:
> > > Here is the screenshot on my machine
> > 
> > Just make the window less in height:
> > The YAML editor scroll appears for a long YAML properties file, but the whole chart window still doesn't show the button if it doesn't fit.
> 
> Can you please check with the latest code?

What should be changed now? I still see  the same 16 rows in the editor... Or it's about font size/spacing?
Or I didn't updated properly?

Anyway, imho, looks good. (However for the small screen/lower resolution users it still could be problematic.)

I would probably make it having like 12 rows of text and think about making the header part (Icon/description + name + version selection) more compact - currently it spreads to at least 1/3 of the overall height. But we can improve it later. 
@datho7561 WDYT?

The following questions are not for this PR, but probably for a future:
- As far as I understand, there are some values that are predefined and can be used as is, but there are some that are required to be edited/entered - can we highlight them somehow, so user surely know which values are mandatory and must be provided? 
- Are we able to validate the entered values before creating a chart? I mean, f. i. `port: admin` doesn't look like a correct value in the Redhat Wildfly Chart shown above:
```
  livenessProbe:
    httpGet:
      path: /health/live
      port: admin
```

msivasubramaniaan:
> > > > Here is the screenshot on my machine
> > > 
> > > 
> > > Just make the window less in height:
> > > The YAML editor scroll appears for a long YAML properties file, but the whole chart window still doesn't show the button if it doesn't fit.
> > 
> > 
> > Can you please check with the latest code?
> 
> What should be changed now? I still see the same 16 rows in the editor... Or it's about font size/spacing? Or I didn't updated properly?
> 
> Anyway, imho, looks good. (However for the small screen/lower resolution users it still could be problematic.)
> 
> I would probably make it having like 12 rows of text and think about making the header part (Icon/description + name + version selection) more compact - currently it spreads to at least 1/3 of the overall height. But we can improve it later. @datho7561 WDYT?
> 
> The following questions are not for this PR, but probably for a future:
> 
> * As far as I understand, there are some values that are predefined and can be used as is, but there are some that are required to be edited/entered - can we highlight them somehow, so user surely know which values are mandatory and must be provided?
> * Are we able to validate the entered values before creating a chart? I mean, f. i. `port: admin` doesn't look like a correct value in the Redhat Wildfly Chart shown above:
> 
> ```
>   livenessProbe:
>     httpGet:
>       path: /health/live
>       port: admin
> ```

@vrubezhny The schema validation will be taken care on feature PR. @mohitsuman wants this feature in the upcoming release.

PR #4408, opened 2024-09-02
Increasing timeout to fix instability in debug test

lgrossma:
This PR aims to stabilize frequently failing test "Component Context Menu - Debug works". I tried running the test multiple times and the instability seems to be solved by this change


vrubezhny (approved):
Looks good to me.  Thanks!

PR #4439, opened 2024-09-08
Use SSO account to configure sandbox in one click

vrubezhny:
Uses a POC ['feat: use sso account to configure sandbox in one click #4232'](https://github.com/redhat-developer/vscode-openshift-tools/pull/4232) by @dgolovin.

Allows logging in to the DevSandbox using a SSO account by creating a token and installing it into the pipeline Service Account, otherwise, if a token cannot be created/installed, a token from the Clipboard is to be used.

https://github.com/user-attachments/assets/61b612c3-a086-47a4-8619-66215f50370e

### DIFF
--- a/src/oc/ocWrapper.ts
+++ b/src/oc/ocWrapper.ts
@@ -9,10 +9,10 @@ import * as tmp from 'tmp';
 import validator from 'validator';
 import { CommandOption, CommandText } from '../base/command';
 import { CliChannel, ExecutionContext } from '../cli';
+import { CliExitData } from '../util/childProcessUtil';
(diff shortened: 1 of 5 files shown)


datho7561:
I gave it a try and couldn't get it working; I'll try again tomorrow morning

vrubezhny:
> I gave it a try and couldn't get it working; I'll try again tomorrow morning

@datho7561, Hm... Still works for me with no any problem. Rebased, please update before give it a try again.

datho7561 (approved):
Works very well for me! Thanks, Victor

PR #4460, opened 2024-09-12
Always trim the values entered with InputField before validation and use

vrubezhny:
This will help with values pasted from the Clipboard using Ctrl-V


datho7561 (approved):
Looks good. Thanks, Victor!

PR #4473, opened 2024-09-15
"Components" sidebar section take a while to load #3850

vrubezhny:
Fixes: #3850
Depends on: #4467

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/src/devfile-registry/devfileRegistryWrapper.ts
+++ b/src/devfile-registry/devfileRegistryWrapper.ts
@@ -6,7 +6,7 @@ import * as https from 'https';
 import * as YAML from 'js-yaml';
 import { ExecutionContext } from '../cli';
 import { Registry } from '../odo/componentType';
-import { Odo } from '../odo/odoWrapper';
+import { OdoPreference } from '../odo/odoPreference';
 import { DevfileData, DevfileInfo } from './devfileInfo';
 
 export const DEVFILE_VERSION_LATEST: string = 'latest';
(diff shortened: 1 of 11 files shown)


vrubezhny:
This PR depends on #4467

datho7561:
If I add a new registry (eg. https://registry.stage.devfile.io) it appears in the sidebar. However, when I open the Devfile browser, the new Devfiles aren't shown. If I manually hit the refresh button and reopen the Devfile browser, then it seems to work.

If there is an error in the registry URL (eg. I use http instead of https or I misspell some part of the URL), there is an error in the logs, but no notification pops up to tell me that it failed. The only way I can tell that it failed is that there is nothing new in the sidebar.

vrubezhny:
> If I add a new registry (eg. https://registry.stage.devfile.io) it appears in the sidebar. However, when I open the Devfile browser, the new Devfiles aren't shown. If I manually hit the refresh button and reopen the Devfile browser, then it seems to work.

This is for #4467 as this PR only removes the usages of `odo preference ...`-like  command  usages.
A believe it's fixed now in #4467. There is still an issue with adding a registry when the Devfile Editor is already opened, but, imho, it's not about the changes made in #4467  nor in this PR - to be fixed later.

> If there is an error in the registry URL (eg. I use http instead of https or I misspell some part of the URL), there is an error in the logs, but no notification pops up to tell me that it failed. The only way I can tell that it failed is that there is nothing new in the sidebar. 

Fixed in #4467

vrubezhny:
This PR makes it free of any use of `odo preference view/add registry/remove registry`  and makes `~/.odo/preference.yaml` file to be correctly initialized and ready to be used by `odo` (while we continue using it).

This removes the risk of hanging up for about 40 seconds while getting the list of configured Devfile registries in cases when a cluster is not accessible.

datho7561 (approved):
Works well for me. Thanks, Victor!

Issue #4479, opened 2024-09-17
Add UI test for creating Operator-Backend service


lgrossma:
Needs to be done in order to complete https://github.com/redhat-developer/vscode-openshift-tools/issues/4397

PR #4490, opened 2024-09-20
Add UI tests for operator backed service

lgrossma:
See: https://github.com/redhat-developer/vscode-openshift-tools/issues/4479, https://github.com/redhat-developer/vscode-openshift-tools/issues/4397


vrubezhny (approved):
Looks good to me. Thanks!

Issue #4425, opened 2024-09-04
Add additional tag based filtering in helm UI Page 


msivasubramaniaan:
![image](https://github.com/user-attachments/assets/d8fa797e-ac50-40cf-9c6f-2c1b2a6aadfa)

PR #4525, opened 2024-10-03
4425 add additional tag based filtering in helm UI page

msivasubramaniaan:
**Ref:** https://github.com/redhat-developer/vscode-openshift-tools/issues/4425
helm-filter-treeView


datho7561 (changes requested):
One small change to simplify the code, but otherwise looks good and works well

PR #4566, opened 2024-10-17
removed ODO log command

msivasubramaniaan:
Ref: https://github.com/redhat-developer/vscode-openshift-tools/issues/4396

Removed `Show Log` and `Follow Log` on the dropdown of component. We can use `Watch Logs` on `Deployments` and `Pods` in application explorer view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -230,10 +230,6 @@
 		"onCommand:openshift.component.describe",
 		"onCommand:openshift.component.openInBrowser",
 		"onCommand:openshift.component.describe.palette",
-		"onCommand:openshift.component.log",
-		"onCommand:openshift.component.log.palette",
-		"onCommand:openshift.component.followLog",
-		"onCommand:openshift.component.followLog.palette",
 		"onCommand:openshift.componentTypesView.registry.closeView",
 		"onCommand:openshift.component.debug",
 		"onCommand:openshift.component.debug.palette",
@@ -561,26 +557,6 @@
 			{
 				"command": "openshift.component.describe.palette",
 				"title": "Describe Component",
-				"category": "OpenShift"
-			},
-			{
-				"command": "openshift.component.log",
-				"title": "Show Log",
-				"category": "OpenShift"
-			},
-			{
-				"command": "openshift.component.log.palette",
-				"title": "Show Component Log",
-				"category": "OpenShift"
-			},
-			{
-				"command": "openshift.component.followLog",
-				"title": "Follow Log",
-				"category": "OpenShift"
-			},
-			{
-				"command": "openshift.component.followLog.palette",
-				"title": "Follow Component Log",
 				"category": "OpenShift"
 			},
 			{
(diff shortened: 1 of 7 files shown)


src/openshift/component.ts
```diff
@@ -430,26 +430,6 @@ export class Component extends OpenShiftItem {
             `Describe '${componentFolder.component.devfileData.devfile.metadata.name}' Component`);
     }
 
-    @vsCommand('openshift.component.log', true)
-    static async log(componentFolder: ComponentWorkspaceFolder): Promise<void> {
-        const componentName = componentFolder.component.devfileData.devfile.metadata.name;
-        const showLogCmd = Command.showLog(Component.getDevPlatform(componentFolder));
```
vrubezhny, line 436: Shouldn't we remove the command itself then?

vrubezhny (changes requested):
Once we don't use show&watch logs - shouldn't we remove the commands themselves here: https://github.com/redhat-developer/vscode-openshift-tools/blob/3d7ccfc447cd9501fcb578691e6d365e1a2a649e/src/odo/command.ts#L49-L61 ?

src/openshift/component.ts
```diff
@@ -430,26 +430,6 @@ export class Component extends OpenShiftItem {
...
-        await OpenShiftTerminalManager.getInstance().executeInTerminal(
-            showLogCmd,
-            componentFolder.contextPath,
-            `Show '${componentName}' Component Log`);
-    }
-
-    @vsCommand('openshift.component.followLog', true)
-    static async followLog(componentFolder: ComponentWorkspaceFolder): Promise<void> {
-        const componentName = componentFolder.component.devfileData.devfile.metadata.name;
-        const showLogCmd = Command.showLogAndFollow(Component.getDevPlatform(componentFolder));
```
vrubezhny, line 446: Same here. Shouldn't we remove the command itself?

package.json
```diff
@@ -230,10 +230,6 @@
 		"onCommand:openshift.component.describe",
 		"onCommand:openshift.component.openInBrowser",
 		"onCommand:openshift.component.describe.palette",
-		"onCommand:openshift.component.log",
```
msivasubramaniaan, line 233: @vrubezhny I removed the command as well

datho7561 (approved):
Seems good to me. Thanks, Muthu!

vrubezhny:
@msivasubramaniaan Please take a look at https://github.com/redhat-developer/vscode-openshift-tools/pull/4566#pullrequestreview-2375800902 - I mean, you're removing all the "show/watch log"-like commands from palette/context menu/etc., but you're not removing the code in `odo/command.ts` that contains the methods that provide the implementations for these 'ODO` commands:

```
static showLog(platform?: string): CommandText {...}

static showLogAndFollow(platform?: string): CommandText {...}
```
... which look like aren't used anymore (after your PR is merged).

Do you want to keep these methods for any reason?

test/ui/suite/componentContextMenu.ts
lgrossma, line 1: Is it necessary to delete these two tests entirely? If `odo log` gets replaced by different solution, these tests could be still used.. ()

msivasubramaniaan: As plan to get rid of ODO, I removed the whole code and these log functionality is working through `oc logs`. But not in the component level. It will be on deployment and pod level.

vrubezhny (approved):
Look good to me. Thanks!

test/ui/suite/componentContextMenu.ts
lgrossma: I still think it is pretty handy to be able to check the logs in component view and not having to look for the right deployment in the deployments view..

PR #4569, opened 2024-10-17
fix the correct variable name on Alizer workflow.yaml

msivasubramaniaan:


vrubezhny (approved):
Looks good to me. Thanks Muthu!

PR #4584, opened 2024-10-25
Make esbuild to produce bundled & minified VSIX 

### LINKED ISSUE #4226: Extension is slow to activate

vrubezhny:
This fix makes 'esbuild' to produce the bundled and minified VSIX extension archive, 
free of unneeded dependency modules, when the packaging is started with `vsce package`.

On other hand, when building with `npm install && npm run build` and testing with `npm test` 
the extension file structure is kept unchange and the trunspiled scripts aren't minified, 
so the unit testing and coverage tests work as usual.

Note: `npm install` is needed to be executed after executing `vsce package` as the last one 
clears the `node_modules/` of the dependencies not needed in production.

Fixes #4226


vrubezhny:
@msivasubramaniaan @datho7561 @fbricon, When you have time, could you please test the VSIX produced from this PR by executing `vsce package` ?

The overall VSIX size got decreased by ~20MB:

image

... however, imho, the activation time is kept on the same level (when installed from VSIX file) or even increased after this change.

src/util/disposableInstance.ts
```diff
@@ -0,0 +1,18 @@
+/*-----------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE file in the project root for license information.
+ *-----------------------------------------------------------------------------------------------*/
+import { Disposable } from 'vscode';
+
+export class DisposableInstance implements Disposable {
```
datho7561, line 7: This is just used to force the files to have a method to call so that esbuild doesn't mark the files as unused right? If so, I think it would be a good idea to write that down to document it.

datho7561: Also, consider making `getDisposableInstance()` to avoid the naming clashes. For singleton classes, it's a common pattern that we use in the codebase for the `getInstance()` to get the instanceof the singleton.

src/webview/create-deployment/createDeploymentLoader.ts
```diff
@@ -243,7 +245,7 @@ export default class CreateDeploymentLoader {
                 action: 'getRecommendedBuilderImageStart'
             });
             analyzeRes = await Alizer.Instance.alizerAnalyze(uri);
-            normalizedBuilderImages = await BuilderImageWrapper.Instance.getBuilder();
+            normalizedBuilderImages = await (BuilderImageWrapper.getInstance() as BuilderImageWrapper).getBuilder();
```
datho7561, line 248: I think this will fail, because `getInstance()` returns a `DisposableInstance`, not the `BuilderImageWrapper` singleton.

src/odo/builderImage.ts
```diff
@@ -57,26 +56,13 @@ export interface NormalizedBuilderImages {
...
-
-    public static get Instance(): BuilderImageWrapper {
-        if (!BuilderImageWrapper.instance) {
-            BuilderImageWrapper.instance = new BuilderImageWrapper();
-        }
-        return BuilderImageWrapper.instance;
-    }
+export class BuilderImageWrapper extends DisposableInstance {
 
     private constructor() {
```
datho7561, line 61: Is this constructor needed? Do you need to call super in it?

src/util/disposableInstance.ts
vrubezhny: > This is just used to force the files to have a method to call so that esbuild doesn't mark the files as unused right?

In most cases - yes, that was the purpose of having this class. However there are few classes extending this one instead of implementing `getInstance()` in their body (not needed to be mandatory instantiated at the extension activation)

OK. Will fix.

src/vscommand.ts
```diff
@@ -42,7 +42,10 @@ function displayResult(result: string | undefined): void {
 export async function registerCommands(...modules: string[]): Promise<Disposable[]> {
     // this step is required to get all annotations processed and this is drawback,
     // because it triggers javascript loading when extension is activated
-    await Promise.all(modules.map((module) => import(module)));
+    await Promise.all(modules.map(async (module) => {
+            try {await import(module);} catch { /* Ignore */ }
```
datho7561, line 46: This is effectively silencing any errors caused by importing the modules.

When it's bundled, the modules shouldn't need to be imported. Is there a way that we can avoid trying to import the modules when bundled?

I think it's bad that we are silencing all import errors here.

src/webview/create-deployment/createDeploymentLoader.ts
vrubezhny: Imho, it still returns the same instance, doesn't it?  But yes now it needs type conversion like `as BuilderImageWrapper` in order to access `getBuilder()` and/or others `BuilderImageWrapper`'s properties/methods...

test/unit/extension.test.ts
```diff
@@ -103,10 +103,16 @@ suite('openshift toolkit Extension', () => {
 	});
 
     test('should register all extension commands declared commands in package descriptor', async function() {
+        const notRegisteredCommands = [];
         const commands = await vscode.commands.getCommands(true);
         packagejson.contributes.commands.forEach((value)=> {
-            expect(commands.includes(value.command), `Command '${value.command}' handler is not registered during activation`).true;
+            try {
```
datho7561, line 109: instead if a try/catch, you can use an if:

```typescript
if (!commands.includes(value.command)) {
  notRegisteredCommands.push(value.command);
}
```

src/odo/builderImage.ts
vrubezhny: I think, this `private` constructor was added in order to prevent the multiple instantiations... Not sure if this really works this way in TS/JS...
 
The `super()` call was started to be required by eslint after my changes...

Anyway, I moved the VSCommand 
```
@vsCommand('openshift.deployment.create.buildConfig')
    static async createComponent(): Promise<void> {
```
out of this module (due to the cycle dependency problem), so it's not needed to be instantiated at activation and can be rolled back to be:

```
    private static instance: BuilderImageWrapper;

    public static get Instance(): BuilderImageWrapper {
        if (!BuilderImageWrapper.instance) {
            BuilderImageWrapper.instance = new BuilderImageWrapper();
        }
        return BuilderImageWrapper.instance;
    }

    private constructor() {
        // no state
    }
```

src/vscommand.ts
vrubezhny: If we don't process these errors, the activation failure gets "effectively silenced" and never ends without any valuable reason, so the extension stay in `Activating...` until restarted.

Actually these errors ALWAYS  happen  when running the extension installed from VSIX as it physically does not have any of the specified `modules` present. Imho, these errors are expected in production and it's better to hide these errors rather then have them always to be reported in production

datho7561: I think it would be better to skip trying to import all the modules in production, what do you think about this?

test/unit/extension.test.ts
vrubezhny: ACK

src/vscommand.ts
vrubezhny: Yes, but do we have any indication of either we run in production or not?

src/webview/create-deployment/createDeploymentLoader.ts
vrubezhny: Fixed.

src/vscommand.ts
datho7561: I don't know, do you know of a way?

vrubezhny: When I started, I was expecting that we don't have to use `import (module)` at all before registering the commands, because I explicitly call `moduleXXX.getInstance()` before the commands registration. But this didn't work for some reason.
Now, as everything looks like working now, we probably can just remove that list of paths of the modules to be imported - they all already referenced ( === 'inported').

vrubezhny: And yes. This actually works.

src/util/disposableInstance.ts
vrubezhny: Turned everything back to use `vscode Disposable` and removed `DisposableInstance`

vrubezhny: > If so, I think it would be a good idea to write that down to document it.

There's a comment describing why we do need these disposables to be created and pushed to subscriptions

datho7561 (changes requested):
Looking almost ready for merge. Found some potentially unused code

src/util/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Returns the absolute path for a specified relative image path
+ *
+ * @param imagePath An image path relative to the 'images/' directory
+ * @returns An absolute path to specified image
+ */
+export function imagePath(imagePath: string): string {
+    return path.join(__dirname, '../../images/', imagePath);
+}
+
+export function log(message: string): void {
```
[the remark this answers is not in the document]
vrubezhny: No, this is not needed anymore. Thanks.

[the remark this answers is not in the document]
vrubezhny: Removed

datho7561 (approved):
Looks good and works for me. Thanks Victor!

vrubezhny:
@lgrossma Could you please take a look at UI tests:

```
Packaged: /home/runner/work/vscode-openshift-tools/vscode-openshift-tools/vscode-openshift-connector-1.16.0.vsix (421 files, 175.44 MB)
Installing extensions...
Extension 'vscode-openshift-connector-1.16.0.vsix' was successfully installed.
Installing extensions...
Extension 'redhat.vscode-redhat-account' is already installed.
Loading mocha configuration from test/ui/.mocharc.js


  0 passing (1ms)
```

vrubezhny:
The Public UI tests are restored.

msivasubramaniaan (approved):
LGTM

vrubezhny:
@datho7561 @msivasubramaniaan Thanks for the reviews and @lgrossma for helping with the  tests!

PR #4633, opened 2024-11-18
Extester 8.9.0

lgrossma:
@vrubezhny This pr aims to fix test failures in new vscode-extension-tester. It should also stabilize them and for now skip some tests if they are missing prerequisites.


vrubezhny (commented):
@lgrossma Please do not push such a non-significant changes.

Also, I wonder why there is NO any checks started for this change on GH... 
You should have probably adding the new commits to the [dependabot/npm_and_yarn/vscode-extension-tester-8.9.0](https://github.com/redhat-developer/vscode-openshift-tools/tree/dependabot/npm_and_yarn/vscode-extension-tester-8.9.0) branch instead of creating a new PR...

vrubezhny (changes requested):
@lgrossma There is definitely something wrong with this PR... You should either add your commits to the existing PR #4628 or create a new one against the `main` branch that includes both - the change that update vscode-extension-tester to v.8.9.0 and your fixes for the integration tests

lgrossma:
@vrubezhny The checks did not start because the GH action is set to only begin on pull requests and pushes to the main branch. The changes in package.json and package-lock.json are due to Prettier removing empty lines at the end of the file. I did this changes on a separate branch because I was testing them on my fork to not mess with the dependabot PR. Due to these reasons I believe there is nothing wrong with this PR. Furthermore, I believe that creating a PR to another branch is better practice than committing to a branch of someone else (even if it is a bot). Once this is merged, I can continue my work directly in the dependapot PR  if needed.

vrubezhny:
> @vrubezhny The checks did not start because the GH action is set to only begin on pull requests and pushes to the main branch. 

If you add new commits to any existing PR (like #4628) and push them back to the same branch - GH action will be invoked, so your changes are getting tested... You actually did it with https://github.com/redhat-developer/vscode-openshift-tools/pull/4531/commits, didn't you?

> I did this changes on a separate branch because I was testing them on my fork to not mess with the dependabot PR. Due to these reasons I believe there is nothing wrong with this PR. Furthermore, I believe that creating a PR to another branch is better practice than committing to a branch of someone else (even if it is a bot). Once this is merged, I can continue my work directly in the dependapot PR if needed.

OK. But then why are you making your PR against [redhat-developer:dependabot/npm_and_yarn/vscode-extension-tester-8.9.0](https://github.com/redhat-developer/vscode-openshift-tools/tree/dependabot/npm_and_yarn/vscode-extension-tester-8.9.0) branch and not `main`? Make your PR  against the `main` - and everything will be tested...

vrubezhny:
@lgrossma  The problem is that it's not built by GH Actions and I can't see the results. And it's going to take an age if I try building and running tests locally.

vrubezhny (approved):
OK. But it requires a manual rebase now.

lgrossma:
> @lgrossma The problem is that it's not built by GH Actions and I can't see the results. And it's going to take an age if I try building and running the integration tests locally.

No need to run it locally. Once it would get merged into a parent branch, the GH actions would run again for the PR parent branch -> main. Maybe it would be a good idea to also enable GH actions for other branches than only main. Anyways, seems this PR got bugged somewhere in the process and changing the target branch to main did not fire up the GH actions. Created a new PR: https://github.com/redhat-developer/vscode-openshift-tools/pull/4636

PR #4667, opened 2024-11-26
Fix code scanning alert no. 28: Client-side cross-site scripting

vrubezhny:
Fixes [https://github.com/redhat-developer/vscode-openshift-tools/security/code-scanning/28](https://github.com/redhat-developer/vscode-openshift-tools/security/code-scanning/28)

To fix the problem, we need to ensure that the URL used in the `src` attribute of the `img` element is sanitized to prevent XSS attacks. The best way to do this is to use a library that provides robust sanitization functions. One such library is `DOMPurify`, which can sanitize HTML and URLs to prevent XSS.

We will modify the `checkedDevfileLogoUrl` function to use `DOMPurify` to sanitize the URL before returning it. This will ensure that any user-provided URL is safe to use in the `src` attribute of the `img` element.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,6 +62,7 @@
 				"codecov": "^3.8.3",
 				"codemirror": "^6.0.1",
 				"dockerode": "^4.0.2",
+				"dompurify": "^3.2.1",
 				"dpdm": "^3.14.0",
 				"esbuild": "^0.19.12",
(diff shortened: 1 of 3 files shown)


vrubezhny (commented):
LGTM

Issue #3872, opened 2024-02-02
Cannot login if my kube config doesn't exist


datho7561:
1. delete kube config (eg. `rm ~/.kube/config`)
2. open vscode openshift toolkit
3. try to login

**EXPECTED**: you are presented with the login dialog
**ACTUAL**: you are presented with an error saying the kubeconfig doesn't exist

datho7561:
Same here

vrubezhny:
![Image](https://github.com/user-attachments/assets/a7f79db6-0061-4a64-9313-7083ea81a228)

PR #4676, opened 2024-11-29
Cannot login if my kube config doesn't exist #3872

vrubezhny:
Creates a minimalistic Kube config file in `~/.kube/` directory needed for `@kubernetes/client-node` API.
The file contents doesn't define any clusters/users/contexts/etc. properties.

Fixes: #3872


datho7561 (approved):
Works for me

PR #4681, opened 2024-12-02
Remove "Bind Service" feature due to the deprecation of the Service Binding Operator #4682

vrubezhny:
Issue: #4682


src/helm/helm.ts
```diff
@@ -26,8 +26,13 @@ export type HelmRelease = {
  * @returns a list of all Helm releases in the current namespace on the current cluster
  */
 export async function getHelmReleases(): Promise<HelmRelease[]> {
-    const res = await CliChannel.getInstance().executeTool(HelmCommands.listHelmReleases(), undefined, false);
-    return JSON.parse(res.stdout) as HelmRelease[];
+    try {
+        const res = await CliChannel.getInstance().executeTool(HelmCommands.listHelmReleases(), undefined, false);
+        return JSON.parse(res.stdout) as HelmRelease[];
+    } catch (err) {
+        window.showErrorMessage(`Cannot list HELM Releases: ${err}`);
```
datho7561, line 33: IMHO it makes more sense to let the error be passed on to whatever method is calling this one. What do you think?

src/oc/ocWrapper.ts
```diff
@@ -85,10 +86,15 @@ export class Oc {
     }
 
     public async getAllKubernetesObjects(namespace?: string, executionContext?: ExecutionContext): Promise<KubernetesObject[]> {
-        const result = await CliChannel.getInstance().executeTool(
-            Oc.getKubernetesObjectCommand('all', namespace),
-            undefined, true, executionContext);
-        return JSON.parse(result.stdout).items;
+        try {
```
datho7561, line 89: Same note as above here

src/oc/ocWrapper.ts
```diff
@@ -85,10 +86,15 @@ export class Oc {
...
+            const result = await CliChannel.getInstance().executeTool(
+                Oc.getKubernetesObjectCommand('all', namespace),
+                undefined, true, executionContext);
+            return JSON.parse(result.stdout).items;
+        } catch (err) {
+            window.showErrorMessage(`Cannot list HELM Releases: ${err}`);
```
datho7561, line 95: These are not helm releases, they are Kubernetes objects

src/helm/helm.ts
vrubezhny: When this method (and yet another one from the `osWrapper` with try-catch block added) are invoked from `OpenShiftExplorer.getChildren()` it breaks the method execution if `JSON.parse()` fails to process the `stdout` of the result - this makes the explorer tree fail on update. I could actually ignore the error (as it's usually "json is malformed"-like error, but I'd like to know when it usually happens

src/oc/ocWrapper.ts
vrubezhny: Same answer as above

src/oc/ocWrapper.ts
vrubezhny: Oh. Sure.

src/helm/helm.ts
datho7561: Okay sounds good!

datho7561 (approved):

Issue #4472, opened 2024-09-15
Updating deployment.yaml doesn't reflect into openshift


surnandi-git:
Type: <b>Bug</b>

## Describe the bug
Since Auto Deploy is not happening, I was copying the image id and putting it into deplyment yaml to get the app refreshed. Performing the same in through code, isn't working.  

## Expected Behavior
Saving the deployment yaml file should be reflected into the OpenShift deployment. And a new pod should be created (terminating existing)

## Current Behavior
It doesn't reflect change made into OpenShift, although it shows save. Upon relaod it brings the older file from server

## Steps to Reproduce
1. Have a application created through using deployment from git
2. Estiblish webhook for auto build
3. Make some code change and observe the build
4. Capture the new image id from VSCode -> Image Stream -> Select your image and copy the image id
5. Go to VSCode -> Deployments -> Select your deployment (yaml opens up)
6. Update the image id in the deployment yaml
7. Save the file.

## Environment
OpenShift 
VSCode Extension

Extension version: 1.16.0
VS Code version: Code 1.93.0 (Universal) (4849ca9bdf9666755eb463db297b69e5385090e3, 2024-09-04T13:02:38.431Z)
OS version: Darwin arm64 23.6.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 (8 x 2400)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|268, 88, 41|
|Memory (System)|16.00GB (3.22GB free)|
|Process Argv||
|Screen Reader|yes|
|VM|0%|
</details>

mohitsuman:
@datho7561 can you please look into this ?

datho7561:
We don't currently support this. The feature of loading the YAML for a resource on the cluster is intended to be read-only.

However, I think it's a good idea to apply the changes to the cluster when the document is saved.

[...]

PR #5074, opened 2025-05-26
Updating deployment.yaml doesn't reflect into openshift #4472

vrubezhny:
Fixes: #4472


vrubezhny:
Editing and saving a custom deployment:

https://github.com/user-attachments/assets/91903b88-4a81-42f5-8810-2df08b6effd0

PR #5361, opened 2025-09-18
Devfile Registries list need not contain Default Devfile Registry.

rgrunber:
- When the Devfile Registries list contains other registries, it should be permitted to remove the Default Devfile Registry
- Replace usage of https.get(..) with fetch

- [ ] Need to test this more thoroughly
- [x] I can probably adjust https://github.com/redhat-developer/vscode-openshift-tools/blob/0ceadd98c8fc5e5d61e01ebf786a182fc6cd3045/src/openshift/cluster.ts#L671 as well

### DIFF
--- a/src/devfile-registry/devfileRegistryWrapper.ts
+++ b/src/devfile-registry/devfileRegistryWrapper.ts
@@ -2,7 +2,8 @@
  *  Copyright (c) Red Hat, Inc. All rights reserved.
  *  Licensed under the MIT License. See LICENSE file in the project root for license information.
  *-----------------------------------------------------------------------------------------------*/
-import * as https from 'https';
+import { get as httpGet } from 'http';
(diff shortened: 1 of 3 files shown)


vrubezhny:
> make-fetch-happen was removed right after that, at #4247 . Is this still an issue ?

Nope, but you're trying to bring it back with your PR

rgrunber:
Ok, I've gone with just switching between `http`/`https`.

vrubezhny:
The test https://github.com/redhat-developer/vscode-openshift-tools/blob/17de0413188d0ce818857267109caa90d402c26d/test/ui/suite/devfileRegistries.ts#L131 expects at least the Default Devfile Registry to be present (as well as some other tests). 
However it looks like there is no Default Devfile registry added to the configuration - previously it was always merged into the "effective" config.

PR #5505, opened 2025-11-24
Log the output of the git-clone command into the OpenShift Logs.

rgrunber:
- Convert 'exec' call to use 'spawn'


rgrunber:
I'm going to merge this. I haven't applied the same changes to the deployment workflow, but I can't seem to get that wizard page to come up. It can be done some other time.

Issue #4682, opened 2024-12-03
Make "Bind Service" feature to use `servicebinding.io` API instead of the deprecated Service Binding Operator


vrubezhny:
We have no need to keep "Bind Service" command and the  related UI due to the deprecation of the Service Binding Operator. 

See: https://github.com/redhat-developer/service-binding-operator

![Image](https://github.com/user-attachments/assets/e3e17867-bb9d-4ac0-8a71-a76ae1bb3c0e)

PR #6000, opened 2026-07-28
list of service binding and added binding using oc

msivasubramaniaan:


vrubezhny:
SBO Migration Required

  PR #6000 brings the service binding feature back by:

  1. Removing odo dependency — addBinding() and loadView() no longer use Odo.Instance, all binding operations go through BindableService and the K8s API directly
  2. Fixing bugs — early return with empty password preventing TokenStore persistence, stale KubeConfig in singleton, wrong module import, missing componentName parameter
  3. Restoring tests — unit, integration, and UI tests for the binding flow, renamed from operatorBackedService to bindableService to reflect the direction away from SBO

  However, the current implementation still depends on the Service Binding Operator (SBO) for two critical functions:

  - Service discovery via the BindableKinds CRD (oc get bindablekinds.binding.operators.coreos.com)
  - Binding creation via the ServiceBinding CRD

  SBO was deprecated in February 2024. Testing revealed that SBO's BindableKinds does not auto-discover operators that implement the Provisioned Service pattern (e.g., CloudNativePG), even when both SBO and the
  operator are installed on the same cluster. The old odo-based code worked because odo had its own discovery mechanism independent of BindableKinds.

  This means the binding feature restored in this PR is effectively non-functional on clusters where operators don't explicitly register themselves with SBO via CSV descriptors — which includes most modern
  operators that follow the servicebinding.io spec instead.

  This PR does not resolve #4682 — Make "Bind Service" feature to use servicebinding.io API instead of the deprecated Service Binding Operator. It removes the odo dependency and restores the infrastructure, but
  the underlying SBO dependency remains and must be replaced in a follow-up.

  Next step: Replace SBO-based discovery and binding with pure K8s API operations:
  - Discovery: Scan cluster CRDs/resources for .status.binding.name (Provisioned Service pattern)
  - Binding: Patch Deployments directly to mount the binding secret as a volume at /bindings/<name>/ and set SERVICE_BINDING_ROOT=/bindings

  This eliminates the SBO dependency entirely and supports any operator that implements the Provisioned Service pattern.

vrubezhny:
Merging as it's a good base to start migrating out of SBO

PR #6038, opened 2026-08-16
[build] Configure npm 11.x allowScripts for native module builds

vrubezhny:
## Summary

Replace `--ignore-scripts` with npm 11.x's `allowScripts` configuration for more secure and compatible native module handling.

- Use explicit whitelist of packages allowed to run install scripts
- Replace `npm install` with `npm ci` in build scripts (works better with allowScripts)
- Use package name wildcards in allowScripts to handle transitive dependency version changes
- Remove redundant `.npmrc` (npm 11.x prioritizes package.json allowScripts)

**Why:** npm 11.x enforces allow-scripts configuration. The blanket `--ignore-scripts` flag conflicts with this system. The new approach is more secure (explicit whitelist vs blanket ban) and compatible with Node 24.19's npm.

**Fixes:** `vsce package` command now works with Node 24.19 + npm 11.x

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -63,8 +63,7 @@ jobs:
       # Run install dependencies
       - name: Install dependencies
         run: |
-          npm ci --ignore-scripts
-          npm rebuild cpu-features ssh2
+          npm ci
 
       # Run tests without xvfb, if not running on ubuntu
       - name: Build and Test (non-Linux)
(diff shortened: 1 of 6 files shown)


vrubezhny:
Closing this - after further testing, the original `vsce package` failure doesn't appear to be caused by npm 11.x's allow-scripts policy itself.

Verified from a genuinely clean state (`rm -rf node_modules`, no local script approvals) that `npm ci --ignore-scripts && npm rebuild cpu-features ssh2` - the exact commands used by `continuous-integration-workflow.yml` and effectively by `release.yml`/`vsce package`'s own `vscode:prepublish` - completes successfully on Node 24.19/npm 11.17.0, and that the full `vscode:prepublish` pipeline (lint, esbuild build, tool bundling) also produces a working `.vsix`. CI doesn't cache `node_modules` either, so it's always doing the same from-scratch install.

The original local failure was most likely caused by stale native binaries (keytar, cpu-features, ssh2, esbuild, etc.) left over in `node_modules` from before the Node upgrade, rather than anything `--ignore-scripts` itself does wrong.

The `allowScripts` config here would still have value for silencing the `npm warn allow-scripts` noise on every install, but it's not fixing a real breakage, so closing for now.

PR #6088, opened 2026-09-06
[tests] Replace clipboard with tinyclip

vrubezhny:


vrubezhny:
tinyclip doesn't work properly on GH CI runner, so we stay stick to clipboardy. It's bigger but works fine.

PR #6097, opened 2026-09-08
[feature] Replace odo dev/odo dev --debug with a native TypeScript engine (CRW-10963)

vrubezhny:
Replaces the CLI shell-out to `odo dev` / `odo dev --debug` with a native TypeScript dev-mode engine, following the same pattern already used to replace `odo init`/`odo deploy`/`odo undeploy`/`odo describe`. Dev mode now runs directly against the current Kubernetes/OpenShift cluster or a local podman runtime — deploying/starting the dev container(s), performing an initial `.gitignore`-aware file sync plus an ongoing file watcher that syncs and restarts the run/debug command on change, executing that command, and forwarding its endpoints — with no remaining dependency on the `odo` CLI for this feature. The component context-menu commands (dev/dev.manual/dev.onPodman/dev.onPodman.manual/exitDevMode/ forceExitDevMode/showDevTerminal) are rewired to this engine, and the OpenShift Terminal panel UX (live output, Ctrl-C-to-stop) is preserved via a generalized virtual-terminal mode rather than a real spawned process. `describe`'s status reporting was also reworked to accurately reflect simultaneous Dev/Deploy state, `managed-by`, and the actual cluster platform (OpenShift/Kubernetes/Kind/Minikube) for both engines.

Extends the Debug command to support Go components in addition to Java, Node.js, and Python, with automatic Delve debugger configuration and VS Code debugger attachment.

Adds unit and integration test coverage for all of the new dev-engine modules (session lifecycle, cluster/podman platforms, file sync, command execution, port-forwarding, terminal bridging) and extends existing test suites (`describe`, `kubeUtils`, `component`) to cover the reworked behavior.

Assisted-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/doc/plans/project_inner_loop_dev_plan.md
+++ b/doc/plans/project_inner_loop_dev_plan.md
@@ -1,0 +1,717 @@
+# Native `odo dev` / `odo dev --debug` Replacement — Plan
+
+Status legend: `[ ]` not started · `[~]` in progress · `[x]` done
+
+## Goal
+
+Replace the CLI shell-out to `odo dev` / `odo dev --debug` (currently `src/odo/command.ts`'s
+`Command.dev()`, invoked from `src/openshift/component.ts`'s `devRunOn`) with a native
+TypeScript implementation, following the same pattern already used to replace `odo init`
+(`src/devfile/init.ts`), `odo deploy` (`src/devfile/deploy.ts`), and `odo undeploy`
+(`src/devfile/undeploy.ts`).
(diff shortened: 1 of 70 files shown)


vrubezhny:
A short demo of Start/Stop Dev Mode on local Kind cluster and Podman:


https://github.com/user-attachments/assets/c9504e9c-9fce-4ea7-9bf3-6011562686cf

